PR #5, opened 2018-12-21
Rename kymactl to kyma

lilitgh:


Makefile
```diff
@@ -18,17 +18,17 @@ validate:
 .PHONY: build
 build:
 	go generate ./...
-	CGO_ENABLED=0 GOOS=windows GOARCH=amd64 go build -o ./bin/kymactl.exe $(FLAGS) ./cmd/kymactl.go
-	CGO_ENABLED=0 GOOS=linux GOARCH=amd64 go build -o ./bin/kymactl-linux $(FLAGS) ./cmd/kymactl.go
-	CGO_ENABLED=0 GOOS=darwin GOARCH=amd64 go build -o ./bin/kymactl-darwin $(FLAGS) ./cmd/kymactl.go
+	CGO_ENABLED=0 GOOS=windows GOARCH=amd64 go build -o ./bin/kyma.exe $(FLAGS) ./cmd/kymactl.go
+	CGO_ENABLED=0 GOOS=linux GOARCH=amd64 go build -o ./bin/kyma-linux $(FLAGS) ./cmd/kymactl.go
+	CGO_ENABLED=0 GOOS=darwin GOARCH=amd64 go build -o ./bin/kyma-darwin $(FLAGS) ./cmd/kymactl.go
```
sjanota: What about 386?

a-thaler: will complicate the binary naming even more, do you think it is relevant nowadays?

sjanota: Nope, just asking

pkg/kymactl/cmd/version.go
```diff
@@ -42,7 +42,7 @@ func (o *VersionOptions) Run() error {
 	if version == "" {
 		version = "N/A"
 	}
-	fmt.Printf("Kymactl version:      %s\n", version)
+	fmt.Printf("Kyma CLI version:      %s\n", version)
```
a-thaler: Amount of spaces should be reduced by 1 to have formatting aligned with server side version

pkg/kymactl/cmd/version.go
```diff
@@ -25,8 +25,8 @@ func NewVersionOptions() *VersionOptions {
 func NewVersionCmd(o *VersionOptions) *cobra.Command {
 	cmd := &cobra.Command{
 		Use:   "version",
-		Short: "Version of the kymactl and connected Kyma cluster",
-		Long: `Prints the version of kymactl itself and the version of the kyma cluster connected by current KUBECONFIG
+		Short: "Version of the kyma CLI and connected Kyma cluster",
```
a-thaler: Maybe we should call it kyma client in the version context (versus kyma server)

README.md
```diff
@@ -9,28 +9,30 @@ A command line tool to support developers of and with Kyma
 - `version`: Shows the kyma cluster version and the kymactl version. The kymactl version is set at compile time passing it to the go linker as a flag:
```
sjanota:
```suggestion
- `version`: Shows the kyma cluster version and the kyma CLI version. The kyma CLI version is set at compile time passing it to the go linker as a flag:
```

pkg/kymactl/cmd/version.go
sjanota: I'm ok with Kyma CLI (capitalisation btw), but I'm thinking about renaming whole project to `kyma-cli`. `kymactl` feels like binary name, but resulting binary is named differently.

a-thaler: interesting point, let's talk about it after this PR. I think we should do it indeed.

pkg/kymactl/cmd/version.go
lilitgh: done

lilitgh:
Renamed cmd/kymactl.go -> cmd/kyma.go , pkg/kymactl.go -> pkg/kyma.go. Adjusted Makefile and Readme accordingly

a-thaler:
There is no docker image, it is just the project name used in prow. As we decided to rename the project later it should stay untouched.

sjanota:
My mistake. I've just seen the IMG at the top :D

PR #6, opened 2018-12-21
List releases

a-thaler:
Added new command to list available releases, done together with @atakanyenel


sjanota:
Do we plan to support installing all releases using this CLI? If not I don't see any value from this command. If I'm not able to do anything else with the release from CLI I could as well go to GitHub releases page.

a-thaler:
My feeling is that we should be able to install an older version even with the newest CLI version. Otherwise, I need to start downgrading my client just because I need to install an older cluster...

So I added the option already in the install command to specify the release version, it identifies the config yaml to download from github by the CLI. For that, I wanted to have an easy way to see available versions and later leverage the functionality to also pick just the latest.

Maybe lets have the command for now and let's see later how to integrate into the install command in a convenient way

a-thaler:
After long discussions we decided that for now it is better if a cli binary know how to handle a specific kyma version only (not supporting older or newer versions). Let's introduce support for multiple versions more corrdinated as it will require more efforts (think of kubectl/kubernetes/helm version support)

Furthermore the used solution requires to use a private github token which is not feasible in real life. In general we should introduce release channels for kyma, hosted on something like cloud storage or similar so that the rate limiting of github API will be no problem.

I will close that PR without merge

PR #8, opened 2018-12-28
Add possibility to install Kyma from local sources

sjanota:
New option has been added to `kyma install kyma` command: `--local`. When set CLI will install Kyma from local sources instead of release. Source path defaults to `$GOPATH/src/github.com/kyma-project/kyma and can be overriden with `--src-path` flag.


sjanota:
@a-thaler @mszostok @lilitgh I'm thinking about some renaming. How about making `install kyma` two separate commands: `install local` and `install release`?

sjanota:
There is still one difference compared to original run.sh: In Installation CR placeholders are left untouched while when installing with run.sh they are simply removed.

PR #9, opened 2019-01-04
Add make command to install kymactl

sjanota:
This PR adds `make install` which compiles installs kyma binary in `GOPATH/kyma`.


sjanota:
Closing to fix CLA issue

PR #11, opened 2019-01-08
Add test command

sjanota:
This PR includes new `kyma test` command which is a replacement for the `testing.sh` script.


internal/step/spinner.go
```diff
@@ -0,0 +1,58 @@
...
+	"time"
+)
+
+func NewStepWithSpinner(msg string) Step {
+	s := spinner.New([]string{"/", "-", "\\", "|"}, time.Millisecond * 200)
+	s.Prefix = msg+" "
+	return  &stepWithSpinner{s, msg}
+}
+
+type stepWithSpinner struct {
```
a-thaler: That is a nice new spinner, thanks for adding it. However, you have not replaced the usage of the old one, please apply it consistently to all. I can support you with that if you prefer.
I realized that the new spinner has a much more light-grey appearance on my terminal then the old one. Can you check if the tile/frame itself can be a bit more black..

pkg/kyma/cmd/test/cmd.go
```diff
@@ -0,0 +1,155 @@
+package test
```
a-thaler: why is that a whole new package, not just a new test.go in package above?

pkg/kyma/cmd/test/cmd.go
```diff
@@ -0,0 +1,155 @@
...
+	"k8s.io/client-go/tools/clientcmd"
+	"k8s.io/helm/pkg/helm"
+	"k8s.io/helm/pkg/helm/environment"
+	helm_release "k8s.io/helm/pkg/proto/hapi/release"
+)
+
+type Options struct {
+	step.Factory
+	skip []string
+	verbose bool
```
a-thaler: Do you have an idea how to have it as a global parameter? The verbose flag will be important for all commands and I don't want to duplicate it's description in every command.

pkg/kyma/cmd/test/cmd.go
```diff
@@ -0,0 +1,155 @@
...
+
+	cmd.Flags().StringArrayVar(&opts.skip, "skip", []string{}, "Skip tests for these releases")
+	cmd.Flags().BoolVar(&opts.NonInteractive, "non-interactive", false, "Do not use spinners")
+	cmd.Flags().BoolVarP(&opts.verbose, "verbose", "v", false, "Print additional output")
+	return cmd
+}
+
+func (opts *Options) Run() error {
+	helmConfig := &environment.EnvSettings{TillerConnectionTimeout: 300}
+	kubeConfig, err := clientcmd.BuildConfigFromFlags("", clientcmd.RecommendedHomeFile)
```
a-thaler: Can be on a global level probably as we should convert out kubectl commands to use the go client

pkg/kyma/cmd/test/cmd.go
```diff
@@ -0,0 +1,155 @@
...
+		}
+		if err != nil {
+			return err
+		}
+	}
+
+	return nil
+}
+
+func (opts *Options) skipRelease(client helm.Interface, release string) bool {
```
a-thaler: Why this is a method defined on the struct instead of just an internal function having an options parameter?

pkg/kyma/cmd/test/cmd.go
```diff
@@ -0,0 +1,155 @@
...
+	for _, skipped := range opts.skip {
+		if skipped == release {
+			return true
+		}
+	}
+
+	if !testReleases[release] {
```
a-thaler: Somehow the testing failed for me for the "knative" release. It is not installed in my case, the same as monitoring but monitoring got skipped. Please re-check.

pkg/kyma/cmd/test/cmd.go
```diff
@@ -0,0 +1,155 @@
...
+		s.Failure()
+		return err
+	}
+
+	s.Success()
+	return nil
+}
+
+func (opts *Options) testRelease(client helm.Interface, s step.Step, release string) (string, bool, error) {
+	c, errc := client.RunReleaseTest(release, helm.ReleaseTestTimeout(600))
```
a-thaler: After I executed all tests (with failing knative release) I tried to re-run the test command. It passed all stages again but was not executing any test anymore. Please re-check

pkg/kyma/cmd/test/cmd.go
a-thaler:
````
Cleaning up helm test pods in namespace kyma-system ✅
Cleaning up helm test pods in namespace istio-system ✅
Cleaning up helm test pods in namespace knative-serving ✅
Cleaning up helm test pods in namespace kyma-integration ✅
Testing release core ✅
Skipping release monitoring ✅
Skipping release logging ✅
Testing release istio ✅
Testing release knative ❌
Error: rpc error: code = Unknown desc = release: "knative" not found
rpc error: code = Unknown desc = release: "knative" not found
```

pkg/kyma/cmd/test/cmd.go
```diff
@@ -0,0 +1,155 @@
...
+	cmd := &cobra.Command{
+		Use:   "test",
+		Short: "test kyma installation",
+		RunE: func(cmd *cobra.Command, args []string) error {
+			return opts.Run()
+		},
+	}
+
+	cmd.Flags().StringArrayVar(&opts.skip, "skip", []string{}, "Skip tests for these releases")
+	cmd.Flags().BoolVar(&opts.NonInteractive, "non-interactive", false, "Do not use spinners")
```
a-thaler: am thinking if the non-interactive mode for your test command has the same semantic like the silent mode for minikube install. I think there are different as I might want to disable the dialogues (do you really want to delete) but still want to have nice spinner. But that is maybe so rare that we still should unify them to a "non-interactive" but what could be a short key?

sjanota: In case of dialogues I would use something like `--default/-d` which would always use less intrusive action and `--force/-f` to do the opposite. Not sure why, but people tend to use `--force` more frequently then its counterpart which often is not even implemented :D

pkg/kyma/cmd/test/cmd.go
sjanota: You probably have kyma installed from release 0.5 (knative is not there). I'm thinking about reading list of modules (helm releases) instead of having them hardcoded.

pkg/kyma/cmd/test/cmd.go
sjanota: This is probably because test pods were not removed before tests. I need to investigate further.

pkg/kyma/cmd/test/cmd.go
sjanota: Just my nasty habit of having "main" object for a command. It makes it easier to pass parameters and common items around when it is a receiver. I like to make functions which parameters are really the varying part while receiver represents context of the operation (OOP legacy thinking). In this particular example `client helm.Interface` could even be part of receiver to achieve this separation.

TL;DR Because I wrote it so.

pkg/kyma/cmd/test/cmd.go
sjanota: I just followed the convention. `install` and `uninstall` have their own packages. Now I see that they also have subcommands, that's why they are separated. Should I flatten `test` package?

pkg/kyma/cmd/test/cmd.go
sjanota: Now, instead of having hardcoded list of components tool is reading them from installation CR.
This should fix the problem.

pkg/kyma/cmd/test/cmd.go
sjanota: I wasn't able to reproduce after I made changes with reading components from installation CR. Could you recheck on new version?

internal/step/spinner.go
sjanota: The spinner had foreground set to white. I've cleared all styles on spinner, now it's back to normal.

pkg/kyma/cmd/test/cmd.go
sjanota: I've flattened the directory

pkg/kyma/cmd/test/cmd.go
sjanota: I moved the Verbose flag to KymaOptions and made TestOptions extend it. I also moved step.Factory there so every command can have it available

pkg/kyma/cmd/test/cmd.go
sjanota: I've created `kube.ConfigFactory` and made `KymaOptions` implement it. It is now available to all commands.

a-thaler:
I applied the new spinner everywhere and introduced global config for all commands

internal/helm/helm.go
```diff
@@ -0,0 +1,102 @@
...
+type Client struct {
+	helm.Interface
+	forwarder chan<- struct{}
+}
+
+func New(settings *environment.EnvSettings, config *rest.Config) (*Client, error) {
+	forwarder, err := setupTillerConnection(settings, config)
+	if err != nil {
+		return nil, err
+	}
```
a-thaler: Somehow the helm client makes a lot of verbose logs now, can you please double check

internal/step/spinner.go
a-thaler: looks fine now, thanks

pkg/installer/installer.go
```diff
@@ -0,0 +1,24 @@
...
+
+func GetComponents(kubeConfig *rest.Config) ([]installer_api.KymaComponent, error) {
+	installerClient, err := installer.NewForConfig(kubeConfig)
+	if err != nil {
+		return nil, err
+	}
+
+	installation, err := installerClient.InstallerV1alpha1().Installations("kyma-installer").Get("kyma-installation", metav1.GetOptions{})
+	if err != nil {
+		installation, err = installerClient.InstallerV1alpha1().Installations("default").Get("kyma-installation", metav1.GetOptions{})
```
a-thaler: see https://github.com/kyma-project/kyma/issues/2358, dependent on how you installed it, the namespace is different

pkg/kyma/core/options.go
```diff
@@ -0,0 +1,18 @@
+package core
```
a-thaler: I didn't knew where to place it better to be usable across all components, feel free to move it

pkg/kyma/cmd/test/cmd.go
a-thaler: works like a charm. I rolled it out to all commands

pkg/kyma/cmd/test/cmd.go
a-thaler: looks good

pkg/kyma/cmd/test/cmd.go
a-thaler: fine for me now.

pkg/kyma/cmd/test/cmd.go
a-thaler: For simplicity I used the non-interactive flag for now. Maybe the --force is better indeed.

internal/helm/helm.go
sjanota: Can you paste an example? I haven't noticed.

pkg/kyma/core/options.go
sjanota: Why separate package is needed in your opinion? ' so tightly coupled with `kyma` command that I would see it in `kyma` package.

internal/helm/helm.go
a-thaler: I had an env variable active "DEBUG=true". After removing that the logs disappeared.

pkg/kyma/core/options.go
a-thaler: Not sure what you mean with "kyma" package, can you point me to a folder

PR #17, opened 2019-01-24
restructuring of commands

a-thaler:
I tried to improve the "kyma install kyma" repetition by changing it to "kyma install" only. The cluster preparation I switched to "kyma prepare minikube"

The new structure is:
kyma install
kyma uninstall
kyma test
kyma prepare minikube

It might be still improved but it is already much better then before :)


a-thaler:
I was thinking of "setup" as well... ok, you convinced me, provision it will be

a-thaler:
Maybe the CLI can be smart and try figuring it out by cluster specificas..
anyway, I know your issue https://github.com/kyma-incubator/kyma-cli/issues/10 and I agree that it needs to be solved. However, this change improves the existing logic into something more usable with small effort. Let's come up with a solution for the issue separately.

a-thaler:
Renamed to "provision", please check again

sjanota:
I was thinking about different structure. Because the tool aims to install Kyma after all I'd focus on doing exactly that. What do you think about something like this instead:
```
$ kyma install --provider=minikube --minikube-vm-driver=hyperkit --local
```
In case you have existing minikube you'd like to use you add `--skip-provision` flag, like this:
```
$ kyma install --provider=minikube --skip-provision --local
```
I agree that it may not be change we want to do now as it's a lot bigger, but is it even worth considering?

a-thaler:
I updated your ticket description to the new command structure. Let's have a discussion on that, it is worth of I think :)

Issue #19, opened 2019-02-04
Error in flag --format, {{.MinikubeStatus}} is not supported


venturasr:
**Description**

Functions **checkIfMinikubeIsInitialized** and **waitForMinikubeToBeUp** in [provision/minikube.go](
        kyma-cli/pkg/kyma/cmd/provision/minikube.go
      ) use flag `--format`  as is shown below.

```statusText, err := internal.RunMinikubeCmdE([]string{"status", "-b=" + bootstrapper, "--format", "'{{.MinikubeStatus}}'"})```

This originates the following error 
```Minikube is initialized and status is 'E0204 08:05:51.556475  180376 status.go:137] Error executing status template: template: status:1:3: executing "status" at <.MinikubeStatus>: cant evaluate field MinikubeStatus in type cmd.Status' ```

In order to solve this issue **{{.MinikubeStatus}}** must be replaced by **{{.Host}}** as is indicated in `minikube status --help`

```
Flags:
      --format string   Go template format string for the status output.  The format for Go templates can be found here: https://golang.org/pkg/text/template/
                        For the list accessible variables for the template, see the struct values here:
https://godoc.org/k8s.io/minikube/cmd/minikube/cmd#Status (default "host: {{.Host}}\nkubelet: {{.Kubelet}}\napiserver: {{.ApiServer}}\nkubectl: {{.Kubeconfig}}")```

PR #21, opened 2019-02-04
Minikube status bug

venturasr:
Fix issue [#19](https://github.com/kyma-incubator/kyma-cli/issues/19)

Minikube version updated to 0.31.0
`minikubeVersion string = "0.31.0"`

### DIFF
--- a/internal/minikube.go
+++ b/internal/minikube.go
@@ -10,7 +10,7 @@ import (
 )
 
 const (
-	minikubeVersion string = "0.28.2"
+	minikubeVersion string = "0.31.0"
 )
 
 //RunMinikubeCmd executes a minikube command with given arguments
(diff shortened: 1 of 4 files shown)


sjanota:
@venturasr Have you tested it with release 0.6 ? It was using older minikube, so we need to verify if this change is compatible. We need to add a not if kyma-cli would not support 0.6 after this is merged.

sjanota:
I was running it with minikube 0.33 wich is supported by kyma master and it works fine.

internal/minikube.go
```diff
@@ -10,7 +10,7 @@ import (
 )
 
 const (
-	minikubeVersion string = "0.28.2"
+	minikubeVersion string = "0.31.0"
```
sjanota: The supported version is 0.33.0

sjanota:
Already merged in #22

Issue #20, opened 2019-02-04
'exit status 1' with '-bootstrapper=localkube '


venturasr:
**Description**
Error while running `.\kyma.exe provision minikube --vm-driver virtualbox`

```
...
Starting VM...
Getting VM IP address...
E0204 08:12:50.106726  182476 start.go:210] Error parsing version semver:  Version string empty
F0204 08:12:50.115248  182476 start.go:244] Error getting cluster bootstrapper: Unknown bootstrapper: 

localkube
' and error message 'exit status 1'
Failed executing minikube command 'minikube [start --memory 8192 --cpus 4 --extra-config=apiserver.Authorization.Mode=RBAC --extra-config=apiserver.GenericServerRunOptions.CorsAllowedOriginList='.*' --extra-config=controller-manager.ClusterSigningCertFile='/var/lib/local
kube/certs/ca.crt' --extra-config=controller-manager.ClusterSigningKeyFile='/var/lib/localkube/certs/ca.key' --extra-config=apiserver.admission-control='LimitRanger,ServiceAccount,DefaultStorageClass,MutatingAdmissionWebhook,ValidatingAdmissionWebhook,ResourceQuota' --ku
bernetes-version=v1.10.0 --vm-driver=virtualbox --disk-size=20g --feature-gates='MountPropagation=false'  -b=localkube]' with output 'Starting local Kubernetes v1.10.0 cluster...
Starting VM...
Getting VM IP address...
E0204 08:12:50.106726  182476 start.go:210] Error parsing version semver:  Version string empty
F0204 08:12:50.115248  182476 start.go:244] Error getting cluster bootstrapper: Unknown bootstrapper: localkube
' and error message 'exit status 1'
PS C:\golang\work\src\github.com\kyma-incubator\kyma-cli\bin> minikube version
minikube version: v0.31.0
PS C:\golang\work\src\github.com\kyma-incubator\kyma-cli\bin>
```

`-bootstrapper=localkube` must be replaced by `--bootstrapper=kubeadm`.

PR #22, opened 2019-02-11
Adapt to 0.7 Kyma release

sjanota:
In Kyma 0.7 newer kubernetes and Minikube are used. This PR bumps this in CLI code and adapts to new versions of those tools.

As a bonus following bugs are fixed:
#20 
#19 
Problem with displaying installation steps

### DIFF
--- a/pkg/kyma/cmd/provision/minikube.go
+++ b/pkg/kyma/cmd/provision/minikube.go
@@ -3,6 +3,7 @@ package provision
 import (
 	"bufio"
 	"fmt"
+	"github.com/kyma-incubator/kyma-cli/internal/step"
 	"os"
 	"runtime"
 	"strings"
@@ -14,8 +15,8 @@ import (
 )
 
 const (
-	kubernetesVersion string = "1.10.0"
-	bootstrapper      string = "localkube"
+	kubernetesVersion string = "1.11.5"
+	bootstrapper      string = "kubeadm"
 	vmDriverHyperkit  string = "hyperkit"
 	vmDriverHyperv    string = "hyperv"
 	vmDriverNone      string = "none"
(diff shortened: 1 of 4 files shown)


pkg/kyma/cmd/provision/minikube.go
```diff
@@ -129,30 +130,34 @@ func (o *MinikubeOptions) Run() error {
...
 		s.Failure()
 		return err
 	}
 
-	err = internal.WaitForPod("kube-system", "k8s-app", "kube-dns")
-	if err != nil {
-		s.Failure()
-		return err
-	}
+	//s.Status("Await kube-dns to be up and running")
```
a-thaler: Can you please decide if it is needed or not and then keep or delete the code.

pkg/kyma/cmd/provision/minikube.go
```diff
@@ -179,16 +184,12 @@ func (o *MinikubeOptions) Run() error {
 }
 
 func checkIfMinikubeIsInitialized(o *MinikubeOptions) error {
-	statusText, err := internal.RunMinikubeCmdE([]string{"status", "-b=" + bootstrapper, "--format", "'{{.MinikubeStatus}}'"})
```
a-thaler: Shouldn't we keep the bootstrapper argument? yes, we are using the default, but then we should delete it from start command as well and delete the variable..

pkg/kyma/cmd/provision/minikube.go
```diff
@@ -233,27 +233,23 @@ func initializeMinikubeConfig() error {
...
 	startCmd := []string{"start",
 		"--memory", o.Memory,
 		"--cpus", o.CPU,
-		"--extra-config=apiserver.Authorization.Mode=RBAC",
-		"--extra-config=apiserver.GenericServerRunOptions.CorsAllowedOriginList='.*'",
-		"--extra-config=controller-manager.ClusterSigningCertFile='/var/lib/localkube/certs/ca.crt'",
-		"--extra-config=controller-manager.ClusterSigningKeyFile='/var/lib/localkube/certs/ca.key'",
-		"--extra-config=apiserver.admission-control='LimitRanger,ServiceAccount,DefaultStorageClass,MutatingAdmissionWebhook,ValidatingAdmissionWebhook,ResourceQuota'",
+		"--extra-config=apiserver.authorization-mode=RBAC",
+		"--extra-config=apiserver.cors-allowed-origins='http://*'",
```
a-thaler: Will it work for https?

pkg/kyma/cmd/provision/minikube.go
sjanota: What I'm sure is it has changed in new minikube. What I can do is write it the same way installation scripts are written (they work on labels which haven't changed).

pkg/kyma/cmd/provision/minikube.go
sjanota: I'll add it back

pkg/kyma/cmd/provision/minikube.go
sjanota: Shouldn't be a problem, apiserver is not hidden by SSL anyway. That's how Kyma is being installed right now.

pkg/kyma/cmd/provision/minikube.go
sjanota: I thought the problem was with renaming in minikube, but the problem was with IsPodReady function. Fix is en route.

a-thaler: jihaa, thanks

Issue #23, opened 2019-02-11
Check dependencies up to minor version, not patch


sjanota:
There is upcoming feature in Kyma: https://github.com/kyma-project/kyma/issues/1923

It will change how dependencies versions (minikube specifically) are verified by installation scripts:
 * They will not verify patch version, only minor.
 * They will not crash if wrong version is used, only print warning

1. We must definitely add it for minikube to be compatible.  
2. We need to discuss if adding it for kubernetes and other tools makes sense. @a-thaler 
3. There surely is a library for handling semver out in the wild. Probably we could utilise one.

PR #27, opened 2019-02-18
Minikube versioning

sjanota:
Changes: 
* Be more liberal when checking minikube version
* Cleanup error logging in minikube command


sjanota:
#23

PR #28, opened 2019-02-18
Allow to change installer image for local development

sjanota:
Two new flags are available on `kyma install --local`:
* `--installer-version` allows to set image tag
* `--installer-dir` allows to set installer image path

### DIFF
--- a/pkg/kyma/cmd/install.go
+++ b/pkg/kyma/cmd/install.go
@@ -149,8 +149,10 @@ func validateFlags(o *InstallOptions) error {
 		if _, err := os.Stat(filepath.Join(o.LocalSrcPath, "installation", "resources")); err != nil {
 			return fmt.Errorf("Configured 'src-path=%s' seems to not point to a Kyma repository, please verify if your repository contains a folder 'installation/resources'", o.LocalSrcPath)
 		}
+		
+		// This is to help developer and use appropriate repository if PR image is provided
 		if o.LocalInstallerDir == "" && strings.HasPrefix(o.LocalInstallerVersion, "PR-") {
-			o.LocalInstallerDir = " "
+			o.LocalInstallerDir = "eu.gcr.io/kyma-project/pr"
 		}
 	} else {
 		if o.LocalSrcPath != "" {
--- a/pkg/kyma/cmd/install.go
+++ b/pkg/kyma/cmd/install.go
@@ -81,7 +81,7 @@ func (o *InstallOptions) Run() error {
 	}
 
 	s := o.NewStep(fmt.Sprintf("Checking requirements"))
-	err := checkReqs(o)
+	err = checkReqs(o)
 	if err != nil {
 		s.Failure()
 		return err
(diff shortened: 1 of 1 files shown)


pkg/kyma/cmd/install.go
```diff
@@ -125,34 +134,42 @@ func (o *InstallOptions) Run() error {
...
+			o.LocalSrcPath = filepath.Join(goPath, "src", "github.com", "kyma-project", "kyma")
 		}
 		if _, err := os.Stat(o.LocalSrcPath); err != nil {
 			return fmt.Errorf("Configured 'src-path=%s' does not exist, please check if you configured a valid path", o.LocalSrcPath)
 		}
 		if _, err := os.Stat(filepath.Join(o.LocalSrcPath, "installation", "resources")); err != nil {
 			return fmt.Errorf("Configured 'src-path=%s' seems to not point to a Kyma repository, please verify if your repository contains a folder 'installation/resources'", o.LocalSrcPath)
 		}
+		if o.LocalInstallerDir == "" && strings.HasPrefix(o.LocalInstallerVersion, "PR-") {
+			o.LocalInstallerDir = " "
```
a-thaler: Sounds like hidden magic, why only for a PR and why a space? Whatever the reason is, I guess no one else then you will understand it when reading the code...

pkg/kyma/cmd/install.go
```diff
@@ -65,14 +67,21 @@ The command will:
 	cmd.Flags().StringVarP(&o.Domain, "domain", "d", "kyma.local", "domain to use for installation")
 	cmd.Flags().BoolVarP(&o.Local, "local", "l", false, "Install from sources")
 	cmd.Flags().StringVarP(&o.LocalSrcPath, "src-path", "", "", "Path to local sources to use")
+	cmd.Flags().StringVarP(&o.LocalInstallerVersion, "installer-version", "", "", "Version of installer docker image to use while building locally")
+	cmd.Flags().StringVarP(&o.LocalInstallerDir, "installer-dir", "", "", "Directory of installer docker image to use while building locally")
```
a-thaler: It is not the directory to the image but to the Dockerfile, isn't it?

pkg/kyma/cmd/install.go
sjanota: It was supposed to be `eu.gcr.io/.../pr` not space :D. It's a shortcut so you don't have to provide `--installer-dir` when running Kyma from PR.

pkg/kyma/cmd/install.go
sjanota: No, it's not path do Dockerfile, it's path to image. For example `eu.gcr.io/kyma-project/develop`, but you could provide your own if you run from PR or your dockerhub. Those two flags map 1-1 to ARGs from [kyma-installer Dockerfile](https://github.com/kyma-project/kyma/blob/master/tools/kyma-installer/kyma.Dockerfile)

a-thaler: ok fine. Maybe path is then misleading and it is more a docker repository.. However, I'm fine :)

pkg/kyma/cmd/install.go
a-thaler: What means "supposed", will you change it?

sjanota: Yes

sjanota: Is it better now?

PR #29, opened 2019-02-21
consistent requirement checks

a-thaler:
- Adjusted requirement checks to use spinner logs for warnings
- Adjusted minikube version warning to include the version numbers as it was hard to guess which version es recommended
- Moved kubectl helper file into dedicated package
- Used verbose flag for kubectl and minikube commands
- Moved gliph in spinner message to start of line
- introduced new gliphs for spinner
- Have a better vmdriver default for windows
- Using simple spinner only for windows

### DIFF
--- a/internal/kubectl/kubectl.go
+++ b/internal/kubectl/kubectl.go
@@ -1,7 +1,8 @@
-package internal
+package kubectl
 
 import (
 	"bufio"
+	"errors"
 	"fmt"
 	"os/exec"
 	"regexp"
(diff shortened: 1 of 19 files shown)


internal/kubectl/kubectl.go
```diff
@@ -131,11 +133,11 @@ func IsPodReady(namespace string, labelName string, labelValue string) (bool, er
 	return true, nil
 }
 
-//CheckKubectlVersion assures that the kubectl version used is compatible
-func CheckKubectlVersion() error {
```
sjanota: How about using `semver` here? Like I did for minikube: https://github.com/kyma-incubator/kyma-cli/blob/master/internal/minikube/minikube.go#L48

a-thaler: I was too lazy to change it, will do it

a-thaler: I did it

PR #33, opened 2019-03-05
Simplify glyphs

sjanota:
Emojis used on OSX caused some problems as spinners were not able to calculate number of terminal characters used for rendering. Example:
screen shot 2019-03-05 at 18 45 02

Here is my proposal how to address the problem. Simply:
- Do not use emojis on OSX, but use the same set of characters throughout all systems
- Add colours to plain characters

How it looks now:
screen shot 2019-03-05 at 19 20 27
screen shot 2019-03-05 at 19 21 09

Last, but not least: it's a matter of taste, but emojis are a little flashy.


sjanota:
Damn, I was too quick. @venturasr could you confirm that it's working fine on Windows? I'll do a follow up if it doesn't.

PR #34, opened 2019-03-06
Do not log sensitive data in logs

### LINKED ISSUE #26: Logs with sensitive data are logged in case of installation error
If there is an error during installation installer logs are printed to the console. As all overrides are written there they may contain sensitive data. This is security vulnerability and prevents kyma-cli from being used in production environment. There is a plan to add additional error log to `Installation` CRD status. As soon as it is done we need to utilise it and remove printing installer logs.

sjanota:
This PR changes the behaviour of install command regarding watching installation. The algorithm is taken from `is-installed.sh`. Following changes are introduced:
- Do not print installer logs - they may contains sensitive information
- On error print a line explaining how user cen fetch the logs himself
- There is a possibility to specify watch timeout
- Installation goes indefinitely unless timeout is specified
- On timeout `errorLog` from installation status is printed

Fixes #26


a-thaler:
The uninstall is doing exactly the same, could you please apply the same fix

PR #48, opened 2019-03-20
Added timeout to "version" command

igor-karpukhin:
"version" command won't freeze if minikube is not running


igor-karpukhin:
I've added a message when CLI can't get the Kyma version. @a-thaler please review.

PR #57, opened 2019-04-09
update to kyma 0.9.1

a-thaler:
Updated to kyma 0.9.1
Adjusted disk size setting for minikube
Adjusted domains for hosts file
Fixed minikube IP problem: https://github.com/kyma-project/kyma/issues/3651
Configuration of local helm with certificates of tiller installation


pkg/kyma/cmd/install/cmd.go
```diff
@@ -193,6 +202,64 @@ func (cmd *command) installTiller() error {
...
+}
+
+func writeHelmFile(data map[interface{}]interface{}, helmData string, helmHome string, filename string) error {
+	value, ok := data[helmData].(string)
+	if !ok {
+		return fmt.Errorf("unable to get %s from helm secret data", filename)
+	}
+	valueDecoded, err := base64.StdEncoding.DecodeString(value)
+
+	err = ioutil.WriteFile(filepath.Join(helmHome, filename), valueDecoded, 0777)
```
igor-karpukhin: @a-thaler Why would you give execution permission to the helm file?

a-thaler: fixed

PR #65, opened 2019-05-02
fixed helm init problem

a-thaler:
By a silly mistake, the helm tls configuration was not applied to your local helm installation anymore. This PR will fix it

### DIFF
--- a/pkg/kyma/cmd/install/cmd.go
+++ b/pkg/kyma/cmd/install/cmd.go
@@ -230,7 +230,7 @@ func (cmd *command) configureHelm() error {
 
 	data, ok := cfg["data"].(map[interface{}]interface{})
 	if !ok {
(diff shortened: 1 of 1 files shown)


pkg/kyma/cmd/install/cmd.go
```diff
@@ -228,7 +230,7 @@ func (cmd *command) configureHelm() error {
 
 	data, ok := cfg["data"].(map[interface{}]interface{})
 	if !ok {
-		return fmt.Errorf("unable to get data from helm secret")
+		return fmt.Errorf("Unable to get data from helm secret")
```
igor-karpukhin: I think It is better to keep log messages in the lower case all the time, since it is easier to find the specific message on production, in the stackdriver for example.

PR #75, opened 2019-05-16
Refactored editing /etc/hosts file

igor-karpukhin:
CLI will now explicitly ask a user to get permissions to edit `/etc/hosts` if CLI is running with `sudo`. It will also print the command which will add Kyma DNS names to `/etc/hosts` if CLI was launched without `sudo` and a user has denied providing `sudo` password.


pkg/kyma/cmd/provision/minikube/hosts_darwin.go
```diff
@@ -11,18 +16,65 @@ const hostsFile = "/etc/hosts"
 
 const defaultVMDriver = vmDriverHyperkit
 
+func isWithSudo() bool {
```
clebs: Although this is correct, to me it looks more readable as:
```
func isWithSudo() bool {
   return os.Getenv("SUDO_UID") != ""
}
```
It avoids having to go through the branches of the `if-else` and putting the pieces together in my head.

pkg/kyma/cmd/provision/minikube/hosts_darwin.go
```diff
@@ -11,18 +16,65 @@ const hostsFile = "/etc/hosts"
...
+		return true
+	}
+	return false
+}
+
+func promptUser() bool {
+	for {
+		fmt.Print("Type [y/n]: ")
+		var res string
+		_, err := fmt.Scanf("%s", &res)
```
clebs: Since `res` is already declared outside, this can be expressed as:
```
if _, err := fmt.Scanf("%s", &res); err != nil {
  return false
}
...
```
Saves 1 line and err only exists in the scope of that if statement.

pkg/kyma/cmd/provision/minikube/hosts_darwin.go
```diff
@@ -11,18 +16,65 @@ const hostsFile = "/etc/hosts"
...
+		if err != nil {
+			return false
+		}
+		switch res {
+		case "yes":
```
clebs: Is it correct to check for the words "yes" and "no" when the prompt asks for "y" or "n"?
I usually only type "y" in these cases.

pkg/kyma/cmd/provision/minikube/hosts_darwin.go
```diff
@@ -11,18 +16,65 @@ const hostsFile = "/etc/hosts"
...
+		if !promptUser() {
+			notifyUserFunc()
+			return nil
+		}
+	}
 	_, err := internal.RunCmd("sudo", []string{"/bin/sh", "-c", "sed -i '' \"/" + o.Domain + "/d\" " + hostsFile})
 	if err != nil {
-		return err
+		notifyUserFunc()
+		return nil
```
clebs: Is it ok to just ignore the error here? Maybe it is not permission related.

pkg/kyma/cmd/provision/minikube/hosts_darwin.go
igor-karpukhin: Fixed

pkg/kyma/cmd/provision/minikube/hosts_darwin.go
igor-karpukhin: Fixed

pkg/kyma/cmd/provision/minikube/hosts_darwin.go
igor-karpukhin: Fixed

pkg/kyma/cmd/provision/minikube/hosts_darwin.go
igor-karpukhin: As discussed with @a-thaler.

igor-karpukhin:
@clebs I'm wondering if we really need this file `hosts_others.go`? It has the same contents as `hosts_darwin`.

clebs:
Good question @igor-karpukhin... I don't think it is necessary. We could have just `hosts_windows.go` with the build tag for windows and another `hosts.go` which builds for anything that is not windows.

PR #84, opened 2019-05-20
Fixed the way of removing existing aliases from /etc/hosts

igor-karpukhin:
Fixed the way of removing existing aliases from /etc/hosts. Removed sh execution `/bin/sh`

### DIFF
--- a/pkg/kyma/cmd/provision/minikube/common.go
+++ b/pkg/kyma/cmd/provision/minikube/common.go
@@ -36,11 +36,14 @@ func promptUser() bool {
 
 func addDevDomainsToEtcHostsOSSpecific(o *MinikubeOptions, s step.Step, hostAlias string) error {
 	notifyUserFunc := func(err error) {
-		s.LogInfof("Error: %s. Execute the following command manually to add mappings: sudo sed -i '' \"/"+o.Domain+"/d\" "+hostsFile+" && echo '%s' | sudo tee -a /etc/hosts\r\n", err.Error(), hostAlias)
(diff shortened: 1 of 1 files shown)


pkg/kyma/cmd/provision/minikube/common.go
```diff
@@ -35,36 +35,43 @@ func promptUser() bool {
...
 
 func addDevDomainsToEtcHostsOSSpecific(o *MinikubeOptions, s step.Step, hostAlias string) error {
-	notifyUserFunc := func() {
-		s.LogInfof("Execute the following command manually to add mappings: sudo sed -i '' \"/"+o.Domain+"/d\" "+hostsFile+" && echo '%s' | sudo tee -a /etc/hosts\r\n", hostAlias)
+	notifyUserFunc := func(err error) {
+		if err != nil {
+			s.LogInfof("Error: %s", err.Error())
+		}
+		s.LogInfof("Execute the following command manually to add domain entries: sudo sed -i '' \"/"+o.Domain+"/d\" "+hostsFile+" && echo '%s' | sudo tee -a /etc/hosts\r\n", hostAlias)
 		return
```
clebs: I know you did not change this line, but this `return` statement is pointless, it can just be deleted.

The function does not return anything and it is also not breaking out of any control structure.

igor-karpukhin: Yes, but in general, I think it is a good practice to mark the exit explicitly. Anyway, I removed that for now.

PR #95, opened 2019-05-28
fix hosts lines on windows

### LINKED ISSUE #93: Windows hosts should be separated
In windows /etc/hosts file can't read more than 9 hostnames per ip. That's why , when outputting the steps for windows users during installation, there should be several lines with the same ip and different hostnames. The related kyma installation code is here:

https://github.com/kyma-project/kyma/blob/ea007479b7b3f229b6020c319a91bcc946deeac1/installation/scripts/minikube.ps1#L89

atakanyenel:
**Description**

- Windows hosts file does not accept more than 9 hostname per line. This PR divides the `hostAlias` variable to 7 hostnames per line for windows. It doesn't touch *nix systems.
- It fixes a typo `minukube` -> `minikube`
- ...

**Related issue(s)**

fixes #93

### DIFF
--- a/pkg/kyma/cmd/provision/minikube/hosts_windows.go
+++ b/pkg/kyma/cmd/provision/minikube/hosts_windows.go
@@ -4,6 +4,7 @@ package minikube
 
 import (
 	"fmt"
+	"strings"
 
 	"github.com/kyma-project/cli/internal/step"
 )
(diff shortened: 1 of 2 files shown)


pkg/kyma/cmd/provision/minikube/hosts_windows.go
```diff
@@ -4,14 +4,34 @@ package minikube
...
 
 func addDevDomainsToEtcHostsOSSpecific(o *MinikubeOptions, s step.Step, hostAlias string) error {
 
 	s.LogErrorf("Please add these lines to your " + hostsFile + " file:")
-	fmt.Println(hostAlias)
+	hostsArray := strings.Split(hostAlias, " ")
+	ip := hostsArray[0]
+	hostsPerLine := split(hostsArray[1:], 7) // 7 hostnames per line
+
+	for k := 0; k < len(hostsPerLine); k++ {
```
clebs: This looks like a lot of code to do a very simple splitting of slices. You can achieve the same with the following code without the `split` function at all:
```go
hostsArray = hostsArray[1:]
for len(hostsArray) > 0 {
	chunkLen := 7
	if len(hostsArray) < chunkLen {
		chunkLen = len(hostsArray)
	}
	fmt.Printf("%s\n", strings.Join(hostsArray[:chunkLen], " "))
	chunks = chunks[chunkLen:]
}
```

Basically you strip out and print chunks of your desired size until the slice is empty.

Issue #80, opened 2019-05-20
Code quality for release


clebs:
**Description**

The source code for CLI should follow standards and common best practices for Go code:
- Package cohesion: Go packages should not blindly export structure and functionality, but instead have an intentional and meaningful public API.
- Unit tests: All go packages should have unit tests (if possible).

**Reasons**
- Improve overall code quality.
- Improve developer confidence in a new release.
- Avoid unintentional behavior breaking and/or security issues.
- Introduce best practices.
- Make reviewing easier.

PR #99, opened 2019-06-03
Quality improvements

clebs:
**Description**

- Refactor packages to improve cohesion, encapsulation and avoid bad usages.
- Add unit tests where possible.
- Correct dangerous slice and varargs usages.

**Related issue(s)**
Belongs to #80


internal/helm/helm.go
```diff
@@ -124,13 +124,15 @@ func setupTillerConnection(settings *environment.EnvSettings, config *rest.Confi
 }
 
 func (c *Client) Close() {
-	if c.forwarder != nil {
+	if c != nil && c.forwarder != nil {
```
igor-karpukhin: Why check the pointer to the object?

clebs: It is not really necessary, I agree that it is not good to check this in all pointer receiver methods.

But in this specific method I thought it is a good idea, because often this method will be called during cleanup or disposal of resources and it is likely to happen that `Close()` is called after the receiver is set to `nil`.

Issue #105, opened 2019-06-05
windows powershell renders symbols incorrectly


atakanyenel:
![cli-output](https://user-images.githubusercontent.com/7947572/58939238-16823200-8777-11e9-8a63-e02aab0955ac.png)

PR #114, opened 2019-06-07
removing console colors from non-darwin systems

a-thaler:
**Description**
Fixes encoding problems on non-darwin systems by disabling coloring and using plain ascii chars always

Fixes https://github.com/kyma-project/cli/issues/105

### DIFF
--- a/internal/step/spinner.go
+++ b/internal/step/spinner.go
@@ -9,6 +9,7 @@ import (
 	"time"
 
 	"github.com/briandowns/spinner"
+	"github.com/fatih/color"
 )
 
 func newStepWithSpinner(msg string) Step {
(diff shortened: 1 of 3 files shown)


internal/step/spinner.go
```diff
@@ -75,11 +76,11 @@ func (s *stepWithSpinner) LogInfof(format string, args ...interface{}) {
 }
 
 func (s *stepWithSpinner) LogError(msg string) {
-	s.logTof(os.Stderr, warningGlyph+msg)
+	s.logTof(os.Stderr, color.YellowString(warningGlyph)+msg)
```
igor-karpukhin: Does windows terminal correctly display colorful output?

a-thaler: No, that's why I moved it into the non-windows code base

PR #115, opened 2019-06-07
update to minikube 1.1.0

a-thaler:
**Description**
There is a new minikube version available, let's update


a-thaler:
I will close it for now as it must be aligned with kyma and test-infra

PR #124, opened 2019-06-13
Fixed resource patching method

igor-karpukhin:
Fixed the way CLI patches resources. If a resource is not presented it won't be patched now.

### DIFF
--- a/pkg/kyma/cmd/install/cmd.go
+++ b/pkg/kyma/cmd/install/cmd.go
@@ -886,8 +886,7 @@ func (cmd *command) patchMinikubeIP() error {
 
 	for k, v := range patchMap {
 		for _, pData := range v {
-			_, err := cmd.Kubectl().RunCmd("-n", "kyma-installer", "get", k)
-			if err != nil {
+			if _, err := cmd.Kubectl().RunCmd("-n", "kyma-installer", "get", k); err != nil {
 				continue
 			}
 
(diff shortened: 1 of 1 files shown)


pkg/kyma/cmd/install/cmd.go
```diff
@@ -886,7 +886,13 @@ func (cmd *command) patchMinikubeIP() error {
 
 	for k, v := range patchMap {
 		for _, pData := range v {
-			_, err := cmd.Kubectl().RunCmd("-n", "kyma-installer", "patch", k, "--type=json",
+			_, err := cmd.Kubectl().RunCmd("-n", "kyma-installer", "get", k)
```
clebs: If you are not using the result of the method and only checking the error, why not:
```go
if _, err := cmd.Kubectl().RunCmd("-n", "kyma-installer", "get", k); err != nil {
  continue
}
```

igor-karpukhin: Fixed, thanks :)

clebs: Also just swallowing the error and continuing could be hiding a serious problem. Maybe it would be good to have a similar pattern as some stdlib packages use such as bufio.Writer, net/http.
Here the pattern is explained and there are code samples: https://blog.golang.org/errors-are-values

igor-karpukhin: Fixed. I added one additional check, so the resource won't be patched if it's not installed

Issue #177, opened 2019-07-15
Improve inline documentation for CLI


bszwarc:
**Description**

This is a follow-up issue for #153. We should have a look at the following inline entries:
- command descriptions
- error descriptions
- system notifications
- any other descriptions

**Reasons**

Inline documentation should be unified.

PR #180, opened 2019-07-16
Improve inline documentation for CLI

bszwarc:
**Description**

- Improve inline documentation for commands
- Improve error messages

**Related issue(s)**
See also #177


internal/kube/client.go
```diff
@@ -82,7 +82,7 @@ func (c *client) Octopus() octopus.OctopusInterface {
 func (c *client) IsPodDeployed(namespace, name string) (bool, error) {
 	_, err := c.Static().CoreV1().Pods(namespace).Get(name, metav1.GetOptions{IncludeUninitialized: true})
 	if err != nil {
-		if strings.Contains(err.Error(), "not found") {
+		if strings.Contains(err.Error(), "Not found") {
```
igor-karpukhin: That won't work. The error messages coming from a kubernetes API are in lower case. Either leave it in lower case or all checks should be wrapped with `strings.ToLower()`

internal/kube/client.go
```diff
@@ -104,7 +104,7 @@ func (c *client) IsPodDeployedByLabel(namespace, labelName, labelValue string) (
 func (c *client) WaitPodStatus(namespace, name string, status corev1.PodPhase) error {
 	for {
 		pod, err := c.Static().CoreV1().Pods(namespace).Get(name, metav1.GetOptions{IncludeUninitialized: true})
-		if err != nil && !strings.Contains(err.Error(), "not found") {
+		if err != nil && !strings.Contains(err.Error(), "Not found") {
```
igor-karpukhin: Same thing here

pkg/kyma/cmd/install/cmd.go
```diff
@@ -830,16 +830,16 @@ func (cmd *command) releaseFile(path string) string {
...
 		return nil
 	}
 	minikubeIP = strings.TrimSpace(minikubeIP)
 
 	for k, v := range patchMap {
 		for _, pData := range v {
 			if _, err := cmd.Kubectl().RunCmd("-n", "kyma-installer", "get", k); err != nil {
-				if strings.Contains(err.Error(), "not found") {
-					cmd.CurrentStep.LogInfof("resource '%s' not found, won't be patched", k)
+				if strings.Contains(err.Error(), "Not found") {
```
igor-karpukhin: Here is the same thing

internal/kube/client.go
bszwarc: got it - I will change it :)

pkg/kyma/cmd/install/cmd.go
bszwarc: Will change :)

pkg/kyma/cmd/uninstall/cmd.go
```diff
@@ -182,17 +182,17 @@ func (cmd *command) deleteTiller() error {
 	}
 
 	err = cmd.K8s.Static().RbacV1().RoleBindings("kube-system").Delete("tiller-certs", &metav1.DeleteOptions{})
-	if err != nil && !strings.Contains(err.Error(), "not found") {
+	if err != nil && !strings.Contains(err.Error(), "Not found") {
```
igor-karpukhin: Same thing here as above

pkg/kyma/cmd/uninstall/cmd.go
```diff
@@ -182,17 +182,17 @@ func (cmd *command) deleteTiller() error {
...
 		return err
 	}
 
 	err = cmd.K8s.Static().RbacV1().Roles("kube-system").Delete("tiller-certs-installer", &metav1.DeleteOptions{})
-	if err != nil && !strings.Contains(err.Error(), "not found") {
+	if err != nil && !strings.Contains(err.Error(), "Not found") {
```
igor-karpukhin: Same thing here as above

pkg/kyma/cmd/uninstall/cmd.go
```diff
@@ -182,17 +182,17 @@ func (cmd *command) deleteTiller() error {
...
 		return err
 	}
 
 	err = cmd.K8s.Static().CoreV1().ServiceAccounts("kube-system").Delete("tiller-certs-sa", &metav1.DeleteOptions{})
-	if err != nil && !strings.Contains(err.Error(), "not found") {
+	if err != nil && !strings.Contains(err.Error(), "Not found") {
```
igor-karpukhin: Same thing here as above

internal/kyma_test.go
```diff
@@ -17,26 +17,26 @@ func TestRunCmd(t *testing.T) {
...
 		},
 		{
 			name:           "Incorrect command",
-			description:    "test a command that exits with error",
-			cmd:            []string{"ehco", "this is wrongly spelled"},
+			description:    "Checks if a command is correct. If not, exists and returns an error.",
+			cmd:            []string{"ehco", "This is spelled incorrectly"},
 			expectedOutput: "",
-			expectedErr:    errors.New("Failed executing command 'ehco [this is wrongly spelled]' with output '' and error message 'exec: \"ehco\": executable file not found in $PATH'"),
+			expectedErr:    errors.New("Executing command 'ehco [this is wrongly spelled]' failed with output '' and error message 'exec: \"ehco\": executable file not found in $PATH'"),
```
igor-karpukhin: `"this is wrongly spelled"` should be also replaced in the `expectedErr` field within the `[]`

bszwarc: Done.

Issue #188, opened 2019-07-29
Upgrade CLI Kubernetes API to 1.15


clebs:
**Description**

Upgrade the CLI to be compatible with the Kubernetes API 1.15, since Kyma will be upgraded to that version.

**Reasons**

Compatibility with Kyma clusters.

clebs:
Refatored the CLI octopus rest client to use the client-go one, as the standard API does, getting rid of the dependency to `controller-runtime` and improving the client.


Unfortunately, upgrading the CLI  to k8s 1.15 is still not possible because octopus depends on `controller-runtime` (which locks the k8s API to 1.12.3) and we import it.

Tested the CLI on a k8s 1.15 cluster and it is still compatible.

PR #198, opened 2019-08-08
octopus: refactor k8s rest client

clebs:
**Description**

- Attempted upgrade to k8s client-go 1.15 but octopus is blocking it.
- Octopus k8s rest client now uses the same approach as the standard k8s API.
- Remove dependency to `controller-runtime`.
- Updated some dependencies.
- This change makes future upgrades easier.

**Related issue(s)**
#188

### DIFF
--- a/internal/kube/client.go
+++ b/internal/kube/client.go
@@ -53,7 +53,7 @@ func NewFromConfigWithTimeout(url, file string, t time.Duration) (KymaKube, erro
 		return nil, err
 	}
 
-	octClient, err := octopus.NewOctopusRESTClient(2 * time.Second)
+	octClient, err := octopus.NewFromConfig(config)
 	if err != nil {
 		return nil, err
 	}
--- a/pkg/api/octopus/client.go
+++ b/pkg/api/octopus/client.go
@@ -1,84 +1,107 @@
 package octopus
 
 import (
-	"context"
-	"time"
-
 	"github.com/kyma-incubator/octopus/pkg/apis"
 	oct "github.com/kyma-incubator/octopus/pkg/apis/testing/v1alpha1"
-	"k8s.io/apimachinery/pkg/types"
+	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
+	"k8s.io/apimachinery/pkg/runtime/serializer"
 	"k8s.io/client-go/kubernetes/scheme"
-	"sigs.k8s.io/controller-runtime/pkg/client"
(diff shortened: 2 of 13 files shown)


internal/kube/client.go
```diff
@@ -53,7 +53,7 @@ func NewFromConfigWithTimeout(url, file string, t time.Duration) (KymaKube, erro
 		return nil, err
 	}
 
-	octClient, err := octopus.NewOctopusRESTClient(2 * time.Second)
+	octClient, err := octopus.NewFromConfig(config)
```
igor-karpukhin: Do we have a default timeout when creating a client in this way?

pkg/api/octopus/client.go
```diff
@@ -1,84 +1,107 @@
...
+
 type OctopusInterface interface {
-	ListTestDefinitions() (*oct.TestDefinitionList, error)
-	ListTestSuites() (*oct.ClusterTestSuiteList, error)
-	CreateTestSuite(cts *oct.ClusterTestSuite) error
-	DeleteTestSuite(cts *oct.ClusterTestSuite) error
-	GetTestSuiteByName(name string) (*oct.ClusterTestSuite, error)
+	ListTestDefinitions(opts metav1.ListOptions) (result *oct.TestDefinitionList, err error)
+	ListTestSuites(opts metav1.ListOptions) (result *oct.ClusterTestSuiteList, err error)
+	CreateTestSuite(cts *oct.ClusterTestSuite) (result *oct.ClusterTestSuite, err error)
```
igor-karpukhin: Is there a reason returning created test suite?

internal/kube/client.go
clebs: Yes, the default timeout is 30 seconds.
See: https://github.com/kyma-project/cli/blob/e6ae0cc6e8bb7cf8262fac15682a806e519cc609/internal/kube/client.go#L32

pkg/api/octopus/client.go
clebs: To be consistent with the standard API.

internal/kube/client.go
igor-karpukhin: Thanks. Would you consider to keep the timeout within 2-5 sec range?

clebs: 30 sec is just a reasonable default, but of course different commands will have different needs.
That is why each command can specify a custom timeout when creating the client with `NewFromConfigWithTimeout()`.
If you think a command needs a 2sec timeout then just create the clientset with that timeout.

PR #215, opened 2019-08-28
If core console not found , don't abort the instalation

dbadura:
**Description**
CI fails on Kyma `Compass` project, because we don't use`core-console` virtualservice.

- print message  `Kyma console: 		not found` instead of exiting when `core-console` not found.

**Related issue(s)**

### DIFF
--- a/cmd/kyma/install/cmd.go
+++ b/cmd/kyma/install/cmd.go
@@ -743,10 +743,12 @@ func (cmd *command) printSummary() error {
 	switch {
 	case apiErrors.IsNotFound(err):
 		consoleURL = "not installed"
+	case err != nil:
+		return err
 	case vs != nil && vs.Spec != nil && len(vs.Spec.Hosts) > 0:
 		consoleURL = fmt.Sprintf("https://%s", vs.Spec.Hosts[0])
 	default:
-		return err
+		return errors.New("Console host could not be obtained.")
 	}
 
 	clusterInfo, err := cmd.Kubectl().RunCmd("cluster-info")
--- a/cmd/kyma/install/cmd.go
+++ b/cmd/kyma/install/cmd.go
@@ -742,11 +742,11 @@ func (cmd *command) printSummary() error {
 	vs, err := cmd.K8s.Istio().NetworkingV1alpha3().VirtualServices("kyma-system").Get("core-console", metav1.GetOptions{})
 	switch {
 	case apiErrors.IsNotFound(err):
-		consoleURL = "not found"
-	case err != nil:
-		return err
-	case vs != nil:
+		consoleURL = "not installed"
+	case vs != nil && vs.Spec != nil && len(vs.Spec.Hosts) > 0:
 		consoleURL = fmt.Sprintf("https://%s", vs.Spec.Hosts[0])
+	default:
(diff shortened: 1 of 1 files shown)


cmd/kyma/install/cmd.go
```diff
@@ -738,9 +738,15 @@ func (cmd *command) printSummary() error {
 		return err
 	}
 
+	var consoleURL string
 	vs, err := cmd.K8s.Istio().NetworkingV1alpha3().VirtualServices("kyma-system").Get("core-console", metav1.GetOptions{})
-	if err != nil {
+	switch {
+	case apiErrors.IsNotFound(err):
+		consoleURL = "not found"
```
aszecowka: maybe "not installed"?

cmd/kyma/install/cmd.go
```diff
@@ -738,9 +738,15 @@ func (cmd *command) printSummary() error {
...
+	case err != nil:
 		return err
+	case vs != nil:
```
aszecowka: what if vs.Spec.Hosts is empty? probably there will be panic. What message should be displayed then, "no information available"?

cmd/kyma/install/cmd.go
```diff
@@ -752,7 +758,7 @@ func (cmd *command) printSummary() error {
 	fmt.Println(clusterInfo)
 	fmt.Println()
 	fmt.Printf("Kyma is installed in version %s\n", v)
-	fmt.Printf("Kyma console:\t\thttps://%s\n", vs.Spec.Hosts[0])
+	fmt.Printf("Kyma console:s\t\t%s\n", consoleURL)
```
suleymanakbas91: is this `s` a typo? 
`Kyma console:s		https://console.kyma.local`

cmd/kyma/install/cmd.go
```diff
@@ -738,8 +738,14 @@ func (cmd *command) printSummary() error {
...
 
+	var consoleURL string
 	vs, err := cmd.K8s.Istio().NetworkingV1alpha3().VirtualServices("kyma-system").Get("core-console", metav1.GetOptions{})
-	if err != nil {
+	switch {
+	case apiErrors.IsNotFound(err):
+		consoleURL = "not installed"
+	case vs != nil && vs.Spec != nil && len(vs.Spec.Hosts) > 0:
+		consoleURL = fmt.Sprintf("https://%s", vs.Spec.Hosts[0])
+	default:
```
clebs: This logic looks sound but if the call is correct (err is nil), but there are no hosts on the spec (for whatever reason), you would be returning a nil error and it would be impossible to know what happened and very hard to debug.
Also, it would not work if both `vs` and `err` are not nil. Usually it is not the case, but we can not assume.

I would suggest the following logic to cover those 2 corner cases:
```go
switch {
	case apiErrors.IsNotFound(err):
		consoleURL = "not installed"
	case err != nil:
		return err
	case vs != nil && vs.Spec != nil && len(vs.Spec.Hosts) > 0:
                consoleURL = fmt.Sprintf("https://%s", vs.Spec.Hosts[0])
       default:
                return errors.New("Console host could not be obtained.")
}
```

PR #221, opened 2019-08-30
Source option added

igor-karpukhin:
* Removed the following flags: `--installer-version`, `--installer-dir`
* Flag `--release` replaces with `--source`
* `--source` flag is used to configure the way to install kyma. `local` value will install kyma from local sources, any tag or commit name (ex: `1.3.0`, `63f27f76`), or a specific installer image path (ex: `eu.gcr.io/kyma-project/kyma-installer:1.3.0`)

### DIFF
--- a/cmd/kyma/install/cmd.go
+++ b/cmd/kyma/install/cmd.go
@@ -13,6 +13,7 @@ import (
 	"strings"
 	"time"
 
+	"github.com/Masterminds/semver"
 	"github.com/kyma-project/cli/internal/step"
 
 	"github.com/kyma-project/cli/cmd/kyma/version"
@@ -52,11 +53,12 @@ type clusterInfo struct {
 }
 
 const (
-	sleep                  = 10 * time.Second
-	releaseSrcUrlPattern   = "https://raw.githubusercontent.com/kyma-project/kyma/%s/%s"
-	releaseResourcePattern = "https://raw.githubusercontent.com/kyma-project/kyma/%s/installation/resources/%s"
-	registryMasterPattern  = "eu.gcr.io/kyma-project/develop/kyma-installer:master-%s"
-	localDomain            = "kyma.local"
+	sleep                       = 10 * time.Second
+	releaseSrcUrlPattern        = "https://raw.githubusercontent.com/kyma-project/kyma/%s/%s"
+	releaseResourcePattern      = "https://raw.githubusercontent.com/kyma-project/kyma/%s/installation/resources/%s"
+	registryReleaseImagePattern = "eu.gcr.io/kyma-project/kyma-installer:%s"
+	registryMasterImagePattern  = "eu.gcr.io/kyma-project/develop/kyma-installer:%s"
+	localDomain                 = "kyma.local"
 )
 
 //NewCmd creates a new kyma command
@@ -78,7 +80,6 @@ func NewCmd(o *Options) *cobra.Command {
 		Aliases: []string{"i"},
 	}
 
-	cobraCmd.Flags().StringVarP(&o.ReleaseConfig, "config", "c", "", "URL or path to the Installer configuration yaml file")
 	cobraCmd.Flags().BoolVarP(&o.NoWait, "noWait", "n", false, "Do not wait for the Kyma installation to complete")
 	cobraCmd.Flags().StringVarP(&o.Domain, "domain", "d", localDomain, "Domain used for installation")
 	cobraCmd.Flags().StringVarP(&o.TLSCert, "tlsCert", "", "", "TLS certificate for the domain used for installation")
@@ -111,7 +112,7 @@ func (cmd *command) Run() error {
 	if cmd.opts.Local {
 		s.LogInfof("Installing Kyma from local path: '%s'", cmd.opts.LocalSrcPath)
 	} else {
-		s.LogInfof("Installing Kyma in version '%s'", cmd.opts.ReleaseVersion)
+		s.LogInfof("Installing Kyma in version '%s'. Config version '%s'", cmd.opts.ReleaseVersion, cmd.opts.ConfigVersion)
 	}
 	s.Successf("Installation source checked")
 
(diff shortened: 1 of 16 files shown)


cmd/kyma/install/cmd.go
```diff
@@ -205,7 +206,9 @@ func (cmd *command) Run() error {
 }
 
 func (cmd *command) validateFlags() error {
-	if cmd.opts.Local {
+	if cmd.opts.Source == "local" {
```
clebs: I have some points regarding the flag parsing:
- I do not see the option latest? how do I install master now?
- What happens if i pass a version that does not follow semantic versioning (e.g. 123456)?
- What happens if I pass a docker image without a tag so there is no `:`? Basing selection on that is very unstable.
- The whole nested conditionals are not robust and not readable, I think we should have something more clean:
```go
switch o.Source {
    case "local":
        <do local setup>
    case "latest":
        <do master setup>
    case isSemVer(o.Source):
        <do release setup>
    case isDockerImage(o.Source):
        <do custom image setup>
    default:
        return errors.New("Invalid installation source")
}
```

cmd/kyma/install/cmd.go
```diff
@@ -78,16 +80,13 @@ func NewCmd(o *Options) *cobra.Command {
 		Aliases: []string{"i"},
 	}
 
-	cobraCmd.Flags().StringVarP(&o.ReleaseVersion, "release", "r", DefaultKymaVersion, "Kyma release or Git revision to be installed")
-	cobraCmd.Flags().StringVarP(&o.ReleaseConfig, "config", "c", "", "URL or path to the Installer configuration yaml file")
 	cobraCmd.Flags().BoolVarP(&o.NoWait, "noWait", "n", false, "Do not wait for the Kyma installation to complete")
 	cobraCmd.Flags().StringVarP(&o.Domain, "domain", "d", localDomain, "Domain used for installation")
 	cobraCmd.Flags().StringVarP(&o.TLSCert, "tlsCert", "", "", "TLS certificate for the domain used for installation")
 	cobraCmd.Flags().StringVarP(&o.TLSKey, "tlsKey", "", "", "TLS key for the domain used for installation")
+	cobraCmd.Flags().StringVarP(&o.Source, "source", "s", DefaultKymaVersion, "Installation source")
```
suleymanakbas91: You can list all the possible types that source can take. Something like ("local", "latest", release version (e.g. 1.4.1), or installer image)

cmd/kyma/install/cmd.go
```diff
@@ -102,15 +101,18 @@ func (cmd *command) Run() error {
...
 	}
+	s.Successf("Configuration validated")
 
-	s := cmd.NewStep("Checking installation source")
+	s = cmd.NewStep("Checking installation source")
 	if cmd.opts.Local {
 		s.LogInfof("Installing Kyma from local path: '%s'", cmd.opts.LocalSrcPath)
 	} else {
-		s.LogInfof("Installing Kyma in version '%s'", cmd.opts.ReleaseVersion)
+		s.LogInfof("Installing Kyma in version '%s'. Config version '%s'", cmd.opts.ReleaseVersion, cmd.opts.ConfigVersion)
```
suleymanakbas91: As a user, I wouldn't understand the difference between the installer image version and config version from this message. Either we need to make it more clear what we mean here, or just print the kyma version.

suleymanakbas91: Also, when I provide a docker image, this is the output `Installing Kyma in version ''`. It should print the docker image I provided.

cmd/kyma/install/cmd.go
```diff
@@ -219,21 +235,33 @@ func (cmd *command) validateFlags() error {
...
+		cmd.opts.RegistryTemplate = registryReleaseImagePattern
+		break
+
+	//Install the kyma with the specific installer image (docker image URL)
+	case cmd.isDockerImage(cmd.opts.Source):
+		cmd.opts.RemoteImage = cmd.opts.Source
+		cmd.opts.ConfigVersion = DefaultKymaVersion
+		break
+	default:
+		return fmt.Errorf("Source flag is not specified or it is not 'local' or a valid semver (ex: 1.3.0) or a docker image url")
```
suleymanakbas91:
```suggestion
		return fmt.Errorf("Failed to parse the source flag. It can take one of the following: 'local', 'latest', release version (e.g. 1.4.1), or installer image")
```

cmd/kyma/install/cmd.go
```diff
@@ -102,15 +100,21 @@ func (cmd *command) Run() error {
...
-	s := cmd.NewStep("Checking installation source")
+	s = cmd.NewStep("Checking installation source")
 	if cmd.opts.Local {
 		s.LogInfof("Installing Kyma from local path: '%s'", cmd.opts.LocalSrcPath)
 	} else {
-		s.LogInfof("Installing Kyma in version '%s'", cmd.opts.ReleaseVersion)
+		if cmd.opts.ReleaseVersion == cmd.opts.ConfigVersion {
+			s.LogInfof("Installing Kyma in version '%s'.", cmd.opts.ReleaseVersion)
+		}
+		s.LogInfof("Installing Kyma in version '%s'. Config version '%s'", cmd.opts.ReleaseVersion, cmd.opts.ConfigVersion)
```
suleymanakbas91:
```suggestion
		s.LogInfof("Installing Kyma version '%s'", cmd.opts.ReleaseVersion)
```

cmd/kyma/install/cmd.go
```diff
@@ -102,15 +100,21 @@ func (cmd *command) Run() error {
...
+	s.Successf("Configuration validated")
 
-	s := cmd.NewStep("Checking installation source")
+	s = cmd.NewStep("Checking installation source")
 	if cmd.opts.Local {
 		s.LogInfof("Installing Kyma from local path: '%s'", cmd.opts.LocalSrcPath)
 	} else {
-		s.LogInfof("Installing Kyma in version '%s'", cmd.opts.ReleaseVersion)
+		if cmd.opts.ReleaseVersion == cmd.opts.ConfigVersion {
+			s.LogInfof("Installing Kyma in version '%s'.", cmd.opts.ReleaseVersion)
```
suleymanakbas91:
```suggestion
			s.LogInfof("Using the installation configurations from '%s'.", cmd.opts.ConfigVersion)
```

cmd/kyma/install/cmd.go
```diff
@@ -102,15 +100,21 @@ func (cmd *command) Run() error {
...
 	}
+	s.Successf("Configuration validated")
 
-	s := cmd.NewStep("Checking installation source")
+	s = cmd.NewStep("Checking installation source")
 	if cmd.opts.Local {
 		s.LogInfof("Installing Kyma from local path: '%s'", cmd.opts.LocalSrcPath)
 	} else {
-		s.LogInfof("Installing Kyma in version '%s'", cmd.opts.ReleaseVersion)
+		if cmd.opts.ReleaseVersion == cmd.opts.ConfigVersion {
```
suleymanakbas91:
```suggestion
		if cmd.opts.ReleaseVersion != cmd.opts.ConfigVersion {
```

PR #222, opened 2019-09-03
Kyma CLI documentation generation

igor-karpukhin:
Added target `docs` to the Makefile to generate documentation for Kyma CLI commands and flags. After generation docs will be stored in the `docs/gen-docs` folder.


cmd/main.go
```diff
@@ -1,15 +1,30 @@
...
 )
 
+var AutoGenDoc string
+
+const DocsTargetDir = "./docs"
+
 func main() {
 	command := kyma.NewCmd(cli.NewOptions())
 
+	if AutoGenDoc == "true" {
```
clebs: I find it quite strange to have the docs generation on the main command, we have to compile the whole CLI just to generate a couple of MD files.

It would be better to have a docs command or ideally use `go generate` so we do not need to compile and run things.

Gopkg.toml
```diff
@@ -45,3 +45,7 @@ required = [
 [[constraint]]
   name = "github.com/stretchr/testify"
   version = "1.3.0"
+
+[[override]]
+  name = "github.com/russross/blackfriday"
```
clebs: What is this dependency used for? I do not see it imported anywhere.

igor-karpukhin: This is used implicitly by `github.com/spf13/cobra` -> `github.com/cpuguy83/go-md2man` -> `github.com/russross/blackfriday`, and it's not working with the latest version.

cmd/main.go
igor-karpukhin: I'm afraid this is the only way of doing that. https://github.com/spf13/cobra/blob/master/doc/md_docs.md

Gopkg.toml
clebs: I see, strange that they do not lock in the version that works on cobra...

cmd/main.go
clebs: hmm.. that is not what I understand from that document.

What I see is that you need to instantiate the command tree you want to document, but it is not required to do it in the exact same place right? We can have a doc generation specific main program which instantiates the Kyma root command and documents it.

Makefile
```diff
@@ -39,6 +39,12 @@ build-darwin: generate
 generate:
 	go generate ./...
 
+.PHONY: docs
+docs:
+	CGO_ENABLED=0 GOOGS=darwin go build -o ./bin/kyma-darwin-doc -ldflags '-X main.AutoGenDoc=true' ./cmd
```
clebs: We can just do:
```bash
go run -ldflags '-X main.AutoGenDoc=true' ./cmd
```

Benefits:
- It works in any platform (current command only works on OSX).
- No need to crete a binary
- No need to delete it afterwards.

cmd/main.go
igor-karpukhin: Moved generation to a separate binary `/cmd/gendocs/`

Makefile
igor-karpukhin: Fixed

cmd/main.go
clebs: looks great thanks!

Makefile
clebs: great thanks

PR #224, opened 2019-09-03
Added OWNERS files for code and docs

igor-karpukhin:
This PR relates to the Tide integration process when OWNERS file that will replace CODEOWNERS.


suleymanakbas91:
I think it'd be great if we could use this [`filter`](https://github.com/kubernetes/community/blob/master/contributors/guide/owners.md#filters) field to set tws as approvers for the changes including markdown files.

Issue #158, opened 2019-07-08
Remove kubectl binary usage


a-thaler:
**Description**
In https://github.com/kyma-project/cli/issues/85 the main parts of kubectl binary usage were replaced. Still left are all apply methods (for the installer.yaml, tiller.yaml and overrides). Furthermore, all management of the installer installation resource is still based on the binary.

**Reasons**
Consistent usage of kubectl library instead of binary, see also https://github.com/kyma-project/cli/issues/85

suleymanakbas91:
Implemented with https://github.com/kyma-project/cli/pull/470.

PR #233, opened 2019-09-09
Improve summary and patch configmaps with K8s API

clebs:
**Description**

- Improved the summary after installing Kyma
Screen Shot 2019-09-09 at 17 38 19
- Now configmap patching uses K8s API instead of Kubectl
- The kyma k8s API now exposes the client configuration
- Removed some unnecessary explicitly ignored return values

**Related issue(s)**
A bit of #158

### DIFF
--- a/internal/nice/nice.go
+++ b/internal/nice/nice.go
@@ -1,0 +1,25 @@
+// package nice provides nice printing for the CLI
+package nice
+
+import (
+	"fmt"
+
+	ct "github.com/daviddengcn/go-colortext"
+)
+
+// PrintKyma prints the Kyma word with its identity color
+func PrintKyma() {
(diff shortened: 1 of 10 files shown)


internal/nice/nice.go
```diff
@@ -0,0 +1,25 @@
...
+package nice
+
+import (
+	"fmt"
+
+	ct "github.com/daviddengcn/go-colortext"
+)
+
+// PrintKyma prints the Kyma word with its identity color
+func PrintKyma() {
```
igor-karpukhin: Will this colorful output work in Windows?

clebs: According to the docs at https://github.com/daviddengcn/go-colortext it already handles both unix and windows terminals accordingly.

clebs: Tested on power shell and it works :)

Issue #241, opened 2019-09-20
Add profile flag for minikube provision


iancrowther:
**Description**

Provisioning a new minikube cluster defaults to 'minikube'.  I want to have multiple clusters to allow me to experiment with multiple features in isolation.

**Reasons**

It's a default minikube config feature

iancrowther:
merged https://github.com/kyma-project/cli/pull/242

PR #242, opened 2019-09-20
adding profile flags to cli

iancrowther:
**Description**

- adding minikube profile flag to the cli

**Related issue(s)**
See also #241


cmd/kyma/provision/minikube/cmd.go
```diff
@@ -230,6 +231,7 @@ func (c *command) startMinikube() error {
 		"--vm-driver", c.opts.VMDriver,
 		"--disk-size", c.opts.DiskSize,
 		"-b", bootstrapper,
+		"-profile", c.opts.Profile,
```
clebs: Usually, when a flag is not a shorthand, it uses double hyphen `--` for its full name.

clebs:
Hi @iancrowther,
thanks for your PR! LGTM except for a small detail which I commented on the code.

a-thaler:
Hi @iancrowther,
can you please fix the problem detected by @clebs , then it will be ready to merge.

After merge we should change the integration job to use the new flag

cmd/kyma/provision/minikube/cmd.go
iancrowther: no probs

Issue #255, opened 2019-10-09
Use upper case for default values when asking user for confirmation


nachtmaar:
**Description**

When asking the user to confirm/decline an operation, usually the default value is written in upper case letters. Let's apply this standard to the CLI.

https://stackoverflow.com/questions/7803728/standard-format-for-yes-no-questions-in-the-terminal

Example:
```shell
$ kyma provision minikube
...
? Do you want to remove the existing Minikube cluster? Type [y/n]:
```

https://stackoverflow.com/questions/7803728/standard-format-for-yes-no-questions-in-the-terminal

**Expected result**

Example:
```shell
$ kyma provision minikube
...
? Do you want to remove the existing Minikube cluster? Type [Y/n]:
```

**Actual result**

**Steps to reproduce**

**Troubleshooting**

jbleduigou:
Looking into it

clebs:
This issue was already foxed by #256. It stayed open by mistake, closing

PR #256, opened 2019-10-09
Use upper case

jbleduigou:
Use upper case for default values when asking user for confirmation 
https://github.com/kyma-project/cli/issues/255


internal/root/root.go
```diff
@@ -4,13 +4,13 @@ import "fmt"
...
 	for {
-		fmt.Print("Type [y/n]: ")
+		fmt.Print("Type [Y/n]: ")
 		var res string
 		if _, err := fmt.Scanf("%s", &res); err != nil {
 			return false
 		}
 		switch res {
-		case "yes", "y":
+		case "Yes", "Y":
```
clebs: The yes option should not only accept a capital Y, people are used to just press y + enter.
Please could you add also lower case `y` and `yes`? Thanks!

clebs:
Hi @jbleduigou,

thanks for your Pull Request!

We just migrated the Kyma CLI to `go modules`. Would you be so kind to rebase your branch with the latest changes from master? Otherwise the CI/CD will fail.

Thanks and sorry for the inconvenience!

internal/root/root.go
jbleduigou: Yes indeed, work was done

PR #264, opened 2019-10-16
Fix 'make install'

sjanota:
**Description**

- Move main.go from `./cmd` to `./cmd/kyma` so it can be `make install`ed as `kyma` binary.
- Adapt rest of the scripts to use this path

**Related issue(s)**


clebs:
Hi @sjanota,

thanks for the PR!

Unfortunately we can not merge this, the package structure we have on the CLI is deliberate and has been designed after a long analysis for the following benefits:
- Extensibility: both in kyma and command line functionality.
- Maintainability: much easier to navigate and keep minimise side effects.
- Consistency: with other similar CLI tools we studied on our analysis.

Issue #248, opened 2019-10-04
Restarting minikube throws an error


pbochynski:
**Description**
If you provision minikube without deleting the existing machine it fails with the error:
```
X Creating cluster info ConfigMap
Error: configmaps "kyma-cluster-info" already exists
```

**Expected result**
I wold like to see success message.

**Actual result**

```
X Creating cluster info ConfigMap
Error: configmaps "kyma-cluster-info" already exists
```

**Steps to reproduce**

```
kyma provision minikube
kyma install --source local --src-path ./
minikube stop
kyma provision minikube
```
Answer no for the question: "Do you want to remove the existing Minikube cluster?"

a-thaler:
Should not fail for sure. I even would say it should update the configmap with the potential new settings if exist

a-thaler:
In case of existence of either owndomain-overrides or kyma-cluster-info it now prints: "ConfigMap already exists". No further context gets provided

That is a CLI and every line printed to the console needs to be meaningful and useful for the user. Please re-think if you should inform the user at all (maybe in verbose level?) and if yes if all needed info is provided

PR #266, opened 2019-10-17
Don't fail if cm exists

suleymanakbas91:
**Description**

- Don't fail if cm exists

**Related issue(s)**

See https://github.com/kyma-project/cli/issues/248

### DIFF
--- a/cmd/kyma/install/cmd.go
+++ b/cmd/kyma/install/cmd.go
@@ -1067,7 +1067,15 @@ func (cmd *command) patchMinikubeIP(minikubeIP string) error {
 }
 
 func (cmd *command) createOwnDomainConfigMap() error {
-	_, err := cmd.K8s.Static().CoreV1().ConfigMaps("kyma-installer").Create(&corev1.ConfigMap{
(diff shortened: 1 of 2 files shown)


cmd/kyma/install/cmd.go
```diff
@@ -1067,7 +1067,15 @@ func (cmd *command) patchMinikubeIP(minikubeIP string) error {
 }
 
 func (cmd *command) createOwnDomainConfigMap() error {
-	_, err := cmd.K8s.Static().CoreV1().ConfigMaps("kyma-installer").Create(&corev1.ConfigMap{
+	cm, err := cmd.K8s.Static().CoreV1().ConfigMaps("kyma-installer").Get("owndomain-overrides", metav1.GetOptions{})
+	if err == nil && cm != nil {
+		fmt.Println("ConfigMap already exists")
```
clebs: Would it be necessary to update the config map in this case? Or can we just reuse the existing one?

suleymanakbas91: This ConfigMap reflects the information about the cluster it is deployed on. Since the cluster has not changed, it does not need to be updated.

suleymanakbas91: In case the user wants to install kyma with another domain, it makes sense to update this configmap.

Issue #283, opened 2019-10-24
Installing Kyma from source local does not work


aszecowka:
**Steps to reproduce**
1. Fetch Kyma CLI from master branch
2. Execute `make build-darwin`
3. Move binary to your path: `./bin/kyma-darwin /usr/local/bin/kyma`
4. Provision minikube: `kyma provision minikube`
5.  Install Kyma from local sources: `kyma install --source local`

Installation hangs on:
` Deploying Kyma Installer`

Status of `kyma-installer` namespace is the following:
```
NAME                              READY   STATUS              RESTARTS   AGE
helm-certs-job-5ds26              0/1     CrashLoopBackOff    5          5m43s
kyma-installer-6c479467f6-64xjr   0/1     ContainerCreating   0          5m43s
```
Helm certs job fails because:
```"tiller-secret" not found```

When I checked how Tiller is installed, I noticed that instead of using my local `tiller.yaml` this file is used:
`https://raw.githubusercontent.com/kyma-project/kyma/latest//installation/resources/tiller.yaml`

As you can see, `latest` version is used, which is really old. In this version, tiller is installed with disabled TLS and probably because of that `tiller-secret` does not exist.

a-thaler:
Solved by the linked PR

Issue #245, opened 2019-09-26
Decouple install functionality from command


clebs:
**Description**
Right now the install command contains all the logic to install Kyma. It would be very beneficial to extract and encapsulate that logic into a standalone package for reusability.

**Reasons**
- Make the CLI more modular
- Separate command and terminal logic from Kyma specific logic
- Allow others to import the package and use it too.

a-thaler:
Please assure that the new API was reviewed by the compass team before merge

PR #286, opened 2019-10-29
Decouple installation logic from command

suleymanakbas91:
**Description**

- Decouple installation logic from command

**Related issue(s)**

See https://github.com/kyma-project/cli/issues/245
Solves #283


pkg/installation/utils.go
```diff
@@ -0,0 +1,232 @@
...
+				return deployment.Spec.Template.Spec.Containers[0].Image, nil
+			}
+		}
+	}
+	return "", errors.New("'kyma-installer' deployment is missing")
+}
+
+func replaceInstallerImage(resources *[]map[string]interface{}, imageURL string) error {
+	for _, config := range *resources {
+		kind, ok := config["kind"]
```
clebs: Maybe we can reduce the length and improve readability of this function by using positive if statements instead of negative ones + continue.
Something like:
```go
for _, config := range *resources {
    if kind, ok := config["kind"]; ok && kind == "deployment" {
        if spec, ok := config["spec"].(map[interface{}]interface{}); ok {
            if template, ok := spec["template"].(map[interface{}]interface{}); ok {
                ....
            }
        }
    }
}
```

The downside of this is a lot of nested if statements, not sure which is better. 🤔

pkg/installation/utils.go
```diff
@@ -0,0 +1,232 @@
...
+	_, err := i.k8s.Static().CoreV1().ConfigMaps("kyma-installer").Patch("installation-config-overrides", types.JSONPatchType,
+		[]byte(fmt.Sprintf("[{\"op\": \"replace\", \"path\": \"/data/global.adminPassword\", \"value\": \"%s\"}]", encPass)))
+	if err != nil {
+		err = errors.Wrap(err, "Error setting admin password")
+	}
+	return err
+}
+
+func removeActionLabel(acc *[]map[string]interface{}) error {
+	for _, config := range *acc {
```
clebs: This function can also be smaller by not using continue statements.

suleymanakbas91: Done

pkg/installation/installation.go
```diff
@@ -0,0 +1,685 @@
...
+		i.Options.remoteImage = i.Options.Source
+		i.Options.configVersion = "master"
+	default:
+		return fmt.Errorf("Failed to parse the source flag. It can take one of the following: 'local', 'latest', release version (e.g. 1.4.1), or installer image")
+	}
+
+	// If one of the --domain, --tlsKey, or --tlsCert is specified, the others must be specified as well (XOR logic used below)
+	if (i.Options.Domain != localDomain || i.Options.TLSKey != "" || i.Options.TLSCert != "") &&
+		!(i.Options.Domain != localDomain && i.Options.TLSKey != "" && i.Options.TLSCert != "") {
+		return errors.New("You specified one of the --domain, --tlsKey, or --tlsCert without specifying the others. They must be specified together")
```
akgalwas: The error message is specific to CLI.

pkg/installation/installation.go
```diff
@@ -0,0 +1,685 @@
...
+	releaseSrcURLPattern   = "https://raw.githubusercontent.com/kyma-project/kyma/%s/%s"
+	releaseResourcePattern = "https://raw.githubusercontent.com/kyma-project/kyma/%s/installation/resources/%s"
+	registryImagePattern   = "eu.gcr.io/kyma-project/kyma-installer:%s"
+	localDomain            = "kyma.local"
+)
+
+// Installation contains the installation elements and configuration options.
+type Installation struct {
+	k8s         kube.KymaKube
+	kubectl     *kubectl.Wrapper
```
akgalwas: Introducing some interface instead of kubectl.Wrapper will make the design more flexible. The client could provide specific implementation.

pkg/installation/installation_test.go
```diff
@@ -0,0 +1,140 @@
+package installation
```
akgalwas: There is a lot of logic in Installation component that is not unit tested.

pkg/installation/installation.go
```diff
@@ -0,0 +1,685 @@
...
+					currentDesc = desc
+				}
+
+			case "":
+				i.currentStep.LogInfo("Failed to get the installation status. Will retry later...")
+
+			default:
+				i.currentStep.Failure()
+				fmt.Printf("Unexpected status: %s\n", status)
+				os.Exit(1)
```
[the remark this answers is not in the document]
suleymanakbas91: Removed

pkg/installation/installation.go
```diff
@@ -0,0 +1,685 @@
...
+			}
+			time.Sleep(10 * time.Second)
+		}
+	}
+}
+
+func (i *Installation) printSummary() error {
```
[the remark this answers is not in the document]
suleymanakbas91: Done

pkg/installation/installation.go
```diff
@@ -0,0 +1,685 @@
...
+
+		_, err = i.getKubectl().RunCmd("apply", "-f", path)
+		if err != nil {
+			return err
+		}
+	}
+	return i.k8s.WaitPodStatusByLabel("kube-system", "name", "tiller", corev1.PodRunning)
+}
+
+func (i *Installation) loadAndConfigureInstallationFiles() ([]map[string]interface{}, error) {
```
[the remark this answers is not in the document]
suleymanakbas91: I renamed the function as prepareInstallationFiles.

PR #296, opened 2019-11-04
Add watch option for `kyma test run` command

mszostok:
**Description**

- Add watch option for `kyma test run` command with the `--timeout` support
- Add watch method for octopus client
- Add test coverage for new functionality
   Test was executed with -race and -count 50 flags to check it stability.
   ```bash
    go test ./cmd/kyma/test/run/... -run=Test_WaitForTestSuite -count 50 
    go test ./cmd/kyma/test/run/... -run=Test_WaitForTestSuite -race
    ```
- Add missing error handling for octopus client

Showcase success:
[asciicast](https://asciinema.org/a/fWVCp8cYfgya9SInajpKwxn6D)

Showcase failure:
[asciicast](https://asciinema.org/a/JiF6UH2ShHpSKjzTFv4QfwO8w)

**Reproduce locally**
```bash
# Build binary 
make build-darwin

# Create test definitions always-pass
kubectl create -f https://gist.github.com/mszostok/b0bdaa929edc036321ecfc4278b7aa22/raw
kubectl create -f https://gist.github.com/mszostok/cabb47e8356a33b1c5215fd43eea1890/raw

# Create test definition always-fail
kubectl create -f https://gist.github.com/mszostok/0ff2353866bac900ff2ac28943d7b073/raw


# Run always-pass
/bin/kyma-darwin test run test-always-pass test-always-pass-v2 --watch --timeout "1m"

# Run always-fail
/bin/kyma-darwin test run test-always-pass  test-always-fail --watch --timeout "1m"

```
**Related issue(s)**

- https://github.com/kyma-project/test-infra/issues/1636

/cc @a-thaler @suleymanakbas91


mszostok:
Could you tell me what I should do in case of docs? 

when I'm executing `make docs` then mostly the whole documentation is regenerated.

btw. the `pre-kyma-cli-integration` is quite flaky it failed 2 times in row: https://status.build.kyma-project.io/pr-history/?org=kyma-project&repo=cli&pr=296

cmd/kyma/test/run/cmd.go
```diff
@@ -42,17 +50,25 @@ To execute all test defintions, run ` + "`kyma test run -n example-test`" + `.
...
-	cobraCmd.Flags().Int64VarP(&o.MaxRetries, "max-retries", "", 0, "Defines how many times a given test is retried when it fails. A suite is marked with a \"succeeded\" status even if some tests failed at first and then finally succeeded. The default value of 0 means that there are no retries of a given test.")
+	cobraCmd.Flags().Int64VarP(&o.ExecutionCount, "count", "c", 1, `Defines how many times every test should be executed. "count" and "max-retries" flags are mutually exclusive.`)
+	cobraCmd.Flags().Int64VarP(&o.MaxRetries, "max-retries", "", 0, `Defines how many times a given test is retried when it fails. A suite is marked with a "succeeded" status even if some tests failed at first and then finally succeeded. The default value of 0 means that there are no retries of a given test.`)
 	cobraCmd.Flags().Int64VarP(&o.Concurrency, "concurrency", "", 1, "Specifies the number of tests to be executed in parallel.")
+	cobraCmd.Flags().DurationVar(&o.Timeout, "timeout", 10*time.Minute, `Maximum time during which the test suite is being watched. Valid time units are "ns", "us" (or "µs"), "ms", "s", "m", "h".`)
```
a-thaler: A typical kyma test suite will take around 30minutes I guess..
usually the watch gets interrupted by finishing the test suite. The timeout is only for the case of hanging tests and not having a hanging CI. To detect hanging tests you need to wait much more than the regular suite execution duration.
So shouldn't be the default something around 1 hour?

Or maybe better to have no default at all, so by default it runs forever. For local execution it is probably what you want. For CI situation you will specify a timeout always.

cmd/kyma/test/run/cmd.go
```diff
@@ -42,17 +50,25 @@ To execute all test defintions, run ` + "`kyma test run -n example-test`" + `.
...
 	cobraCmd.Flags().Bool("help", false, "Displays help for the command.")
 	return cobraCmd
 }
 
 func (cmd *command) Run(args []string) error {
 	var err error
-	if cmd.K8s, err = kube.NewFromConfig("", cmd.KubeconfigPath); err != nil {
-		return errors.Wrap(err, "Could not initialize the Kubernetes client. Make sure your kubeconfig is valid.")
+	if cmd.opts.Watch {
+		if cmd.K8s, err = kube.NewFromConfigWithTimeout("", cmd.KubeconfigPath, cmd.opts.Timeout); err != nil {
```
a-thaler: The apiserver commands should have a timeout always I guess. And that timeout should not be the one used for the watch

mszostok: dicussed

cmd/kyma/test/run/cmd.go
mszostok: I've checked also the `kubectl` approach and basically it's messy, they have such policies for timeout:

- `"The length of time to wait before giving up on a delete, zero means determine a timeout from the size of the object"`
- "The length of time to wait before giving up, zero means infinite"
- "The length of time to wait before giving up.  Zero means check once and don't wait, negative means wait for a week."



but IMO adding opt that default it runs forever is the best option, so I changed and now the default value is zero which means infinite timeout.

cmd/kyma/test/run/cmd.go
```diff
@@ -45,13 +53,21 @@ To execute all test defintions, run ` + "`kyma test run -n example-test`" + `.
 	cobraCmd.Flags().Int64VarP(&o.ExecutionCount, "count", "c", 1, `Defines how many times every test should be executed. "count" and "max-retries" flags are mutually exclusive.`)
 	cobraCmd.Flags().Int64VarP(&o.MaxRetries, "max-retries", "", 0, `Defines how many times a given test is retried when it fails. A suite is marked with a "succeeded" status even if some tests failed at first and then finally succeeded. The default value of 0 means that there are no retries of a given test.`)
 	cobraCmd.Flags().Int64VarP(&o.Concurrency, "concurrency", "", 1, "Specifies the number of tests to be executed in parallel.")
+	cobraCmd.Flags().DurationVar(&o.Timeout, "timeout", 0, `Maximum time during which the test suite is being watched, zero means infinite. Valid time units are "ns", "us" (or "µs"), "ms", "s", "m", "h".`)
```
bszwarc:
```suggestion
	cobraCmd.Flags().DurationVar(&o.Timeout, "timeout", 0, `Maximum time during which the test suite is being watched, where "0" means "infinite". Valid time units are "ns", "us" (or "µs"), "ms", "s", "m", "h".`)
```

cmd/kyma/test/run/cmd.go
```diff
@@ -45,13 +53,21 @@ To execute all test defintions, run ` + "`kyma test run -n example-test`" + `.
...
+	cobraCmd.Flags().BoolVarP(&o.Watch, "watch", "w", o.Watch, "Watch the status of the test suite until it finishes or the defined `--timeout` occurs.")
```
bszwarc:
```suggestion
	cobraCmd.Flags().BoolVarP(&o.Watch, "watch", "w", o.Watch, "Watches the status of the test suite until the tests finish or the defined `--timeout` occurs.")
```

clebs: Shouldn't here the default value be `false`?

cmd/kyma/test/run/cmd.go
```diff
@@ -89,15 +105,19 @@ func (cmd *command) Run(args []string) error {
...
-	if err != nil {
+
+	if _, err := cmd.K8s.Octopus().CreateTestSuite(testResource); err != nil {
 		return err
 	}
+	fmt.Printf("- Test suite '%s' successfully created\r\n", testSuiteName)
 
-	if _, err := cmd.K8s.Octopus().CreateTestSuite(testResource); err != nil {
+	waitStep := cmd.NewStep("Waiting for test suite to finish")
+	err = waitForTestSuite(cmd.K8s.Octopus(), testResource.Name, clusterTestSuiteCompleted(waitStep), cmd.opts.Timeout)
```
clebs: We should only wait for the suite if `cmd.opts.Watch` == `true`

cmd/kyma/test/run/cmd.go
mszostok: It depends. In kubectl they always use the zero-value of the given option,  so in this case `o.Watch` but I've check your approach and you always set the "specified" defaults. To be consistent I adjusted that to be `false`

clebs: Yes, I know about that in kubectl. But we decided to put explicit values to make it simpler and more readable.

PR #297, opened 2019-11-05
Add functionality for fetching logs from test suite

mszostok:
**Description**

- Add functionality for fetching logs from test suite

   Now you can execute such command
   ```bash
    kyma test logs test-suite-name --test-status Failed
    ```
   to fetch the logs from `Failed` testing pods created by the `test-suite-name` test suite.

- Add test coverage for implemented functionality

**Showcase:**
[asciicast](https://asciinema.org/a/279238)

**Reproduce locally**
```bash
# Build binary 
make build-darwin

# Create test definition always-fail
kubectl create -f https://gist.github.com/mszostok/2d8653ecd65569be555078b5b3be80e0/raw

# Run always-pass
/bin/kyma-darwin test run always-fail --name failed-test-suite

# Run always-fail
/bin/kyma-darwin test logs failed-test-suite

```

> **NOTE:** This needs to be merged before the `--output=junit` opt will be added to the `kyma test status` command

**Related issue(s)**

- kyma-project/test-infra#1636

/cc @a-thaler @suleymanakbas91


cmd/kyma/test/logs/cmd.go
```diff
@@ -0,0 +1,143 @@
...
+		Long: `Use this command to display the logs of a test executing for given test suite.
+
+To print the status of specific test cases, ` + "`run kyma test logs testSuiteOne testSuiteTwo`" + `.
+Provide at least one test suite name.
+`,
+
+		RunE: func(_ *cobra.Command, args []string) error { return cmd.Run(args) },
+	}
+
+	cobraCmd.Flags().Bool("help", false, "Displays help for the command.")
```
a-thaler: No need to specify help here, it is defined on root level for all commands

cmd/kyma/test/logs/cmd.go
```diff
@@ -0,0 +1,143 @@
...
+	cobraCmd.Flags().StringVar(&o.InStatus, "in-status", defaultLogsInStatus, "Display logs only from testing pods in given status.")
```
a-thaler: We need to find a better param name...
maybe "pod-status" or "test-status"?

cmd/kyma/test/logs/cmd.go
```diff
@@ -0,0 +1,143 @@
...
+func (cmd *command) Run(args []string) error {
+	if err := cmd.validateFlags(); err != nil {
+		return err
+	}
+
+	if len(args) < 1 {
+		return fmt.Errorf("Test suite name required")
+	}
+
+	k8sClient, err := kube.NewFromConfig("", cmd.KubeconfigPath)
```
clebs:
```suggestion
        var err error
	cmd.K8s, err = kube.NewFromConfig("", cmd.KubeconfigPath)
```

mszostok: I will apply your suggestion but two cents from my side:

I'm not convinced of that approach because

> When returning an error from a function with multiple return values, idiomatic Go code also will set each non-error value to a zero value

and basically it also should be followed here. IMO assignment to cmd.K8s should be done only if there is no error. In case of error, you may have sth in cmd.K8s. I know that in that given code is easy to say that it will not cause other problems but basically when I'm writing code then I'm doing that always in the same way

cmd/kyma/test/logs/cmd.go
mszostok: I vote for "test-status" 👍

cmd/kyma/test/logs/cmd.go
clebs: Well, based on that quote this applies about perfectly. Since `cmd.K8s` is an interface, its zero value is nil. So if `kube.NewFromConfig()` follows that, it will return nil for the first return value if there is an error.
So with that principle we can safely assign it.

cmd/kyma/test/common.go
```diff
@@ -42,3 +44,20 @@ func GetNumberOfFinishedTests(testSuite *oct.ClusterTestSuite) int {
...
+	suites, err := cli.ListTestSuites(metav1.ListOptions{})
+	if err != nil {
+		return nil, errors.Wrap(err, "Unable to list test suites")
+	}
+
+	result := []oct.ClusterTestSuite{}
+	for _, suite := range suites.Items {
+		for _, tName := range names {
+			if suite.ObjectMeta.Name == tName {
+				result = append(result, suite)
```
clebs: Once we found a match, we do not need to continue iterating the inner loop, we can jump to the next item.

mszostok: I didn't wrote that code I only copied it from here: https://github.com/kyma-project/cli/blame/cb9b22473c8bb49328a2fd2e09ff8521b24d58f0/cmd/kyma/test/status/cmd.go#L196-L211

for DRY.

but I've also added the indexing the names  as we want to optimize it

cmd/kyma/test/logs/cmd.go
```diff
@@ -0,0 +1,142 @@
...
+
+func NewCmd(o *options) *cobra.Command {
+	cmd := command{
+		Command: cli.Command{Options: o.Options},
+		opts:    o,
+	}
+
+	cobraCmd := &cobra.Command{
+		Use:   "logs <test-suite-1> <test-suite-2> ... <test-suite-N>",
+		Short: "Shows the logs of a related test executions pods.",
```
bszwarc:
```suggestion
		Short: "Shows the logs of tests Pods for a given test suite.",
```

cmd/kyma/test/common.go
clebs: I see... that one is definitely wrong as well.

But then new approach is quite good.

PR #303, opened 2019-11-18
Enable golangci-lint and fix detected errors

mszostok:
**Description**

Currently, there is not checks and linters executed against CLI codebase. There is https://github.com/kyma-project/cli/blob/31f0992a4b2f5ce0ecc5ff284c101aa27c342b10/Makefile#L18-L19

but instead of `golint ./...` you execute `golint` which do not detects problem in subdirectories 😔

This PR adds and enables the [golangci-lint](https://github.com/golangci/golangci-lint) and fixes all problems detected by these linters, **for example:**
- Errors were not returned at all: https://github.com/kyma-project/cli/blob/31f0992a4b2f5ce0ecc5ff284c101aa27c342b10/internal/files/files.go#L37-L39
https://github.com/kyma-project/cli/blob/31f0992a4b2f5ce0ecc5ff284c101aa27c342b10/internal/files/files.go#L57-L59
  
- Missspeling:  
  `defualt` is a misspelling of `default`
   `defintion` is a misspelling of `definition`
   `deleteing` is a misspelling of `deleting` 
   and so on and so forth...

  _doc are generated from source code, so IMO misspeling checker is really useful 
   /cc @bszwarc :)_
- Gofmt is not executed 🧐
   ```File is not `gofmt`-ed with `-s` (really problematic)```
   it's really problematic because Go contributors always have the `gofmt` enabled and when you create a new pr then when you execute `gofmt` (for me is by default) you also change lines totally not connected with your pr

  **Why I added `hack` dir?**
  Because it  is a common pattern in K8s eco-system, e.g.:
    -  https://github.com/jetstack/cert-manager/tree/master/hack
    - https://github.com/kubernetes-sigs/kind/tree/master/hack
    - https://github.com/kubernetes/kubernetes/tree/master/hack
    - and so on and so forth
  I've also added README.md file which describes the purpose of this folder:
    ```
    # Hack directory
    
    This package contains various scripts that are used by Kyma CLI developers.
    
    ## Purpose
    
    This directory contains tools, such as linters, that help to maintain the source code compliant to Go best coding practices. It also includes utility scripts that generate code, and scripts executed on CI pipelines.
    ```
  IMO it's really useful and you can story there other useful scripts, check this out:
    Screen Shot 2019-10-15 at 17 30 21

### How to exceute 
```bash

# Installs golangci-lint and execute it
`make validate`

# Run linters without installing golangci-lint
`env SKIP_INSTALL='true' make validate`
```

### List of detected errors

```
./hack/verify-lint.sh

#################################################################################################
# Sat Nov 16 21:57:44 PST 2019
# Run golangci-lint checks
#################################################################################################

internal/kubectl/kubectl.go:20: G204: Subprocess launched with function call as argument or cmd arguments (gosec)
	cmd := exec.Command("kubectl", fmt.Sprintf("--kubeconfig=%s", kubeconfig), "apply", "-f", "-")
cmd/kyma/test/logs/opts.go:13:33: exported func NewOptions returns unexported type *github.com/kyma-project/cli/cmd/kyma/test/logs.options, which can be annoying to use (golint)
func NewOptions(o *cli.Options) *options {
                                ^
cmd/kyma/uninstall/cmd.go:270:35: Error return value of `cmd.printUninstallationErrorLog` is not checked (errcheck)
			cmd.printUninstallationErrorLog()
			                               ^
cmd/kyma/uninstall/cmd_test.go:20:14: Error return value of `c.ParseFlags` is not checked (errcheck)
	c.ParseFlags([]string{"--timeout=60m0s"})
	            ^
internal/net/net.go:9: G102: Binds to all network interfaces (gosec)
	l, err := net.Listen("tcp", ":0")
pkg/installation/installation.go:189:22: error strings should not be capitalized or end with punctuation or a newline (golint)
			return fmt.Errorf("Configured 'src-path=%s' does not seem to point to a Kyma repository. Check if your repository contains the 'installation/resources' folder.", i.Options.LocalSrcPath)
			                  ^
pkg/installation/installation.go:521:22: error strings should not be capitalized or end with punctuation or a newline (golint)
			return errors.New("Could not generate the Helm certificate.")
			                  ^
pkg/installation/installation.go:629:23: error strings should not be capitalized or end with punctuation or a newline (golint)
				return fmt.Errorf("Unexpected status: %s\n", status)
				                  ^
pkg/installation/utils.go:159:47: `neccessary` is a misspelling of `necessary` (misspell)
	// Check if installer deployment has all the neccessary fields and a container named kyma-installer-container.
	                                             ^
cmd/kyma/console/cmd.go:29: File is not `gofmt`-ed with `-s` (gofmt)
		Use:     "console",
		Short:   "Opens the Kyma Console in a web browser.",
cmd/kyma/test/list/cmd.go:66:56: Using a reference for the variable on range scope `t` (scopelint)
			fmt.Sprintf("%d/%d", test.GetNumberOfFinishedTests(&t), len(t.Status.Results)),
			                                                    ^
cmd/kyma/test/list/opts.go:9:33: exported func NewOptions returns unexported type *github.com/kyma-project/cli/cmd/kyma/test/list.options, which can be annoying to use (golint)
func NewOptions(o *cli.Options) *options {
                                ^
cmd/kyma/test/status/cmd.go:77:40: Using a reference for the variable on range scope `t` (scopelint)
			if err := cmd.printTestSuiteStatus(&t, cmd.opts.OutputFormat); err != nil {
			                                    ^
cmd/kyma/test/status/cmd.go:88:40: Using a reference for the variable on range scope `t` (scopelint)
			if err := cmd.printTestSuiteStatus(&t, cmd.opts.OutputFormat); err != nil {
			                                    ^
cmd/kyma/test/status/opts.go:14:33: exported func NewOptions returns unexported type *github.com/kyma-project/cli/cmd/kyma/test/status.options, which can be annoying to use (golint)
func NewOptions(o *cli.Options) *options {
                                ^
internal/files/files.go:16:7: don't use underscores in Go names; const kyma_folder should be kymaFolder (golint)
const kyma_folder = ".kyma"
      ^
internal/files/files.go:38:14: Error return value of `errors.Wrap` is not checked (errcheck)
		errors.Wrap(err, "Could not save file")
		           ^
internal/files/files.go:58:14: Error return value of `errors.Wrap` is not checked (errcheck)
		errors.Wrap(err, "Could not load file")
		           ^
cmd/kyma/test/definitions/cmd.go:43:9: if block ends with a return statement, so drop this else and outdent its block (move short variable declaration to its own line if necessary) (golint)
	} else {
	       ^
cmd/kyma/test/definitions/opts.go:9:33: exported func NewOptions returns unexported type *github.com/kyma-project/cli/cmd/kyma/test/definitions.options, which can be annoying to use (golint)
func NewOptions(o *cli.Options) *options {
                                ^
cmd/kyma/kyma.go:42:28: `defualt` is a misspelling of `default` (misspell)
	// Kubeconfig env var and defualt paths are resolved by the kyma k8s client using the k8s defined resolution strategy.
	                          ^
cmd/kyma/kyma.go:45: File is not `gofmt`-ed with `-s` (gofmt)
	
cmd/kyma/kyma_test.go:23:14: Error return value of `c.ParseFlags` is not checked (errcheck)
	c.ParseFlags([]string{"--kubeconfig=/some/file", "--non-interactive=true", "--verbose=true"})
	            ^
cmd/kyma/install/cmd_test.go:76:23: Using the variable on range scope `test` in function literal (scopelint)
			cmd.opts.NoWait = !test.wait
			                   ^
cmd/kyma/install/cmd_test.go:77:33: Using the variable on range scope `test` in function literal (scopelint)
			err := cmd.importCertificate(test.cert)
			                             ^
cmd/kyma/install/cmd_test.go:79:21: Using the variable on range scope `test` in function literal (scopelint)
			require.Equal(t, test.expectedErr, err, "Error not as expected")
			                 ^
internal/helm/helm.go:8:2: a blank import should be only in a main or test package, or have a comment justifying it (golint)
	_ "k8s.io/client-go/plugin/pkg/client/auth/gcp"
	^
internal/logs/fetcher_test.go:106:67: exported func NewFakePodsGetter returns unexported type *github.com/kyma-project/cli/internal/logs_test.fakePodGetter, which can be annoying to use (golint)
func NewFakePodsGetter(t *testing.T, pod v1.Pod, podlogs string) (*fakePodGetter, func()) {
                                                                  ^
internal/kube/client_test.go:23:39: Error return value of `(k8s.io/client-go/kubernetes/typed/core/v1.PodInterface).Create` is not checked (errcheck)
	c.Static().CoreV1().Pods("ns").Create(&corev1.Pod{
	                                     ^
internal/kube/client_test.go:53:39: Error return value of `(k8s.io/client-go/kubernetes/typed/core/v1.PodInterface).Create` is not checked (errcheck)
	c.Static().CoreV1().Pods("ns").Create(&corev1.Pod{
	                                     ^
internal/kube/client_test.go:93:39: Error return value of `(k8s.io/client-go/kubernetes/typed/core/v1.PodInterface).Create` is not checked (errcheck)
	c.Static().CoreV1().Pods("ns").Create(pod)
	                                     ^
internal/kube/client_test.go:105:45: Error return value of `(k8s.io/client-go/kubernetes/typed/core/v1.PodInterface).UpdateStatus` is not checked (errcheck)
	c.Static().CoreV1().Pods("ns").UpdateStatus(pod)
	                                           ^
internal/kube/client_test.go:135:45: Error return value of `(k8s.io/client-go/kubernetes/typed/core/v1.PodInterface).UpdateStatus` is not checked (errcheck)
	c.Static().CoreV1().Pods("ns").UpdateStatus(pod)
	                                           ^
internal/kube/client_test.go:146:41: Error return value of `(k8s.io/client-go/kubernetes/typed/core/v1.NamespaceInterface).Create` is not checked (errcheck)
	c.Static().CoreV1().Namespaces().Create(&corev1.Namespace{
	                                       ^
cmd/kyma/provision/gcp/cmd.go:34: File is not `gofmt`-ed with `-s` (gofmt)
		Long:  `Use this command to provision a Kubernetes cluster on GCP for Kyma installation. Use the flags to specify cluster details.
NOTE: To access the provisioned cluster, make sure you get authenticated by Google Cloud SDK. To do so,run `+ "`gcloud auth application-default login`" + ` and log in with your Google Cloud credentials.`,
cmd/kyma/test/delete/opts.go:11:33: exported func NewOptions returns unexported type *github.com/kyma-project/cli/cmd/kyma/test/delete.options, which can be annoying to use (golint)
func NewOptions(o *cli.Options) *options {
                                ^
cmd/kyma/test/run/opts.go:19:33: exported func NewOptions returns unexported type *github.com/kyma-project/cli/cmd/kyma/test/run.options, which can be annoying to use (golint)
func NewOptions(o *cli.Options) *options {
                                ^
cmd/kyma/test/run/cmd.go:39:41: `defintion` is a misspelling of `definition` (misspell)
		Use:   "run <test-definition-1> <test-defintion-2> ... <test-definition-N>",
		                                      ^
cmd/kyma/test/run/cmd.go:139:33: `defintion` is a misspelling of `definition` (misspell)
			return nil, fmt.Errorf("Test defintion '%s' not found in the list of cluster test definitions", tName)
			                             ^
cmd/kyma/test/common.go:42:4: should replace `result += 1` with `result++` (golint)
			result += 1
			^
pkg/api/octopus/client.go:15:6: type name will be used as octopus.OctopusInterface by other packages, and that stutters; consider calling this Interface (golint)
type OctopusInterface interface {
     ^
pkg/api/octopus/client.go:24:6: type name will be used as octopus.OctopusRestClient by other packages, and that stutters; consider calling this RestClient (golint)
type OctopusRestClient struct {
     ^
pkg/api/octopus/client.go:79:25: `deleteing` is a misspelling of `deleting` (misspell)
		// Reenable this when deleteing supports options
		                      ^
cmd/kyma/provision/minikube/cmd.go:27:2: `vmDriverVirtualBox` is unused (varcheck)
	vmDriverVirtualBox string = "virtualbox"
	^
```


internal/logs/fetcher_test.go
```diff
@@ -91,78 +91,78 @@ func fixPodWithContainers(containersName ...string) v1.Pod {
 	return pod
 }
 
-// fakePodGetter implements SIMPLY fake for getting pods and fetching associated logs
+// FakePodGetter implements SIMPLY fake for getting pods and fetching associated logs
 // Unfortunately needs to be implemented by us, because client-go fake client
 // in current implementation (client-go v1.12) returns zero-value `rest.Request`
 // and executing GetLogs(/*..*/).DoRaw() method, throws panic.
-type fakePodGetter struct {
+type FakePodGetter struct {
```
clebs: Why is this type now being exported?
From what I can see, it is only used in this package.

cmd/kyma/test/definitions/opts.go
```diff
@@ -2,10 +2,10 @@ package definitions
 
 import "github.com/kyma-project/cli/internal/cli"
 
-type options struct {
+type Options struct {
```
clebs: Is there a reason to export all the `options` types in all test subcommands?

cmd/kyma/provision/minikube/cmd.go
```diff
@@ -19,11 +19,13 @@ import (
...
-	vmDriverHyperkit   string = "hyperkit"
-	vmDriverHyperv     string = "hyperv"
-	vmDriverNone       string = "none"
+	kubernetesVersion string = "1.14.6"
+	bootstrapper      string = "kubeadm"
+	vmDriverHyperkit  string = "hyperkit"
+	vmDriverHyperv    string = "hyperv"
+	vmDriverNone      string = "none"
+	// it's used in cfg_windows.go
+	// nolint: deadcode,varcheck
```
clebs: Instead of adding this, a better way to solve this warning would be to use this constant in line 35 of this same file (were currently `virtualbox` is a string literal).

mszostok: agree 👍

cmd/kyma/test/definitions/opts.go
mszostok: yes

_Reason:_
The `NewOptions()` functions is used outside the pkg and if the `options` is not exported then callers of that function from other packages will struggle to use it. For instance, they can't put a *options field in their own struct. So this is the reason why the golint returns such error.

internal/logs/fetcher_test.go
mszostok: I've exported that because the constructor was public. 
But as you noticed it's used only in this pkg, so I've just changed the constructor `NewFakePodsGetter` to be  private.

cmd/kyma/test/definitions/opts.go
clebs: What error dos the golint return?

Unexporting the options is intended like that so that no other commands can manipulate the specific options for the command they belong to.

Is there a scenario where a parent command needs to use the internal fields of this struct?

mszostok: > What error dos the golint return?

` exported func NewOptions returns unexported type *github.com/kyma-project/cli/cmd/kyma/test/status.options, which can be annoying to use (golint)
func NewOptions(o *cli.Options) *options {`

---
is not about using the internal fields of that struct is about using that struct at all. 

> Unexporting the options is intended like that so that no other commands can manipulate the specific options for the command they belong to.

it's not exactly true, when you unexport the `Options` struct then you are not allowed to assign that outside of that package even if you are exposing the public constructor. You still can expose the Options and have a private fields which are not accessible outside the package


pseudo-code:
```go
package main 

func main() {
# you can do sth like that when returned option is private
opts := definitions.NewOptions()

opts.Foo()
}
```


```go
package main

// you cannot have a field because options are private
type Bar struct {
	opts *options
}

func main() {

	obj := Bar{
		opts: definitions.NewOptions(),
	}

}
```

clebs: Ah I see... so the problem is not being able to embed this type in others. Do we have a scenario where we need to do this?

Otherwise it is reasonable that we have these options only available on the command they are relevant for. Others should only be able to create an instance to pass it to the command when being created, but that's it.

mszostok: > Do we have a scenario where we need to do this?
Currently, is not a case because of that you code compiles...

Let me put  this in this way:
Current implementation **violate the go lint rule**

What you want to do?
a) silence this error 
b) by idiomatic and fix that problem

Why you do not want to export that? If you have a public constructor then you return also the public struct, simple as that. Why you do not want to do that? It's just the best practices. You know already the pros and cons what else we want to discuss?

clebs: I am trying to come up with a good solution which also makes the linter happy.

I just try to find other ways because there is a general tendency when writing go code to export everything blindly.

One way I can come up with is just changing the subcommand constructor to take `cli.Options` as param and the unexported options creation and initialisation is completely happening inside the constructor.

antoineco: How about `// nolint`?

hack/README.md
```diff
@@ -0,0 +1,7 @@
+# Hack directory
+
+This package contains various scripts that are used by Kyma CLI developers.
+
+## Purpose
+
+This directory contains tools, such as linters, that help to maintain the source code compliant to Go best coding practices. It also includes utility scripts that generate code, and scripts executed on CI pipelines.
```
bszwarc:
```suggestion
This directory contains tools, such as linters, that help to maintain the source code compliant with Go best coding practices. It also includes utility scripts that generate code and scripts executed on CI pipelines.
```

PR #360, opened 2020-01-24
Increased default concurrency of test execution

a-thaler:
**Description**

- Increased default concurrency for test execution from 1 to 5 as it is not harming and execution is soo much faster
- ...
- ...

**Related issue(s)**


a-thaler:
done

Issue #353, opened 2020-01-23
Installing kyma using kyma-cli fails when helm 3.X is present.


rakesh-garimella:
**Description**

If a user is installing kyma using kyma-cli and has helm 3.x present then installation fails. Kyma-cli should try to check if helm 3.X is present on the system and should not start the installation giving the info message.

a-thaler:
The CLI will configure your Helm installation at the "kyma install" command with the required certificates retrieved from the installed kyma cluster. That's all, there is no real dependency in order to install kyma using the CLI. If helm is not installed it just passes the point and continues. Same should happen with a Helm 3 installation, just ignore it and continue.

PR #362, opened 2020-01-27
Helm v3 not supported

suleymanakbas91:
**Description**

- Continue the installation in case of Helm problems

**Related issue(s)**

See https://github.com/kyma-project/cli/issues/353

### DIFF
--- a/internal/helm/helm.go
+++ b/internal/helm/helm.go
@@ -16,7 +16,7 @@ var helmHomeCmd = exec.Command("helm", "home")
 func SupportedVersion() (bool, error) {
 	helmVersionRaw, err := helmVersionCmd.CombinedOutput()
 	if err != nil {
-		return false, nil
+		return false, err
 	}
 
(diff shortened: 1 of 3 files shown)


internal/helm/helm.go
```diff
@@ -9,11 +9,26 @@ import (
...
 
-var helmCmd = exec.Command("helm", "home")
+var helmVersionCmd = exec.Command("helm", "version", "--short", "--client")
+var helmHomeCmd = exec.Command("helm", "home")
+
+// SupportedVersion returns if the Helm version is supported by Kyma (currently v2.x.x)
+func SupportedVersion() (bool, error) {
+	helmVersionRaw, err := helmVersionCmd.CombinedOutput()
+	if err != nil {
+		return false, err
```
clebs: Here if the error is that helm is not installed we should not return an error.
If helm is not in the local system at all, Kyma can install normally.

Issue #370, opened 2020-02-13
`kyma provision gardener` command ignores SIGINT


nachtmaar:
**Description**
Kyma cli blocks when provisioning gardener cluster and ignores SIGINT signal.

**Expected result**
When pressing CTRL-C (sends SIGINT), kyma should unblock and the process should exit gracefully.

**Actual result**
Kyma blocks infinetely.

**Steps to reproduce**
```shell
$ kyma provision gardener --target-provider azure --secret tunas-secret --name grdnr-7u9y4m --project kymatunas --credentials /etc/credentials/gardener-kyma-prow/kubeconfig --region westeurope -t Standard_D2_v3 --disk-size 35 --disk-type=Standard_LRS --extra vnetcidr=10.250.0.0/19 --nodes 4
```
- Press CTRL-C or send signal manually to process.
- Here is what happens when I press CTRL-C. No matter how often I press it, kyma never stops to block.

```shell
^C


^C^C^C
^C
^C
^C2020-02-13T13:07:15.176Z [INFO]  plugin: configuring client automatic mTLS
2020-02-13T13:07:15.215Z [DEBUG] plugin: starting plugin: path=/root/.terraform.d/plugins/linux_amd64/terraform-provider-gardener_v0.0.7 args=[/root/.terraform.d/plugins/linux_amd64/terraform-provider-gardener_v0.0.7]
2020-02-13T13:07:15.215Z [DEBUG] plugin: plugin started: path=/root/.terraform.d/plugins/linux_amd64/terraform-provider-gardener_v0.0.7 pid=148
2020-02-13T13:07:15.215Z [DEBUG] plugin: waiting for RPC address: path=/root/.terraform.d/plugins/linux_amd64/terraform-provider-gardener_v0.0.7
2020-02-13T13:07:15.232Z [INFO]  plugin.terraform-provider-gardener_v0.0.7: configuring server automatic mTLS: timestamp=2020-02-13T13:07:15.231Z
^C2020-02-13T13:07:15.263Z [DEBUG] plugin: plugin process exited: path=/root/.terraform.d/plugins/linux_amd64/terraform-provider-gardener_v0.0.7 pid=148 error="signal: interrupt"

^C

^C
^C
^C
```

**Troubleshooting**

PR #382, opened 2020-02-20
Add graceful shutdown to kyma cli

rakesh-garimella:
handle ctrl-c

**Description**

- Handle ctrl+c

**Related issue(s)**

#370

### DIFF
--- a/cmd/main.go
+++ b/cmd/main.go
@@ -23,10 +23,10 @@ func main() {
 
 func SetupCloseHandler() {
 	c := make(chan os.Signal, 2)
-	signal.Notify(c, os.Interrupt, syscall.SIGINT)
+	signal.Notify(c, syscall.SIGTERM, syscall.SIGINT)
 	go func() {
-		<-c
-		fmt.Println("\r- Ctrl+C pressed in Terminal. Exiting...")
+		sig := <-c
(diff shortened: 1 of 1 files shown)


cmd/main.go
```diff
@@ -1,17 +1,32 @@
...
 	err := command.Execute()
 	if err != nil {
 		os.Exit(1)
 	}
+
+}
+
+func SetupCloseHandler() {
+	c := make(chan os.Signal, 2)
+	signal.Notify(c, os.Interrupt, syscall.SIGINT)
```
clebs: The signals `os.Interrupt` and `syscall.SIGINT` are the same, please remove one of them.
See golang's `os` snippet here:
https://golang.org/pkg/os/#Signal

Also, we should probably add `SIGTERM` as well to the signals we react on.

cmd/main.go
```diff
@@ -1,17 +1,32 @@
...
 	command := kyma.NewCmd(cli.NewOptions())
 
 	err := command.Execute()
 	if err != nil {
 		os.Exit(1)
 	}
+
+}
+
+func SetupCloseHandler() {
```
clebs: Probably this function does not need to be exported, since it won't be used anywhere else.

PR #383, opened 2020-02-21
Remove check on helm-certs-job

kfurgol:
**Description**

- Removed check that waits for helm-certs-job since it was removed [here](https://github.com/kyma-project/kyma/pull/7262)

**Related issue(s)**

### DIFF
--- a/pkg/installation/installation.go
+++ b/pkg/installation/installation.go
@@ -17,7 +17,6 @@ import (
 	"github.com/pkg/errors"
 	"gopkg.in/yaml.v2"
 	"k8s.io/apimachinery/pkg/types"
-
 	corev1 "k8s.io/api/core/v1"
 	apiErrors "k8s.io/apimachinery/pkg/api/errors"
 	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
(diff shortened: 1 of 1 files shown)


pkg/installation/installation.go
```diff
@@ -696,3 +681,16 @@ func (i *Installation) releaseSrcFile(path string) string {
 func (i *Installation) releaseFile(path string) string {
 	return fmt.Sprintf(releaseResourcePattern, i.Options.configVersion, path)
 }
+
+func (i *Installation) getHelmSecret(interval time.Duration,retries int) (*corev1.Secret, error){
```
pkosiec: It would be better to use some generic retry package, and then call `getHelmSecret` method, wrapping it with the retry function. Then, the `getHelmSecret` would get the secret only once. Something like this:

```go
var secret *corev1.Secret
retry.Do(func() error {
               secret, err = i.getHelmSecret
	}, retry.Attempts(s.retryAttempts), retry.Delay(retryDelayMilliseconds*time.Millisecond))
```

In this case, the `"github.com/avast/retry-go"` package is used.

Issue #333, opened 2019-12-10
Kyma CLI should validate input params like --override


dbadura:
**Description**
Currently Kyma CLI do not check if file passed to `--override` param exists.

**Expected result**
Kyma CLI should fail with message indicating that file targeting by `--override` param, not exist or is not valid, before doing any installation.
Kyma CLI before installation any components should print all `override` configuration files which will be applied later.

**Actual result**
Kyma install all components without indicating that `override` file wasn't read.

**Steps to reproduce**
run 
`kyma install -o something`

**Troubleshooting**

a-thaler:
Hi @dbadura, thanks for the feedback, I agree that a check needs to be added and overall transparency on what is happening should be improved

a-thaler:
The error message in case an override is present already is wrong and misleading, could you please fix it as well.

a-thaler:
I was wrong, all is fine

PR #386, opened 2020-02-24
Dont continue when there is some error in overrides

rakesh-garimella:
**Description**

- Throw errors when override error occurs.

**Related issue(s)**

#333


pkg/installation/installation.go
```diff
@@ -440,8 +440,8 @@ func (i *Installation) applyOverrideFiles() error {
 				"-p",
 				c)
 			if err != nil {
-				fmt.Printf("unable to override values. File: %s. Error: %s\n", file, err.Error())
-				continue
+				msg := fmt.Sprintf("unable to override values. File: %s. Error: %s\n", file, err.Error())
```
clebs: Please use `errors.Wrap()` instead of concatenating errors.

pkg/installation/installation.go
```diff
@@ -368,13 +368,13 @@ func (i *Installation) applyOverrideFiles() error {
 		if err != nil {
 			fmt.Printf("unable to open file: %s. error: %s\n",
```
clebs: If returning the error, no need to print it here.

Just wrap it and return so that the installation step can print it and fail properly.

pkg/installation/installation.go
```diff
@@ -366,15 +366,13 @@ func (i *Installation) applyOverrideFiles() error {
 	for _, file := range i.Options.OverrideConfigs {
 		oFile, err := os.Open(file)
 		if err != nil {
-			fmt.Printf("unable to open file: %s. error: %s\n",
+			return errors.Wrapf(err, "unable to open file: %s. error: %s\n",
```
clebs: Like this, the error message will appear twice.
this:
`return errors.Wrapf(err, "unable to read data from file: %s. error: %s\n", file, err.Error())`
should be:
`return errors.Wrapf(err, "unable to read data from file: %s.", file)`

clebs: Like this, the error message will appear twice.
this:
`return errors.Wrapf(err, "unable to read data from file: %s. error: %s\n", file, err.Error())`
should be:
`return errors.Wrapf(err, "unable to read data from file: %s.", file)`

pkg/installation/installation.go
```diff
@@ -442,8 +428,8 @@ func (i *Installation) applyOverrideFiles() error {
 				"-p",
 				c)
 			if err != nil {
-				fmt.Printf("unable to override values. File: %s. Error: %s\n", file, err.Error())
-				continue
+				msg := fmt.Sprintf("unable to override values. File: %s.\n", file)
+				return errors.New(msg)
```
clebs:
```suggestion
				return fmt.Errorf("unable to override values. File: %s.\n", file)
```

Issue #391, opened 2020-02-27
Pass the new control plane zone by default if not passed exclusively


rakesh-garimella:
**Description**

We should have a default value for GCP control plane zone if it has not been passed via `--extra` flag

PR #392, opened 2020-02-27
Pass default value for control plane zone in gcp

rakesh-garimella:
**Description**
Pass a default value for the control plane zone if it has not been passed exclusively

**Related issue(s)**

#391 #393


cmd/kyma/provision/gardener/cmd.go
```diff
@@ -142,6 +142,8 @@ func newProvider(o *Options) (*types.Provider, error) {
 		CredentialsFilePath: o.CredentialsFile,
 	}
 
+	gcpControlPlaneSet := false
```
clebs: Instead of having this boolean and all that logic, would it not be easier, before looping through extra options just always assign `Zone[0]` and in case there is an extra config for it, it will get overwritten?
```suggestion
	if o.TargetProvider == "gcp" {
		p.CustomConfigurations["gcp_control_plane_zone"] = o.Zone[0]
	}
```

cmd/kyma/provision/gardener/cmd.go
```diff
@@ -168,8 +170,14 @@ func newProvider(o *Options) (*types.Provider, error) {
 		if len(v) != 2 {
 			return p, fmt.Errorf("wrong format for extra configuration %s. Please provide NAME=VALUE pairs", e)
 		}
+		if v[0] == "gcp_control_plane_zone" && v[1] != "" && o.TargetProvider == "gcp" {
+			gcpControlPlaneSet = true
+		}
 		p.CustomConfigurations[v[0]] = v[1]
 	}
+	if !gcpControlPlaneSet && o.TargetProvider == "gcp" {
+		p.CustomConfigurations["gcp_control_plane_zone"] = o.Zone[0]
```
clebs: If `Zone` now is a slice, should it not be maybe renamed to `Zones`?

cmd/kyma/provision/gardener/cmd.go
```diff
@@ -160,14 +164,17 @@ func newProvider(o *Options) (*types.Provider, error) {
...
-	p.CustomConfigurations["zone"] = o.Zone
+	p.CustomConfigurations["zone"] = o.Zones
 
 	for _, e := range o.Extra {
 		v := strings.Split(e, "=")
 
 		if len(v) != 2 {
 			return p, fmt.Errorf("wrong format for extra configuration %s. Please provide NAME=VALUE pairs", e)
 		}
+		if v[0] == "gcp_control_plane_zone" && v[1] != "" && o.TargetProvider == "gcp" {
```
clebs: This whole `if` block can just be deleted now.

cmd/kyma/provision/gardener/cmd.go
```diff
@@ -49,7 +49,7 @@ Use the following instructions to create a service account for a selected provid
 	cmd.Flags().StringVarP(&o.Secret, "secret", "s", "", "Name of the Gardener secret used to access the target provider. (required)")
 	cmd.Flags().StringVarP(&o.KubernetesVersion, "kube-version", "k", "1.16", "Kubernetes version of the cluster.")
 	cmd.Flags().StringVarP(&o.Region, "region", "r", "europe-west3", "Region of the cluster.")
-	cmd.Flags().StringSliceVarP(&o.Zone, "zone", "z", []string{"europe-west3-a"}, "Zone is a list of availability zones that are used to evenly distribute the worker pool. eg. --zone=\"europe-west3-a,europe-west3-b\"")
+	cmd.Flags().StringSliceVarP(&o.Zones, "zones", "z", []string{"europe-west3-a"}, "Zones is a list of availability zones that are used to evenly distribute the worker pool. eg. --zones=\"europe-west3-a,europe-west3-b\"")
```
bszwarc:
```suggestion
	cmd.Flags().StringSliceVarP(&o.Zones, "zones", "z", []string{"europe-west3-a"}, "Zones specify availability zones that are used to evenly distribute the worker pool. eg. --zones=\"europe-west3-a,europe-west3-b\"")
```

PR #405, opened 2020-03-09
Introduce a new source mode: latest-published

### LINKED ISSUE #206: Installing release "master" can ends up with ImagePullBackOff for kyma-installer
In job `pre-master-compass-integration` we install kyma from release "master"

`kyma install -o $INSTALLER_CR_PATH --release master`
This cause that `installInstallerFromRelease` is executed, which calls `getMasterHash` method, which is implemented in the following way:
```
func (cmd *command) getMasterHash() (string, error) {
	ctx, timeoutF := context.WithTimeout(context.Background(), 1*time.Minute)
	defer timeoutF()
	r, err := git.CloneContext(ctx, memory.NewStorage(), nil,
		&git.CloneOptions{
			Depth: 1,
			URL:   "https://github.com/kyma-project/kyma",
		})
	if err != nil {
		return "", err
	}
	h, err := r.Head()
	if err != nil {
		return "", err
	}

	return h.Hash().String()[:8], nil
}
```

Such an approach can cause a problem if job `post-master-kyma-development-artifacts` fails to build kyma-installer image for the recent commit. This happened recently: https://storage.googleapis.com/kyma-prow-logs/logs/post-master-kyma-development-artifacts/1162292035089076225/build-log.txt
where there was a problem with communication with github.

-----

Suggested solution by @clebs to be implemented:

- Check if the installer image for the current master exists.
- If it does not exist for whatever reason, the previous hash will be taken and the user will be informed with a warning that the CLI resorted to the previous master image.
- There will be a depth of 3 commit hashes to try to find the image, if none is found the CLI will abort installation.

aszecowka:
**Description**

- `source` flag accepts a new value: `latest-published`
- a new flag `fallbackLevel` is added, it accepts ints, default value is 5, it is taken into account only if `source=latest-published`
- if `fallbackLevel` is provided, we check if artifacts for the given version are published, but only `fallbackLevel+1` commits are checked, if none of them exists, we fail fast. 

**Related issue(s)**

fixes #206


cmd/kyma/install/cmd.go
```diff
@@ -100,13 +100,14 @@ The standard installation uses the minimal configuration. The system performs th
...
 	- To use the latest master, write "kyma install --source=latest".
+	- To use the latest published master, write "kyma install --source=latest-published".
 	- To use the local sources, write "kyma install --source=local". 
 	- To use a custom installer image, write kyma "install --source=user/my-kyma-installer:v1.4.0".`)
 	cobraCmd.Flags().StringVarP(&o.LocalSrcPath, "src-path", "", "", "Absolute path to local sources.")
 	cobraCmd.Flags().DurationVarP(&o.Timeout, "timeout", "", 1*time.Hour, "Time-out after which CLI stops watching the installation progress.")
 	cobraCmd.Flags().StringVarP(&o.Password, "password", "p", "", "Predefined cluster password.")
 	cobraCmd.Flags().StringArrayVarP(&o.OverrideConfigs, "override", "o", nil, "Path to a YAML file with parameters to override.")
-
+	cobraCmd.Flags().IntVar(&o.FallbackLevel, "fallbackLevel", 5, "If source=latest-published, defines how many commits from master branch are taken into account if artifacts for newer commits does not exist yet")
```
bszwarc:
```suggestion
	cobraCmd.Flags().IntVar(&o.FallbackLevel, "fallbackLevel", 5, `If "source=latest-published", defines the number of commits from master branch taken into account if artifacts for newer commits do not exist yet`)
```

cmd/kyma/install/cmd.go
```diff
@@ -100,13 +100,14 @@ The standard installation uses the minimal configuration. The system performs th
 	cobraCmd.Flags().StringVarP(&o.Source, "source", "s", DefaultKymaVersion, `Installation source. 
 	- To use the specific release, write "kyma install --source=1.3.0".
 	- To use the latest master, write "kyma install --source=latest".
+	- To use the latest published master, write "kyma install --source=latest-published".
```
clebs: Here it might not be clear what "latest published" means. Maybe we should detail what will happen.
```suggestion
	- To use the latest published master, which is the latest commit with released images, write "kyma install --source=latest-published".
```

pkg/installation/installation.go
```diff
@@ -197,10 +200,19 @@ func (i *Installation) validateConfigurations() error {
...
+			return fmt.Errorf("unable to get latest version of kyma: %s", err.Error())
+		}
+		i.Options.releaseVersion = fmt.Sprintf("master-%s", latest)
+		i.Options.configVersion = "master"
+		i.Options.registryTemplate = registryImagePattern
+
+	case strings.EqualFold(i.Options.Source, sourceLatestPublished):
+		latest, err := i.getLatestAvailableMasterHash()
+		if err != nil {
+			return fmt.Errorf("unable to get latest published version of kyma: %s", err.Error())
```
clebs: Please use `errors.Wrap()` instead.

pkg/installation/installation.go
```diff
@@ -217,7 +229,7 @@ func (i *Installation) validateConfigurations() error {
 		i.Options.remoteImage = i.Options.Source
 		i.Options.configVersion = "master"
 	default:
-		return fmt.Errorf("Failed to parse the source flag. It can take one of the following: 'local', 'latest', release version (e.g. 1.4.1), or installer image")
+		return fmt.Errorf("failed to parse the source flag. It can take one of the following: 'local', 'latest', release version (e.g. 1.4.1), or installer image")
```
clebs:
```suggestion
		return fmt.Errorf("failed to parse the source flag. It can take one of the following: 'local', 'latest', 'latest-published', release version (e.g. 1.4.1), or installer image")
```

Issue #426, opened 2020-04-06
kyma provision minikube -  is not relaying output errors


jcawley5:
If I run 
kyma provision minikube
it will sit at **Create Minikube instance: Start Minikube** forever.

After killing it and running
minikube start
I will get an error that sudo permissions need to be applied to particular items

After fixing the minikube config everything will work, but the kyma commands should be outputting the minikube errors.

suleymanakbas91:
We introduced a timeout for all the minikube commands with [this change](09c114289b673efb28bebd2aadfa2de391b66c1c), so it won't run forever. We also added a note on the timeout message to `try running the command manually` so to solve any sudo permission problem. That looks like the only solution right now.

a-thaler:
I tried it from home having a slack call in parallel. I started with empty minikube folder, so the ISO needs to be downloaded and I was running into the 3min timeout. I have no chance to change the timeout.
As the download already had 90% I think an increase to 5min will be good and making the option configurable already

PR #437, opened 2020-04-21
Add timeout for minikube commands

suleymanakbas91:
**Description**

- Add timeout for minikube commands

**Related issue(s)**

See #426

### DIFF
--- a/internal/minikube/minikube.go
+++ b/internal/minikube/minikube.go
@@ -1,11 +1,13 @@
 package minikube
 
 import (
+	"context"
 	"fmt"
 	"os"
 	"os/exec"
 	"regexp"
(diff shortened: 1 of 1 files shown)


internal/minikube/minikube.go
```diff
@@ -17,16 +19,25 @@ const (
 
 //RunCmd executes a minikube command with given arguments
 func RunCmd(verbose bool, profile string, rawArgs ...string) (string, error) {
+	ctx, cancel := context.WithTimeout(context.Background(), 2*time.Minute)
```
clebs, line 22: Will this timeout work with `minikube start`?
Maybe we should use the timeout that the cli command has?

PR #444, opened 2020-04-27
Make go gettable

k15r:
**Description**

- move contents of cmd/kyma to cmd/kyma/kymacmd

**Related issue(s)**


k15r:
> IMHO is `cmd/kyma/kymacmd` a really strange package name and it also adds yet more depth to the package structure.

You're totally correct, i chose it just because i did want to only make minimal changes to your structure.
 
> I would rather move the main file into the `cmd/kyma` package and just have the main file sit next to the kyma command file. That way the package structure remains unchanged but the go get and install work.

also fine for me

clebs:
After discussing the pros and cons of this change, we have decided that it is better to keep the structure as it is, even if it means that it is not installable via `go get`.

Issue #305, opened 2019-11-21
Use install package of hydroform


a-thaler:
**Description**
In [hydroform](https://github.com/kyma-incubator/hydroform) there is a new install module available simplifying management of the kyma-installer. This should be used instead of custom code. So please adjust the install command to use the install module

**Reasons**
re-usability, less code base

PR #470, opened 2020-06-09
Use hydroform installation package

suleymanakbas91:
**Description**

- Use hydroform installation package

**Related issue(s)**

See https://github.com/kyma-project/cli/issues/305


pkg/installation/service.go
```diff
@@ -0,0 +1,159 @@
...
+func (s *installationService) CheckInstallationState(kubeconfig *rest.Config) (installation.InstallationState, error) {
+	return installation.CheckInstallationState(kubeconfig)
+}
+
+func (s *installationService) TriggerUninstall(kubeconfig *rest.Config) error {
+	return installation.TriggerUninstall(kubeconfig)
+}
+
+func (s *installationService) waitForInstallation(currentStep step.Step, stateChannel <-chan installation.InstallationState, errorChannel <-chan error) error {
+	for {
```
clebs, line 136: How do we know here when we get a timeout from the context?
And how do we print that out to the user?

Issue #473, opened 2020-06-17
Concatenation issue after migrating to Hydroform


pkosiec:
**Description**

After migration to Hydroform, CLI no longer concatenates passed configuration files correctly.

In CLI you use ---\n to concatenate all files:
```go
mergedInstallerFileContent := files[installerFile].StringContent + "---\n" + files[installerCRFile].StringContent
```

 but Hydroform splits them with `\n---\n`:

```go
func ParseYamlToK8sObjects(decoder runtime.Decoder, yamlContent string) ([]K8sObject, error) {
	resources := strings.Split(yamlContent, "\n---\n")
```

In case a file doesn’t have empty line at the end, it fails. (Another side question is if it is really necessary to concatenate config files and split them again)

suleymanakbas91:
Solved by #474.

PR #474, opened 2020-06-17
Add components flag

suleymanakbas91:
**Description**

- Add components flag https://github.com/kyma-project/kyma/issues/8290
- Solve concatenation problem https://github.com/kyma-project/cli/issues/473


crabtree:
Hi @suleymanakbas91, I just tested this PR together with recent changes in the Compass installation process. Overall it works. Just wondering whether the `-c` parameter can handle full installation CR not only the components list, and having that done the download of the installation CR is redundant.

PR #478, opened 2020-06-22
Add commit hash source option

suleymanakbas91:
**Description**

- Kyma CLI now accepts Kyma master commit hash as a source: 
`kyma install -s 34edf09acb15f44b937cf2ca5a6f195db6d3b0ed`
or
`kyma install -s 34edf09a`

**Related issue(s)**

See https://github.com/kyma-project/cli/issues/479


pkg/installation/installation.go
```diff
@@ -237,12 +237,18 @@ func (i *Installation) validateConfigurations() error {
 		i.Options.configVersion = i.Options.Source
 		i.Options.registryTemplate = registryImagePattern
 
+	//Install the specific commit version (ex: 1.3.0)
```
clebs, line 240: Example here should be a hash

Issue #492, opened 2020-07-08
Increase Code Coverage


suleymanakbas91:
**Description**

As Kyma CLI is documented now as the only way to install Kyma, we should increase the overall quality and robustness. One step for this would be increasing code coverage by adding new unit tests. The current status is as follows:

<img width="782" alt="Screenshot 2020-07-08 at 12 29 48" src="https://user-images.githubusercontent.com/5543238/86909489-72588080-c118-11ea-9766-021623505803.png">

As a concrete aim for the task, we can say: 

- All the packages will have unit tests with a minimum of 30% coverage. 

>**NOTE:** This number should not make you ignore the packages with already more than 30% coverage as they might have important parts that are not covered. 

**Hint:**

You can use the following commands to see the current coverage on a browser:

```bash
go test ./... -coverprofile cp.out
go tool cover -html=cp.out
```

**Tasks**
Prerequisite:

- [x] Create mocks for all interfaces used in install package (use mockery).

Parallel tasks (1 file per developer):
- [x] cluster.go (Borja)
- [x] http.go (Borja)
- [x] installation.go (Suleyman)
- [x] service.go (Borja)
- [x] upgrade.go (Borja)
- [x] utils.go (Suleyman)
- [ ] docker.go (rakesh)

suleymanakbas91:
As a result of the new tests:
- Overall test coverage increased from `31.5%` to `52.2%`.
- Installation package coverage increased from `10.1%` to `67.9%`.

PR #529, opened 2020-07-29
Increase code coverage

rakesh-garimella:
**Description**

- Add tests for docker file
- Refactor the code so that its better testable
- Move docker to a separate module.

**Related issue(s)**

https://github.com/kyma-project/cli/issues/492


pkg/docker/docker.go
```diff
@@ -18,41 +18,91 @@ import (
...
+		return nil, err
 	}
 
-	reader, err := archive.TarWithOptions(i.Options.LocalSrcPath, &archive.TarOptions{})
+	return &dockerClient{
+		dClient,
+	}, nil
+}
+
+func NewKymDockerClientService(isLocal bool, verbosity bool, profile string, timeout time.Duration) (KymaDockerService, error) {
```
clebs, line 79: There is a typo here:
```suggestion
func NewKymaDockerClientService(isLocal bool, verbosity bool, profile string, timeout time.Duration) (KymaDockerService, error) {
```

pkg/docker/docker.go
```diff
@@ -18,41 +18,91 @@ import (
...
+	*docker.Client
+	// DockerClientService
+}
+
+type kymaDockerClient struct {
+	kymaDocker ClientService
+}
+
+//go:generate mockery --name ClientService
+type ClientService interface {
```
clebs, line 38: This interface name is too generic. I would suggest something like `DockerClient`.

pkg/installation/installation.go
```diff
@@ -39,9 +40,10 @@ type ComponentsConfig struct {
 
 // Installation contains the installation elements and configuration options.
 type Installation struct {
+	Docker      docker.KymaDockerService
 	K8s         kube.KymaKube
 	Service     Service
-	currentStep step.Step
+	CurrentStep step.Step
```
clebs, line 46: Why is this field now exported?

pkg/docker/docker.go
rakesh-garimella: I tried DockerClient but golint say people would stutter while reading it and suggested to change it to ClientService.

pkg/installation/installation.go
rakesh-garimella: its used in the Docker package with in pushKymaInstaller method so needed to be exported

pkg/docker/docker.go
clebs: Of course, because of being in the context of the docker package. In that case, the interface would be just `Client`.
Same for the `KymaDockerService`.

I would also drop the `Service` suffix alltogether since it does not have any meaning.

pkg/docker/docker.go
```diff
@@ -18,41 +18,91 @@ import (
...
 
-// DockerErrorMessage is used to parse error messages coming from Docker
-type DockerErrorMessage struct {
+type dockerClient struct {
+	*docker.Client
+	// DockerClientService
+}
+
+type kymaDockerClient struct {
+	kymaDocker ClientService
```
clebs, line 34: This is the real docker client, not the Kyma client. This makes it all very confusing and misleading.

pkg/docker/docker.go
```diff
@@ -18,41 +18,91 @@ import (
...
 
 const (
 	defaultRegistry = "index.docker.io"
 )
 
-// DockerErrorMessage is used to parse error messages coming from Docker
-type DockerErrorMessage struct {
+type dockerClient struct {
+	*docker.Client
+	// DockerClientService
```
clebs, line 30: Please remove this comment.

pkg/docker/docker.go
```diff
@@ -18,41 +18,91 @@ import (
...
 
-func (i *Installation) buildKymaInstaller(imageName string) error {
-	var dc *docker.Client
-	var err error
-	if i.Options.IsLocal {
-		dc, err = minikube.DockerClient(i.Options.Verbose, i.Options.LocalCluster.Profile, i.Options.Timeout)
-	} else {
-		dc, err = docker.NewClientWithOpts(docker.FromEnv)
+//NewDockerService creates docker client using docker environment of the OS
+func NewDockerService() (ClientService, error) {
```
clebs, line 56:
```suggestion
func NewClient() (ClientService, error) {
```

pkg/docker/docker.go
```diff
@@ -18,41 +18,91 @@ import (
...
+		return nil, err
 	}
+
+	return &dockerClient{
+		dClient,
+	}, nil
+}
+
+//NewDockerMinikubeService creates docker client for minikube docker-env
+func NewDockerMinikubeService(verbosity bool, profile string, timeout time.Duration) (ClientService, error) {
```
clebs, line 68:
```suggestion
func NewMinikubeClient(verbosity bool, profile string, timeout time.Duration) (ClientService, error) {
```

pkg/installation/installation.go
clebs: You are now passing it as a parameter to `PushKymaInstaller ` so it does not need to be exported

pkg/docker/docker.go
rakesh-garimella: fixed.

pkg/docker/docker.go
rakesh-garimella: fixed.

pkg/docker/docker.go
rakesh-garimella: done

pkg/docker/docker.go
rakesh-garimella: done.

pkg/docker/docker.go
rakesh-garimella: done.

pkg/docker/docker.go
rakesh-garimella: done.

pkg/installation/installation.go
rakesh-garimella: yeah fixed this.

Issue #490, opened 2020-07-07
Enhance the print summary with the installation duration  


sayanh:
**Description**

It'd cool to have the duration of installation in the [printsummary](https://github.com/kyma-project/cli/blob/master/cmd/kyma/install/cmd.go#L280).

**Reasons**

This will let us compare the platforms where the installation takes longer.

a-thaler:
Looks like the duration is not based on the actual installation time on server side but is counted on client side. When getting disconnected that will not work out.
To prove it I just called install command on existing installation and it prints
```
Kyma installation took:		0 hours 0 minutes
```

a-thaler:
There is no easy solution to achieve that, so for now we will stick to the local CLI time only

PR #534, opened 2020-07-30
Print the installation duration

suleymanakbas91:
**Description**

- Print the installation duration

**Related issue(s)**

See https://github.com/kyma-project/cli/issues/490

### DIFF
--- a/pkg/installation/installation.go
+++ b/pkg/installation/installation.go
@@ -71,6 +71,8 @@ type Result struct {
 	AdminPassword string
 	// Warnings includes a set of any warnings from the installation.
 	Warnings []string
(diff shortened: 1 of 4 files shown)


pkg/installation/installation.go
```diff
@@ -405,7 +412,7 @@ func (i *Installation) waitForInstaller() error {
 	}
 }
 
-func (i *Installation) buildResult() (*Result, error) {
+func (i *Installation) buildResult(duration time.Duration) (*Result, error) {
```
clebs, line 415: I am wondering if this duration and all the time measuring logic around it should be inside the `Installation`type 🤔 

But it is perfectly fine as is.

Issue #493, opened 2020-07-09
Minikube installation not working when DNS Port(Port 53) is being used by another service


rakesh-garimella:
**Description**
When we have dns port being used by another service eg. Cisco Umbrella the kyma installation is not working on minikube.
The installer is stuck at installing tiller.

**Expected result**

Kyma should be able to install in such scenarios too.

**Actual result**
Installation stuck in installing tiller.

**Steps to reproduce**

Try to install kyma on minikube using cli on a mac running cisco umbrella.

**Hints for solution**
One can start minikube with following flag as workaround: 

```
minikube start --hyperkit-vpnkit-sock=/.../vpnkit.eth.sock
```

Also look here: https://github.com/kubernetes/minikube/issues/3036#issuecomment-574943561

a-thaler:
Seems that there is a problem when using the flag: https://github.com/kyma-project/kyma/issues/9439, please double-check before merging

PR #555, opened 2020-08-14
Forward the dns requests to dns-proxy if present

rakesh-garimella:
**Description**

- There can be instances when there is a dns-proxy eg Cisco Umbrella running on the dns port (Port 53). This is usually coming from corporate requirements. In such a  case we should use this proxy to resolve the dns-requests.
- If we dont use the dns-proxy then it can lead to failure of dns resolution by minikube  and thus the images wont be pulled from the registry.

**Related issue(s)**
#493


cmd/kyma/provision/minikube/cmd.go
```diff
@@ -70,6 +70,7 @@ func NewCmd(o *Options) *cobra.Command {
 	cmd.Flags().StringVar(&o.CPUS, "cpus", "4", "Specifies the number of CPUs used for installation.")
 	cmd.Flags().StringVar(&o.Profile, "profile", "", "Specifies the Minikube profile.")
 	cmd.Flags().DurationVar(&o.Timeout, "timeout", 5*time.Minute, `Maximum time during which the provisioning takes place, where "0" means "infinite". Valid time units are "ns", "us" (or "µs"), "ms", "s", "m", "h".`)
+	cmd.Flags().BoolVar(&o.UseVPNKitSock, "use-vpnkit-sock", false, `Uses vpnkit sock provided by Docker. This is useful when DNS Port (53) is being used by some other program like dns-proxy (eg. provided by Cisco Umbrella).`)
```
bszwarc, line 73:
```suggestion
	cmd.Flags().BoolVar(&o.UseVPNKitSock, "use-vpnkit-sock", false, `Uses vpnkit sock provided by Docker. This is useful when the DNS Port (53) is used by some other program like dns-proxy (e.g. provided by Cisco Umbrella).`)
```

cmd/kyma/provision/minikube/cmd.go
```diff
@@ -236,6 +237,15 @@ func (c *command) startMinikube() error {
 	if c.opts.VMDriver == vmDriverHyperv {
 		startCmd = append(startCmd, "--hyperv-virtual-switch="+c.opts.HypervVirtualSwitch)
 	}
+
+	if c.opts.UseVPNKitSock {
```
a-thaler, line 241: The "whoami" command is OS specific, the pathToVPNKitSock is OS specific, the --hyperkit-vpnkit-sock flag is MacOS specific as hyperkit is not supported by other platforms. I think you need to make it all MacOS specific so that it is not available at all for other platforms.

cmd/kyma/provision/minikube/cmd.go
```diff
@@ -70,6 +70,7 @@ func NewCmd(o *Options) *cobra.Command {
 	cmd.Flags().StringVar(&o.CPUS, "cpus", "4", "Specifies the number of CPUs used for installation.")
 	cmd.Flags().StringVar(&o.Profile, "profile", "", "Specifies the Minikube profile.")
 	cmd.Flags().DurationVar(&o.Timeout, "timeout", 5*time.Minute, `Maximum time during which the provisioning takes place, where "0" means "infinite". Valid time units are "ns", "us" (or "µs"), "ms", "s", "m", "h".`)
+	cmd.Flags().BoolVar(&o.UseVPNKitSock, "use-vpnkit-sock", false, `Uses vpnkit sock provided by Docker. This is useful when the DNS Port (53) is used by some other program like dns-proxy (e.g. provided by Cisco Umbrella).`)
```
a-thaler, line 73: That is the minikube docu saying: "--hyperkit-vpnkit-sock='': Location of the VPNKit socket used for networking. If empty, disables Hyperkit
VPNKitSock, if 'auto' uses Docker for Mac VPNKit connection, otherwise uses the specified VSock (hyperkit driver only)"

Have you checked if you can have it on by default? Have you checked if it works with "auto", maybe even for all cases?
I'm thinking if we should maybe add "hyperkit" to the naming to build the bridge to the related minikube flag

cmd/kyma/provision/minikube/cmd.go
rakesh-garimella: fixed.

cmd/kyma/provision/minikube/cmd.go
rakesh-garimella: `auto` is looking for the socket at a different path so its not working. Updated the name of the flag

cmd/kyma/provision/minikube/cmd.go
clebs: This is also my thought, the code will be compiled into all OS versions as it is now. My idea when we discussed this was to move the OS specific logic out so that it only compiles in Mac OSX.

How it is currently will not crash because of the condition being only true on Mac OSX but it is not the best way to achieve it. As much as possible should be on the OS specific code, not just a boolean. 

Also, the boolean `allowVPNSock` can be completely replaced by `runtime.GOOS == "darwin"`) so it is not necessary at all.

clebs (approved):

cmd/kyma/provision/minikube/cfg_darwin.go
```diff
@@ -2,4 +2,28 @@
...
+	"fmt"
+	"strings"
+
+	"github.com/kyma-project/cli/internal/cli"
+	"github.com/spf13/cobra"
+)
+
 const defaultVMDriver = vmDriverHyperkit
+
+func osSpecificFlags(o *Options, cmd *cobra.Command) {
```
clebs, line 15: OS specific flags is no longer needed, since the flag is declared in all platforms.

cmd/kyma/provision/minikube/cfg_windows.go
```diff
@@ -2,4 +2,21 @@
...
+
 const defaultVMDriver = vmDriverVirtualBox
+
+func osSpecificFlags(o *Options, cmd *cobra.Command) {
+	cmd.Flags().BoolVar(&o.UseVPNKitSock, "use-hyperkit-vpnkit-sock", false, `Uses vpnkit sock provided by Docker. This is useful when DNS Port (53) is being used by some other program like dns-proxy (eg. provided by Cisco Umbrella.  This flag works only on Mac OS).`)
+}
+
+func osSpecificRun(c *command, startCmd []string) ([]string, error) {
+	if c.opts.UseVPNKitSock {
+		fmt.Println("Warning: This flag is only supported for Mac OS!")
```
clebs, line 19: Please use the `nice.PrintImportant()` function so that the warning is printed in a bold yellow font.
Same as here: https://github.com/kyma-project/cli/blob/1b083c39f7bfe5d433edf9a1799e946e095c8ad0/cmd/kyma/create/system/system.go#L66

clebs (approved):

alexandra-simeonova (approved):

Issue #174, opened 2019-07-12
Add 'upgrade' command to update to latest version


a-thaler:
**Description**
Have an `upgrade` command to update to the newest Kyma version easily.
It should follow mainly this guide: https://kyma-project.io/docs/root/kyma/#installation-upgrade-kyma

Typical user flow can look like this:
1. Apply migration-guide of latest kyma 
1. Download newest CLI supporting latest kyma in version
2. Call `kyma upgrade``
   - Check CLI and cluster version, if version mismatch bigger than one minor version, print error
   - Prompt for migration guide of latest version and ask if it was applied
   - Trigger upgrade using hydroform
   - Watch installation status

Provide option to apply installer CR and overrides which might be needed for custom components or new required overrides.

Please assure to not duplicate code for the status loop. Waiting for upgrade status is very similar to installation status.

**Reasons**
Convenient way to upgrade kyma

a-thaler:
The implementation does not provide the --source flag. With that it is hard to test the upgrade at all as part of the regular prow upgrade jobs.
Of course, it is hard to tell if a local source is older then the current one, but for non-release versions, we might print just a big warning and confirmation dialog and still do the upgrade.

PR #559, opened 2020-08-17
Improve upgrade command

shorim:
**Description**

- Add the following flags for the upgrade command: `source`, `src-path`, `fallbackLevel` and `custom-image`.
- If the current Kyma version or the target version is a non-release version, then print a warning for the user informing him that it is not possible to check the upgrade compatibility. In this case, upgrade is performed only if the user confirms to continue.
- Allow to upgrade even if the target version is the same as the current version in case that an override is provided

**Related issue(s)**

#174

### DIFF
--- a/cmd/kyma/upgrade/cmd.go
+++ b/cmd/kyma/upgrade/cmd.go
@@ -34,14 +34,14 @@ func NewCmd(o *Options) *cobra.Command {
 	cobraCmd := &cobra.Command{
 		Use:   "upgrade",
 		Short: "Upgrades Kyma",
-		Long:  `Use this command to upgrade the Kyma version on a cluster. If no source is specified, then Kyma is upgraded to the same version as the CLI`,
+		Long:  `Use this command to upgrade the Kyma version on a cluster.`,
 		RunE:  func(_ *cobra.Command, _ []string) error { return cmd.Run() },
 	}
 
 	cobraCmd.Flags().BoolVarP(&o.NoWait, "noWait", "n", false, "Determines if the command should wait for the Kyma upgrade to complete.")
 	cobraCmd.Flags().StringVarP(&o.Domain, "domain", "d", defaultDomain, "Domain used for the upgrade.")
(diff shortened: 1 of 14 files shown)


cmd/kyma/upgrade/cmd.go
```diff
@@ -33,20 +33,29 @@ func NewCmd(o *Options) *cobra.Command {
...
+		Short: "Upgrades Kyma",
+		Long:  `Use this command to upgrade the Kyma version on a cluster. If no source is specified, then Kyma is upgraded to the same version as the CLI`,
 		RunE:  func(_ *cobra.Command, _ []string) error { return cmd.Run() },
 	}
 
 	cobraCmd.Flags().BoolVarP(&o.NoWait, "noWait", "n", false, "Determines if the command should wait for the Kyma upgrade to complete.")
 	cobraCmd.Flags().StringVarP(&o.Domain, "domain", "d", defaultDomain, "Domain used for the upgrade.")
 	cobraCmd.Flags().StringVarP(&o.TLSCert, "tlsCert", "", "", "TLS certificate for the domain used for the upgrade. The certificate must be a base64-encoded value.")
+	cobraCmd.Flags().StringVarP(&o.Source, "source", "s", "", `Upgrade source. 
+	- To use the specific release, write "kyma upgrade --source=1.3.0".
```
rakesh-garimella, line 45: should we call it target instead of source ? Usually in target you provide a target version one wants to upgrade to

shorim: I think that `target` is a better name for the upgrade command, but then should it be `target-path` instead of `src-path`?

shorim: @a-thaler what is your opinion here?
Should we have the flags named as `source` and `src-path` the same way as in the install command
or is it better to name them `target` and `target-path`?

shorim: Let us leave it now consistent with the install command and if we got feedback that it is confusing, then we could change it

rakesh-garimella (approved):

bszwarc (approved):

PR #560, opened 2020-08-18
Support matchlabelexpression for test definitions

k15r:
**Description**

- the cli only allows you to specify specific testdefinitions by names for cts creation. Octopus already allows you to use a label expression to select which test definitions should be part of this cts. By introducing a flag called `--label-expression` we can now use this feature in the kyma cli

eg.
```
 kyma test run -l kyma-project.io/test.before-upgrade=true
```

this will run all test definitions with the label `kyma-project.io/test.before-upgrade=true`
This feature can be used in combination with the already existing `matchNames` selector.

**Related issue(s)**


cmd/kyma/test/run/cmd.go
```diff
@@ -50,6 +52,7 @@ To execute all test defintions, run ` + "`kyma test run -n example-test`" + `.
 	}
 
 	cobraCmd.Flags().StringVarP(&o.Name, "name", "n", "", `Name of the new test suite. If you don't specify the value for the "-n" flag, the name of the test suite will be autogenerated.`)
+	cobraCmd.Flags().StringArrayVarP(&o.LabelExpressions, "label-expression", "l", []string{}, `label expression used to select the tests for the new test suite.`)
```
a-thaler, line 55: why not "--labels", the expression part sounds redundant.
Or "--selector" as in kubectl

k15r: --selector it is

a-thaler: besides this it looks good to me

k15r: fixed

cmd/kyma/test/run/cmd.go
```diff
@@ -144,24 +153,41 @@ func matchTestDefinitionNames(testNames []string,
...
 	octTestDefs.Spec.Count = numberOfExecutions
-	octTestDefs.Spec.Selectors.MatchNames = matchNames
+	for _, opt := range opts {
+		opt(octTestDefs)
+	}
 
 	return octTestDefs
 }
 
+type TestSuiteOption func(suite *oct.ClusterTestSuite)
```
rakesh-garimella, line 169: Why do you want to export this variable may be small T

k15r: I refactored the code a bit. now, this was moved to the test package and can be used everywhere.

shorim (approved):

bszwarc (approved):

PR #569, opened 2020-09-02
Check error message for docker-env

adamwalach:
**Description**

- ...
- ...
- ...

**Related issue(s)**

### DIFF
--- a/internal/minikube/minikube.go
+++ b/internal/minikube/minikube.go
@@ -81,6 +81,10 @@ func CheckVersion(verbose bool, timeout time.Duration) (string, error) {
 func DockerClient(verbose bool, profile string, timeout time.Duration) (*docker.Client, error) {
 	envOut, err := RunCmd(verbose, profile, timeout, "docker-env", "--shell", "bash")
 	if err != nil {
+		if strings.Contains(err.Error(), "driver does not support 'minikube docker-env'"){
+			fmt.Println("docker-env not supported, skipped")
+			return docker.NewClientWithOpts(docker.FromEnv)
+		}
 		return nil, err
 	}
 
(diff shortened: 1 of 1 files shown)


clebs (approved):
LGTM!

Only one remark, I understand that this behaviour is only for OS = Linux and driver = none combination, would it make sense to validate that the scenario is indeed what is expected before running this?

Ressetkk:
The problem occurs only when driver=none is used, because minikube's `minikube docker-env` is not compatible when this driver is in use. That fails the command prematurely.

clebs:
Yes I do understand the issue, my question was only if it would make sense to check that the OS is Linux when we get into this branch. Because as I understand, with other OSes it will not work.

But anyways, the PR is approved, so if you think that checking the OS is not necessary it is also ready to merge.

Ressetkk:
Based on the [minikube docs](https://minikube.sigs.k8s.io/docs/drivers/) the driver `none` is supported only by Linux so running it on the platform other than Linux will return minikube error:
```
$ kyma provision minikube --vm-driver=none 
- Requirements verified
  Preparing Minikube using vm-driver 'none'
? Do you want to remove the existing Minikube cluster? Type [Y/n]: y
- Minikube status verified
- Minikube config initialized
X Create Minikube instance
Error: Executing the 'minikube start --memory 8192 --cpus 4 --extra-config=apiserver.authorization-mode=RBAC --extra-config=apiserver.cors-allowed-origins='http://*' --extra-config=apiserver.enable-admission-plugins=DefaultStorageClass,LimitRanger,MutatingAdmissionWebhook,NamespaceExists,NamespaceLifecycle,ResourceQuota,ServiceAccount,ValidatingAdmissionWebhook --kubernetes-version=v1.16.3 --vm-driver none --disk-size 30g -b kubeadm' command with output '* minikube v1.12.3 on Darwin 10.15.6
X The driver 'none' is not supported on darwin
' and error message 'exit status 69' failed
```

PR #583, opened 2020-09-10
Adding more configurations for Minikube provisioning

shorim:
**Description**

- Adding configurations for `Minikube` provisioning which are needed to install `Kyma` with `Istio` 1.5 (https://github.com/kyma-project/kyma/pull/8898/files#diff-471c4a16384c64f102899d38e990a74dR212-R214)

**Related issue(s)**


a-thaler (approved):
Will that new options be work for a istio 1.4 setup? If not think about adding a switch depending on the kyma version

shorim:
@a-thaler 
These additional apiserver flags were also needed in istio 1.4 in order to enable Secret Discovery Service (SDS) (https://istio.io/v1.4/docs/setup/platform-setup/minikube/#prerequisites) `To enable SDS for your mesh, you must add extra configurations to your Kubernetes deployment. Refer to the api-server reference docs for the most up-to-date flags.`. However, we didn't use SDS so we didn't need these flags before.

Now in istio 1.5, SDS is enabled by default (https://istio.io/latest/news/releases/1.5.x/announcing-1.5/change-notes/) `Graduated SDS to stable and enabled by default. It provides identity provisioning for Istio Envoy proxies.`. So, we need to have these apiserver flags.

Issue #551, opened 2020-08-12
Initialise and deploy Function via Kyma CLI


kwiatekus:
**Description**

Support two basic commands for serverless in kyma CLI:
 - kyma function init
 - kyma function apply

Would be great to have it realised via library (serverless domain) which kyma CLI would be calling to delegate serverless oriented tasks

**User Journey**

1. As a CLI user I want to generate a function [development workspace ](https://github.com/kyma-project/community/blob/master/collaboration/sig-core/proposals/extending-kyma-cli.md#local-workspace-setup)on my local machine.
This should generate src files for the function and a config.yaml for the function resources configuration on the target runtime.

    `kyma function init --runtime nodejs ...`

2. I develop the function src. I can run the code locally (i.e via express server). I can test it
3. I can configure a function trigger in the config.yaml
4. I can deploy the function by [applying local code and config to the remote cluster](https://github.com/kyma-project/community/blob/master/collaboration/sig-core/proposals/extending-kyma-cli.md#commands-that-use-config)

   `kyma function apply ... `
 

**Reasons**

Provide CLI tools around serverless for Business Application Developers

**Attachments**
![Screen Shot 2020-09-02 at 14 12 34](https://user-images.githubusercontent.com/6783567/91979666-64913700-ed26-11ea-9f0b-179ca52a21af.png)

kwiatekus:
It should be possible to deploy function from:
 - **git repo**
- local sources

kwiatekus:
`function init` is generating folder layout locally (appropraite to selected runtime)
`function create` will create Function CR (+ GitRepo CR, if function source was defined as git repo via cli command switch)

pPrecel:
Me and @m00g3n prepared POC:

We implemented `function init` (with flags `--dir`, `--name`, `--namespace` and `--runtime`) and `function apply` (with flags `--dir` and `--dry-run`). Our solution is documented in [this PR](https://github.com/kyma-project/cli/pull/573).

My propositions of improves:
* Create a dedicated label for resources created by CLI. Because (for example) when we want to change the name of the function or other resource, CLI should remove from cluster resources with old names and CLI should find those resources using a special label.
* In my opinion whole changes proposed in PR should be located in a dedicated component or project. It's because of problems with different versions of dependencies used in the `cli` project and the `function-controller`.

a-thaler:
To pro-actively start reviewing you can have a look here already:
https://github.com/kyma-project/cli/pull/581
[...]

PR #606, opened 2020-10-05
Function cmd

pPrecel:
**Description**

- Add two new root commands: `apply` and `init` 
- Add `function` subcommand to both new commands

**Related issue(s)**

[Issue](https://github.com/kyma-project/cli/issues/551)


cmd/kyma/apply/function/opts.go
```diff
@@ -0,0 +1,48 @@
...
+type Options struct {
+	*cli.Options
+
+	OnError  OnError
+	Output   Output
+	Filename string
+	DryRun   bool
+}
+
+type OnError = string
```
clebs, line 17:
```suggestion
type OnError string
```

cmd/kyma/init/function/function.go
```diff
@@ -0,0 +1,134 @@
...
+	return nil
+}
+
+// IsValid checks if options are valid
+func (o Options) isValid() (err error) {
+	if o.URL != "" {
+		return nil
+	}
+
+	return reflectcli.NoneOf(o, gitSourceOptionNames)
```
clebs, line 95: - Do we really need to validate options we know statically at compile time using reflection? It looks like unnecessary complexity, slowdown and source of bugs to me. Fields can just be validated directly, it is more readable, flexible, safer and faster (even probably less LoC).
- Shouldn't this method be in the `cmd/kyma/init/function/opts.go` file?

cmd/kyma/init/function/function.go
```diff
@@ -0,0 +1,134 @@
...
+}
+
+func (o *Options) setDefaults() (err error) {
```
clebs, line 98: Shouldn't this method be in the cmd/kyma/init/function/opts.go file?

cmd/kyma/init/function/function.go
```diff
@@ -0,0 +1,134 @@
...
+	return
+}
+
+func setIfZero(val *string, defaultValue string) {
+	if *val == "" {
+		*val = defaultValue
+	}
+}
+
+func (o Options) source() workspace.Source {
```
clebs, line 116: Shouldn't this method be in the cmd/kyma/init/function/opts.go file?

cmd/kyma/apply/function/function.go
```diff
@@ -0,0 +1,281 @@
...
+	"github.com/kyma-incubator/hydroform/function/pkg/workspace"
+	"github.com/kyma-project/cli/internal/cli"
+	"github.com/kyma-project/cli/internal/kube"
+	"github.com/pkg/errors"
+	"github.com/spf13/cobra"
+	"gopkg.in/yaml.v2"
+	"k8s.io/apimachinery/pkg/apis/meta/v1/unstructured"
+)
+
+type Operators = map[operator.Operator][]operator.Operator
```
clebs, line 23:
```suggestion
type Operators map[operator.Operator][]operator.Operator
```

cmd/kyma/apply/function/opts.go
```diff
@@ -0,0 +1,48 @@
...
+	NothingOnError OnError = "nothing"
+	PurgeOnError   OnError = "purge"
+)
+
+var validOnError = []string{
+	NothingOnError,
+	PurgeOnError,
+}
+
+type Output = string
```
clebs, line 29:
```suggestion
type Output string
```

cmd/kyma/init/init.go
```diff
@@ -0,0 +1,19 @@
...
+	"github.com/kyma-project/cli/internal/cli"
+	"github.com/spf13/cobra"
+)
+
+//NewCmd creates a new function command
+func NewCmd(o *cli.Options) *cobra.Command {
+	cmd := &cobra.Command{
+		Use:   "init",
+		Short: "Creates local resources for your project.",
+		Long:  "Use this command to create the local files and folders for your project.",
```
clebs, line 14: Please make this description more generic, we might have other init subcommands that might not create files, folders or projects.

cmd/kyma/kyma.go
```diff
@@ -78,5 +80,9 @@ Kyma CLI allows you to install, test, and manage Kyma.
 	testCmd.AddCommand(testRunCmd, testStatusCmd, testDeleteCmd, testListCmd, testDefsCmd, testLogsCmd)
 	cmd.AddCommand(testCmd)
 
+	applyCmd := apply.NewCmd(o)
+	initCmd := initial.NewCmd(o)
+	cmd.AddCommand(applyCmd, initCmd)
```
clebs, line 85:
```suggestion
	cmd.AddCommand(
		apply.NewCmd(o),
		initial.NewCmd(o))
```

cmd/kyma/apply/function/opts.go
pPrecel: Why? This is only an alias to the string, to clarify the code

cmd/kyma/apply/function/function.go
pPrecel: Why? This is only an alias to the string, to clarify the code

cmd/kyma/apply/function/opts.go
pPrecel: Why? This is only an alias to the string, to clarify the code

cmd/kyma/apply/function/opts.go
clebs: Ah I see... So it is not your intention to create a new string type?
I assumed it was that since you have fields and constants with that type.

cmd/kyma/apply/function/function.go
clebs: Type aliases should not be used in regular code. They were introduced in Golang to do codebase refactoring as temporary in between for old/new types migration.

See: https://yourbasic.org/golang/type-alias/ 
And here detailed explanation: https://talks.golang.org/2016/refactor.article

cmd/kyma/apply/function/opts.go
clebs: Type aliases should not be used in regular code. They were introduced in Golang to do codebase refactoring as temporary in between for old/new types migration.

See: https://yourbasic.org/golang/type-alias/ 
And here detailed explanation: https://talks.golang.org/2016/refactor.article

cmd/kyma/init/function/function.go
```diff
@@ -0,0 +1,79 @@
...
+Use the flags to specify the initial configuration for your Function or to choose the location for your project.`,
+		RunE: func(cmd *cobra.Command, args []string) error {
+			return c.Run()
+		},
+	}
+
+	cmd.Flags().StringVar(&o.Name, "name", defaultName, `Function name.`)
+	cmd.Flags().StringVar(&o.Namespace, "namespace", defaultNamespace, `Namespace to which you want to apply your Function.`)
+	cmd.Flags().StringVarP(&o.Dir, "dir", "d", "", `Full path to the directory where you want to save the project.`)
+	cmd.Flags().StringVarP(&o.Runtime, "runtime", "r", defaultRuntime, `Flag used to define the environment for running you Function. Use one of:
```
kazydek, line 41:
```suggestion
	cmd.Flags().StringVarP(&o.Runtime, "runtime", "r", defaultRuntime, `Flag used to define the environment for running you Function. Use one of these values:
```

cmd/kyma/init/function/function.go
```diff
@@ -0,0 +1,79 @@
...
+	- nodejs12
+	- nodejs10
+	- python38`)
+
+	// git function options
+	cmd.Flags().StringVar(&o.URL, "url", "", `Git repository URL`)
+	cmd.Flags().StringVar(&o.RepositoryName, "repository-name", o.Name, `The name of the git repository to be created`)
```
kazydek, line 48:
```suggestion
	cmd.Flags().StringVar(&o.RepositoryName, "repository-name", o.Name, `The name of the Git repository to be created`)
```

cmd/kyma/init/function/function.go
```diff
@@ -0,0 +1,79 @@
...
+	cmd.Flags().StringVar(&o.Reference, "reference", defaultReference, `Commit hash or branch name`)
+	cmd.Flags().StringVar(&o.BaseDir, "base-dir", defaultBaseDir, `A directory in repository containing function sources`)
```
kazydek, line 50:
```suggestion
	cmd.Flags().StringVar(&o.BaseDir, "base-dir", defaultBaseDir, `A directory in the repository containing the Function's sources`)
```

cmd/kyma/apply/function/function.go
```diff
@@ -0,0 +1,230 @@
...
+		Long: `Use this command to apply the local sources of your Function's code and dependencies to the Kyma cluster. 
+Use the flags to specify the desired location for the source files or run the command to validate and print the output resources.`,
+		RunE: func(cmd *cobra.Command, args []string) error {
+			return c.Run()
+		},
+	}
+
+	cmd.Flags().VarP(&o.Filename, "filename", "f", `Full path to the config file.`)
+	cmd.Flags().BoolVar(&o.DryRun, "dry-run", false, `Validated list of objects to be created from sources.`)
+	cmd.Flags().Var(&o.OnError, "onerror", `Flag used to define reaction to the error. Use one of:
```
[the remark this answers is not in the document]
pPrecel: for me `options` should look better

kazydek (dismissed):
🎉

cmd/kyma/apply/function/opts.go
```diff
@@ -0,0 +1,117 @@
...
+	if v == "" {
+		return nil
+	} else if g.isOneOf(v) {
+		g.value = v
+		return nil
+	}
+	return fmt.Errorf("specified value: %s is not supported", v)
+}
+
+func (g value) isOneOf(item string) bool {
```
clebs, line 84: Maybe we can rename this to something more descriptive, like `isAllowed`.

cmd/kyma/apply/function/opts.go
```diff
@@ -0,0 +1,117 @@
...
+		}
+	}
+	return false
+}
+
+func (g value) Type() string {
+	return "value"
+}
+
+type filename string
```
clebs, line 97: What is the benefit of having this type? It looks like a regular string to me.

cmd/kyma/apply/function/opts.go
```diff
@@ -0,0 +1,117 @@
...
+
+func (f filename) String() string {
+	return string(f)
+}
+
+func (f filename) Type() string {
+	return "filename"
+}
+
+func (f *filename) Set(v string) error {
```
clebs, line 112: No need to return anything if always returning `nil`.

cmd/kyma/apply/function/opts.go
pPrecel: Now nothing. In the previous version, this type had a defaulting constructor and other methods. But I had a problem with keeping $PWD as the default value, because of generated documentation. I will change it ;)

cmd/kyma/apply/function/opts.go
pPrecel: This method is needed by [this interface](https://github.com/spf13/pflag/blob/master/flag.go#L187-L191)

cmd/kyma/apply/function/function.go
```diff
@@ -0,0 +1,230 @@
...
+		gitOperator := operator.NewGenericOperator(client.Resource(operator.GVRGitRepository).Namespace(configuration.Namespace), gitRepository)
+		operators[gitOperator] = nil
+	}
+
+	mgr := manager.NewManager(operators)
+	options := manager.Options{
+		Callbacks:          callbacks(c),
+		OnError:            chooseOnError(c.opts.OnError),
+		DryRun:             c.opts.DryRun,
+		SetOwnerReferences: true, // add one more flag to set it to false or add: c.opts.Output != YAMLOutput && c.opts.Output != JSONOutput
```
tgorgol, line 107: Was this supposed to be added in this PR or will there be another issue for it?

tgorgol (commented):
Good job, I left few small suggestions to the code.

I found one problem with `kyma apply function` command, however, it's probably related to the code from hydroform library. When I have this config.yaml:
```
name: first-function
namespace: tg
runtime: nodejs12
source:
    sourceType: git
    url: https://test/test.git
    repository: first-function
    reference: master
    baseDir: /
```
and I use `kyma apply function -o yaml` I'm getting:
```
---
apiVersion: serverless.kyma-project.io/v1alpha1
kind: Function
metadata:
  name: first-function
  namespace: tg
spec:
  resources:
    limits: {}
    requests: {}
  runtime: nodejs12
  source: first-function

---
apiVersion: serverless.kyma-project.io/v1alpha1
kind: GitRepository
metadata:
  name: first-function
  namespace: tg
spec:
  url: https://test/test.git
``` 
as you can see the function doesn't have the `sourceType: git` field, so the function controller treats it as an inline function.

Lastly, one thing I think we should discuss is the management of created resources. I can see that triggers are nicely managed, when I modify the list of triggers by removing one it gets removed from the cluster etc. But it doesn't work that way for GitRepositories or Functions, when you have a config with 5 triggers and change the name of the function, the old function will be left untouched and new one is created with new triggers. I think this could be improved in the next iteration but I think we should have this kind of control over created resources.

cmd/kyma/apply/function/function.go
pPrecel: After merging this PR I will log an issue for this

cmd/kyma/init/function/function.go
```diff
@@ -0,0 +1,79 @@
...
+import (
+	"github.com/kyma-incubator/hydroform/function/pkg/workspace"
+	"github.com/kyma-project/cli/internal/cli"
+	"github.com/spf13/cobra"
+)
+
+const (
+	defaultRuntime   = "nodejs12"
+	defaultNamespace = "default"
+	defaultName      = "first-function"
```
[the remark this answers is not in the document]
pPrecel: After merging this PR I will log an issue for this

cmd/kyma/init/function/function.go
```diff
@@ -0,0 +1,79 @@
...
+		Long: `Use this command to create the local workspace with the default structure of your Function's code and dependencies. Update this configuration to your references and apply it to a Kyma cluster. 
+Use the flags to specify the initial configuration for your Function or to choose the location for your project.`,
+		RunE: func(cmd *cobra.Command, args []string) error {
+			return c.Run()
+		},
+	}
+
+	cmd.Flags().StringVar(&o.Name, "name", defaultName, `Function name.`)
+	cmd.Flags().StringVar(&o.Namespace, "namespace", defaultNamespace, `Namespace to which you want to apply your Function.`)
+	cmd.Flags().StringVarP(&o.Dir, "dir", "d", "", `Full path to the directory where you want to save the project.`)
```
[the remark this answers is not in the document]
pPrecel: After merging this PR I will log an issue for this

pPrecel:
> Good job, I left few small suggestions to the code.
> 
> I found one problem with `kyma apply function` command, however, it's probably related to the code from hydroform library. When I have this config.yaml:
> 
> ```
> name: first-function
> namespace: tg
> runtime: nodejs12
> source:
>     sourceType: git
>     url: https://test/test.git
>     repository: first-function
>     reference: master
>     baseDir: /
> ```
> 
> and I use `kyma apply function -o yaml` I'm getting:
> 
> ```
> ---
> apiVersion: serverless.kyma-project.io/v1alpha1
> kind: Function
> metadata:
>   name: first-function
>   namespace: tg
> spec:
>   resources:
>     limits: {}
>     requests: {}
>   runtime: nodejs12
>   source: first-function
> 
> ---
> apiVersion: serverless.kyma-project.io/v1alpha1
> kind: GitRepository
> metadata:
>   name: first-function
>   namespace: tg
> spec:
>   url: https://test/test.git
> ```
> 
> as you can see the function doesn't have the `sourceType: git` field, so the function controller treats it as an inline function.
> 
> Lastly, one thing I think we should discuss is the management of created resources. I can see that triggers are nicely managed, when I modify the list of triggers by removing one it gets removed from the cluster etc. But it doesn't work that way for GitRepositories or Functions, when you have a config with 5 triggers and change the name of the function, the old function will be left untouched and new one is created with new triggers. I think this could be improved in the next iteration but I think we should have this kind of control over created resources.

Good point, thanks! @m00g3n fixed part with the `sourceType` of functions and about better control of Functions and GitRepositories, how you said, we can log an issue.

tgorgol (approved):

cmd/kyma/apply/apply_test.go
```diff
@@ -0,0 +1,20 @@
...
+func TestSubcommands(t *testing.T) {
+	c := NewCmd(&cli.Options{})
+	c.SetOutput(ioutil.Discard) // not interested in the command's output
+
+	// test default flag values
+	require.NoError(t, c.Execute(), "Command execution must not fail")
+
+	sub := c.Commands()
+
+	require.Equal(t, 2, len(sub), "Number of create subcommands not as expected")
```
kazydek, line 19:
```suggestion
	require.Equal(t, 2, len(sub), "Number of created subcommands not as expected")
```

cmd/kyma/apply/function/function.go
```diff
@@ -0,0 +1,238 @@
...
+	}
+
+	// Load project configuration
+	var configuration workspace.Cfg
+	if err := yaml.NewDecoder(file).Decode(&configuration); err != nil {
+		return errors.Wrap(err, "Could not decode the configuration file")
+	}
+
+	if c.K8s, err = kube.NewFromConfig("", c.KubeconfigPath); err != nil {
+		return errors.Wrap(err, "Could not initialize the Kubernetes client. Make sure your kubeconfig is valid")
```
kazydek, line 73: Comma is also not required here?

cmd/kyma/init/function/function.go
```diff
@@ -0,0 +1,79 @@
...
+	cmd.Flags().StringVarP(&o.Runtime, "runtime", "r", defaultRuntime, `Flag used to define the environment for running you Function. Use one of these options:
```
kazydek, line 41:
```suggestion
	cmd.Flags().StringVarP(&o.Runtime, "runtime", "r", defaultRuntime, `Flag used to define the environment for running your Function. Use one of these options:
```

cmd/kyma/init/init_test.go
```diff
@@ -0,0 +1,20 @@
...
+func TestSubcommands(t *testing.T) {
+	c := NewCmd(&cli.Options{})
+	c.SetOutput(ioutil.Discard) // not interested in the command's output
+
+	// test default flag values
+	require.NoError(t, c.Execute(), "Command execution must not fail")
+
+	sub := c.Commands()
+
+	require.Equal(t, 2, len(sub), "Number of create subcommands not as expected")
```
kazydek, line 19:
```suggestion
	require.Equal(t, 2, len(sub), "Number of created subcommands not as expected")
```

cmd/kyma/apply/function/function.go
pPrecel: Yes, because the `Wrap` method will merge these two strings (message and an error message) with `:` separator. So the final message will look like `... valid: ...`

tgorgol (approved):

kazydek (approved):

clebs (approved):

PR #609, opened 2020-10-08
treat errors as failure in junit output

k15r:
**Description**

- put apiserver related issues in the junit output
  previously errors encountered during log fetching resulted in an empty junit output. This pr puts a message in the junit output and tries to fill as much of the output report as possible

**Related issue(s)**


clebs (approved):
Great cleanup of the unnecessary `error` return values.
Thanks!

PR #618, opened 2020-10-15
Add sync function command

kfurgol:
**Description**

- Added sync function command

**Related issue(s)**
[#9403](https://github.com/kyma-project/kyma/issues/9403)


cmd/kyma/kyma.go
```diff
@@ -68,6 +69,7 @@ Kyma CLI allows you to install, test, and manage Kyma.
 		console.NewCmd(console.NewOptions(o)),
 		upgrade.NewCmd(upgrade.NewOptions(o)),
 		create.NewCmd(o),
+		sync.NewCmd(o),
```
clebs, line 72: The `sync` command was already added in line 88 together with the other function commands, no need to add it a second time.

cmd/kyma/sync/function/function.go
```diff
@@ -0,0 +1,101 @@
...
+
+	if _, err := os.Stat(c.opts.OutputPath); os.IsNotExist(err) {
+		err = os.MkdirAll(c.opts.OutputPath, 0700)
+		if err != nil {
+			return err
+		}
+	}
+
+
+	crdConfig, err := clientcmd.BuildConfigFromFlags("", c.opts.Kubeconfig)
```
clebs, line 67: What happens here when no kubeconfig is provided? Is then the `KUBECONFIG` env variable used? the default config location (`~/.kube/config`)?

Please use the kubeconfig and k8s client utilities built into the CLI, they have been tested and improved over long time:
https://github.com/kyma-project/cli/blob/bad193c82b5fa538f69335789e64bb536beb2200/internal/kube/client.go#L41

cmd/kyma/sync/function/function.go
```diff
@@ -0,0 +1,101 @@
...
+		Short: "Saves locally resources for requested Function.",
+		Long: `Use this command to create the local workspace with the default structure of your Function's code and dependencies. Update this configuration to your references and apply it to a Kyma cluster. 
+Use the flags to specify the initial configuration for your Function or to choose the location for your project.`, //todo
+		RunE: func(cmd *cobra.Command, args []string) error {
+			return c.Run()
+		},
+	}
+
+	cmd.Flags().StringVar(&o.Name, "name", "", `Function name.`)
+	cmd.Flags().StringVarP(&o.Namespace, "namespace","n", defaultNamespace, `Namespace from which you want to sync Function.`)
```
clebs, line 44: Users might have a different namespace than "default" as their defult namespace in the kubeconfig. This flag should use that if empty.
See example here:
https://github.com/kyma-project/cli/blob/bad193c82b5fa538f69335789e64bb536beb2200/cmd/kyma/create/system/system.go#L74-L76

cmd/kyma/sync/function/function.go
```diff
@@ -0,0 +1,101 @@
...
+	cmd.Flags().StringVarP(&o.OutputPath, "output", "o", "", `Full path to the directory where you want to save the project.`)
+	cmd.Flags().StringVarP(&o.Kubeconfig, "kubeconfig", "k", "", `Full path to the Kubeconfig file.`)
```
clebs, line 46: You do not need to declare this flag, it is already declared as global flag here (all children commands have access to it):
https://github.com/kyma-project/cli/blob/ad16f7d87ea437386c11bb7df08c63a85b5e6e5a/cmd/kyma/kyma.go#L50

cmd/kyma/sync/function/opts.go
```diff
@@ -0,0 +1,40 @@
...
+)
+
+//Options defines available options for the command
+type Options struct {
+	*cli.Options
+
+	Name           string
+	Namespace      string
+	OutputPath     string
+	Kubeconfig     string
```
clebs, line 15: Please remove this.

cmd/kyma/sync/function/opts.go
```diff
@@ -0,0 +1,40 @@
...
+		if err != nil {
+			return err
+		}
+	}
+
+	setIfZero(&o.Namespace, defaultNamespace)
+	return
+}
+
+func setIfZero(val *string, defaultValue string) {
```
clebs, line 36: No need for this function, see previous comments on namespace defaults.

clebs (changes requested):

go.mod
```diff
@@ -3,7 +3,7 @@ module github.com/kyma-project/cli
 go 1.14
 
 replace (
-	github.com/Sirupsen/logrus => github.com/sirupsen/logrus v1.6.0
+	github.com/kyma-incubator/hydroform/function => github.com/m00g3n/hydroform/function v0.0.0-20201021210504-5d757fc21aa6
```
m00g3n, line 6: once `sync` functionality will be in `hydroform` we will change this

go.mod
```diff
@@ -3,7 +3,7 @@ module github.com/kyma-project/cli
 go 1.14
 
 replace (
-	github.com/Sirupsen/logrus => github.com/sirupsen/logrus v1.6.0
```
pPrecel, line 6: Replace from the `Sirupsen/logrus` with the `sirupsen/logrus` is not necessary?

cmd/kyma/sync/sync.go
```diff
@@ -0,0 +1,19 @@
...
+	"github.com/kyma-project/cli/internal/cli"
+	"github.com/spf13/cobra"
+)
+
+//NewCmd creates a new function command
+func NewCmd(o *cli.Options) *cobra.Command {
+	cmd := &cobra.Command{
+		Use:   "sync",
+		Short: "Synchronises the function files from the cluster.",
+		Long:  "Use this command to synchronise the function files from the cluster.",
```
pPrecel, line 14: I think that `sync` command may be used by the other functionalities in the future. Maybe the description should be more generic?

cmd/kyma/sync/function/opts.go
```diff
@@ -0,0 +1,50 @@
...
+		o.OutputPath, err = os.Getwd()
+		if err != nil {
+			return err
+		}
+	}
+
+	setIfZero(&o.Namespace, defaultNamespace, "default")
+
+	if o.Name == "" {
+		return fmt.Errorf("flag 'name' is required")
```
pPrecel, line 37: I think you can use simple `MarkFlagRequired` instead of this `if`
https://pkg.go.dev/github.com/spf13/cobra#Command.MarkFlagRequired

pPrecel (approved):
/meow LGTM

go.mod
m00g3n: There is no usage of "old" `Sirupsen` in module: 
https://github.com/kyma-project/cli/blob/0099ed464748d9a71ab722ac7e9ec67446b02de8/go.sum#L1015-L1021

cmd/kyma/sync/sync.go
kazydek, line 14:
```suggestion
		Long:  "Use this command to download the given resource's code and dependencies from the cluster to create or update these resources in your local workspace Currently, you can only use it for Functions.",
```

cmd/kyma/sync/function/opts.go
m00g3n: Good point, however there is a catch with `MarkFlagRequired`function, it returns `error`.
Since this method needs to be invoked in [NewCmd](https://github.com/kyma-project/cli/pull/618/files#diff-68b06847f40033401830cc910c8ccea68ddbf92fe2991aa7044195b3540bc16cR21) instead in [Run](https://github.com/kyma-project/cli/pull/618/files#diff-68b06847f40033401830cc910c8ccea68ddbf92fe2991aa7044195b3540bc16cR43) we would have to: 
- _either_ `panic` in case of an error 
- _or_ change the signature of the `NewCmd` method to return an `error` and this leads to the refactor of all the code that starts in [main](https://github.com/kyma-project/cli/blob/master/cmd/main.go#L17) function of the CLI.

I believe the 2nd option is better but it is out of the scope of this PR, let's address it in separate issue.

cmd/kyma/sync/function/function.go
```diff
@@ -0,0 +1,82 @@
...
+		Short: "Synchronizes the local resources for your Function.",
+		Long: `Use this command to download the Function's code and dependencies from the cluster to create or update these resources in your local workspace.
+Use the flags to specify the name of your Function, the Namespace, or the location for your project.`,
+		RunE: func(cmd *cobra.Command, args []string) error {
+			return c.Run()
+		},
+	}
+
+	cmd.Flags().StringVar(&o.Name, "name", "", `Function name.`)
+	cmd.Flags().StringVarP(&o.Namespace, "namespace", "n", "", `Namespace from which you want to sync Function.`)
```
kazydek, line 37:
```suggestion
	cmd.Flags().StringVarP(&o.Namespace, "namespace", "n", "", `Namespace from which you want to sync the Function.`)
```

cmd/kyma/sync/sync.go
```diff
@@ -0,0 +1,19 @@
...
+	"github.com/kyma-project/cli/internal/cli"
+	"github.com/spf13/cobra"
+)
+
+//NewCmd creates a new function command
+func NewCmd(o *cli.Options) *cobra.Command {
+	cmd := &cobra.Command{
+		Use:   "sync",
+		Short: "Synchronizes the local resources for your Function.",
+		Long:  "Use this command to download the given resource's code and dependencies from the cluster to create or update these resources in your local workspace Currently, you can only use it for Functions.",
```
kazydek, line 14:
```suggestion
		Long:  "Use this command to download the given resource's code and dependencies from the cluster to create or update these resources in your local workspace. Currently, you can only use it for Functions.",
```

kazydek (dismissed):

Tika6580:
> **Description**
> 
> Changes proposed in this pull request:
> 
> * Added sync function command
> 
> **Related issue(s)**
> [#9403](https://github.com/kyma-project/kyma/issues/9403)

cmd/kyma/sync/function/opts.go
```diff
@@ -0,0 +1,52 @@
...
+
+func (o *Options) setDefaults(defaultNamespace string) (err error) {
+	if o.OutputPath == "" {
+		o.OutputPath, err = os.Getwd()
+		if err != nil {
+			return err
+		}
+	}
+
+	setIfZero(&o.Namespace, defaultNamespace, "default")
```
clebs, line 34: This is not necessary, `c.K8s.DefaultNamespace()` already returns `default` if the user has no default namespace configured in the Kubeconfig.
Check the documentation please:
https://github.com/clebs/kyma-cli/blob/bad193c82b5fa538f69335789e64bb536beb2200/internal/kube/kube.go#L32-L35

cmd/kyma/sync/function/opts.go
```diff
@@ -0,0 +1,52 @@
...
+
+	if o.Name == "" {
+		return fmt.Errorf("flag 'name' is required")
+	}
+
+	return
+}
+
+func setIfZero(val *string, defaultValue string, defaultValues ...string) bool {
```
clebs, line 43: According to the comment on line 34, this function can be completely deleted.

clebs (approved):

PR #633, opened 2020-10-28
Synchronise command adjustment

kfurgol:
**Description**

- Changed --output flag to --dir flag to keep consistent with `init` command
- Name of the function is now passed as the only argument instead of being passed as a flag

**Related issue(s)**

### DIFF
--- a/cmd/kyma/sync/function/function.go
+++ b/cmd/kyma/sync/function/function.go
@@ -30,20 +30,25 @@ func NewCmd(o *Options) *cobra.Command {
 		Long: `Use this command to download the Function's code and dependencies from the cluster to create or update these resources in your local workspace.
 Use the flags to specify the name of your Function, the Namespace, or the location for your project.`,
 		RunE: func(cmd *cobra.Command, args []string) error {
-			return c.Run()
+			return c.Run(args[0])
+		},
+		Args: func(cmd *cobra.Command, args []string) error {
+			if len(args) != 1 {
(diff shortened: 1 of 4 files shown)


cmd/kyma/sync/function/function.go
```diff
@@ -30,20 +30,24 @@ func NewCmd(o *Options) *cobra.Command {
...
 Use the flags to specify the name of your Function, the Namespace, or the location for your project.`,
 		RunE: func(cmd *cobra.Command, args []string) error {
-			return c.Run()
+			return c.Run(args[0])
+		},
+		Args: func(cmd *cobra.Command, args []string) error {
+			if len(args) != 1 {
+				return errors.New("missing name of the function")
+			}
+			return nil
```
clebs, line 39:
```suggestion
		Args: cobra.ExactArgs(1)
```

clebs: Only suggested if error message is not so relevant, otherwise please ignore.

clebs (approved):

cmd/kyma/sync/function/function.go
kfurgol: I didn't use `ExactArgs` on purpose to have the custom error msg, but I will change it if you insist on it

clebs: As I said, only suggested if error message was is not relevant. If it is, leave it as it is.

kazydek (approved):

PR #643, opened 2020-11-12
Speed up test execution 

clebs:
**Description**

- By setting up tests to run in parallel, execution time went down by ~20%
- On Prow (more CPU cores), execution went down ~60%.

### DIFF
--- a/cmd/kyma/apply/apply_test.go
+++ b/cmd/kyma/apply/apply_test.go
@@ -1,13 +1,15 @@
 package apply
 
 import (
-	"github.com/kyma-project/cli/internal/cli"
-	"github.com/stretchr/testify/require"
 	"io/ioutil"
 	"testing"
+
+	"github.com/kyma-project/cli/internal/cli"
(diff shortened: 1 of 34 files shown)


skhalash:
Great, it's a low-hanging fruit indeed! I assume those are all unit tests, right? So they should be independent in terms of a shared state, shouldn't they?

skhalash:
One could possibly go further and parallelise table-driven test-cases, but there is a trick: https://gist.github.com/posener/92a55c4cd441fc5e5e85f27bca008721

suleymanakbas91 (approved):

clebs:
Definitely @skhalash, table driven tests could be executed by calling `t.RunParallel()` instead of `t.Run()`, I had thought about it.

I decided not to add it on this first PR because in most cases table tests share data (e.g. a service you initialise before the loop).
But of course, if we invest a bit more time in analysing those table driven tests and refactoring for parallelism we could be even faster!

PR #644, opened 2020-11-13
Fix: Git function creation fails due to missing GitRepository CR

pPrecel:
**Description**

- Change the order of operators used for the `kyma apply function` command

**Related issue(s)**

https://github.com/kyma-incubator/hydroform/issues/135

### DIFF
--- a/cmd/kyma/apply/function/function.go
+++ b/cmd/kyma/apply/function/function.go
@@ -41,6 +41,8 @@ Use the flags to specify the desired location for the source files or run the co
 
 	cmd.Flags().StringVarP(&o.Filename, "filename", "f", "", `Full path to the config file.`)
 	cmd.Flags().BoolVar(&o.DryRun, "dry-run", false, `Validated list of objects to be created from sources.`)
+	cmd.Flags().DurationVarP(&o.Timeout, "timeout", "t", 0, `Maximum time during which the local resources are being applied, where "0" means "infinite". Valid time units are "ns", "us" (or "µs"), "ms", "s", "m", "h".`)
+	cmd.Flags().BoolVarP(&o.Watch, "watch", "w", false, `Watches resources applied to the cluster to be sure that everything is applying in the correct order.`)
 	cmd.Flags().Var(&o.OnError, "onerror", `Flag used to define the Kyma CLI's reaction to an error when applying resources to the cluster. Use one of these options: 
 - nothing
 - purge`)
@@ -74,6 +76,8 @@ func (c *command) Run() error {
 	}
 	client := c.K8s.Dynamic()
 
+	mgr := manager.NewManager()
+
 	function, err := resources.NewFunction(configuration)
 	if err != nil {
 		return err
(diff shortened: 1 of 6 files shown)


cmd/kyma/apply/function/function.go
```diff
@@ -84,19 +84,21 @@ func (c *command) Run() error {
...
 		if err != nil {
 			return errors.Wrap(err, "Unable to read the Git repository from the provided configuration")
 		}
-		gitOperator := operator.NewGenericOperator(client.Resource(operator.GVRGitRepository).Namespace(configuration.Namespace), gitRepository)
-		operators[gitOperator] = nil
+		gitOperator = operator.NewGenericOperator(client.Resource(operator.GVRGitRepository).Namespace(configuration.Namespace), gitRepository)
+	}
+
+	operators := map[operator.Operator][]operator.Operator{
+		gitOperator: {},
```
clebs, line 98: What happens when you have a different type of source? You will have a `nil` key with an empty slice.

I would refactor this code so that the git operator gets added to the map only if the source type is git.

So first create the map with the generic operator that is added in all cases and later add the git one only if necessary.

pPrecel: Exactly. When the type of source is git, then there is an initialized operator with no child operators (operator: {}), but in a case with inline function, there is just nil with no child operators (nil: {}. `nil: nil` gives the same result). The manager works fine in all those cases ( [click for more](https://github.com/kyma-incubator/hydroform/blob/master/function/pkg/manager/manager_test.go#L239) ). Of course, I can change it and append the operator to the map when the type of source is set as `git`. But I think this is only a visual change because this will not speed up the code ;)

clebs (approved):

cmd/kyma/apply/function/function.go
```diff
@@ -99,12 +95,17 @@ func (c *command) Run() error {
...
+		operator.NewTriggersOperator(client.Resource(operator.GVKTriggers).Namespace(configuration.Namespace), triggers...),
+	}
+
 	mgr := manager.NewManager(operators)
 	options := manager.Options{
 		Callbacks:          callbacks(c),
 		OnError:            chooseOnError(c.opts.OnError),
 		DryRun:             c.opts.DryRun,
 		SetOwnerReferences: true,
+		WaitForApplying:    true,
```
clebs, line 108: Shouldn't this be configurable per flag?

I would also call it `watch` to be consistent with other CLI commands and tools like `kubectl`.
See:
https://github.com/kyma-project/cli/blob/33e45cb9ff60fc2b80f502873618628ea2feb425/cmd/kyma/test/run/cmd.go#L60

pPrecel: There will be more changes so I think that this PR is not ready to review yet. Of course, I will apply your suggestion but in the next commits, I will add more changes ;)

docs/gen-docs/kyma_apply_function.md
```diff
@@ -16,16 +16,18 @@ kyma apply function [flags]
...
+      --onerror value      Flag used to define the Kyma CLI's reaction to an error when applying resources to the cluster. Use one of these options: 
+                           - nothing
+                           - purge (default nothing)
+  -o, --output value       Flag used to define the command output format. Use one of these options:
+                           - text
+                           - json
+                           - yaml
+                           - none (default text)
+  -t, --timeout duration   Maximum time during which the local resources are being applied, where "0" means "infinite". Valid time units are "ns", "us" (or "µs"), "ms", "s", "m", "h".
+  -w, --watch              Watches resources applied to the cluster to be sure that everything is applying in the correct order.
```
kazydek, line 30:
```suggestion
  -w, --watch              Flag used to watch resources applied to the cluster to make sure that everything is applied in the correct order.
```

kazydek (dismissed):

cmd/kyma/apply/function/function_test.go
```diff
@@ -18,25 +19,33 @@ func TestFunctionFlags(t *testing.T) {
...
+		"--timeout", "15s",
+		"--watch", "true",
 	})
 	require.NoError(t, err, "Parsing flags should not return an error")
 	require.Equal(t, "/fakepath/config.yaml", o.Filename, "The parsed value for the --filename flag not as expected.")
 	require.Equal(t, true, o.DryRun, "The parsed value for the --dry-run flag not as expected.")
 	require.Equal(t, "purge", o.OnError.String(), "The parsed value for the --onerror flag not as expected.")
 	require.Equal(t, "json", o.Output.String(), "The parsed value for the --output flag not as expected.")
+	require.Equal(t, time.Duration(15)*time.Second, o.Timeout, "Default value for the --timeout flag not as expected.")
+	require.Equal(t, true, o.Watch, "Default value for the --watch flag not as expected.")
```
clebs, line 40: The message here is misleading, these are not default values.

cmd/kyma/apply/function/function_test.go
```diff
@@ -18,25 +19,33 @@ func TestFunctionFlags(t *testing.T) {
...
 
 	err = c.ParseFlags([]string{
 		"-f", "/config.yaml",
 		"-o", "yaml",
+		"-t", "5s",
```
clebs, line 45: Shouldn't this also test the shorthand flag for `watch` (`-w`)?

kazydek (approved):

clebs (approved):

PR #646, opened 2020-11-17
Append the current commit hash in case of a stable version

### LINKED ISSUE #594: Kyma CLI reports version as stable and it's hard to find out which stable version
**Description**
When user execute `kyma version`, he got such response:
```
kyma version
Kyma CLI version: stable
```

Kyma CLI version: stable says nothing about exact stable version. There is no way to determine version.

**Expected result**
Kyma CLI should raport:
```
kyma version
Kyma CLI version: stable-{hash_id}
```

skhalash:
**Description**

- append the current commit hash in case of a stable version

**Related issue(s)**
Resolves #594

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -8,6 +8,10 @@ ifndef VERSION
 	VERSION = ${shell git describe --tags --always}
 endif
 
+ifeq ($(VERSION),stable)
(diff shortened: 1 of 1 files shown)


clebs (approved):
Very elegant solution :)

PR #647, opened 2020-11-17
Strip ANSI escape sequences while fetching logs from containers

Ressetkk:
**Description**
I have noticed that the logs from the testsuite containers are fetched with the ANSI escape sequences from the console log, which is expected, but XML generator cannot encode ANSI sequences to utf8. This breaks TestGrid and Spyglass JUnit parsing.

I have checked the function on the part of the log https://storage.googleapis.com/kyma-prow-logs/logs/post-master-kyma-gke-integration/1328263533271453697/artifacts/junit_Kyma_octopus-test-suite-1605518975.xml. It should strip the bad stuff properly. If you need unit test for that function then I can sort something out.

- Add `strip()` function that strips the container log string from the ANSI characters based on the regex.

**Related issue(s)**

https://github.com/kyma-project/test-infra/issues/2900

### DIFF
--- a/internal/logs/fetcher.go
+++ b/internal/logs/fetcher.go
@@ -73,6 +73,7 @@ func (f *FetcherForTestingPods) Logs(result oct.TestResult) (string, error) {
 	return strip(logs.String()), nil
 }
 
+// strip removes ANSI escape sequences from the input string
 func strip(s string) string {
 	re := regexp.MustCompile(ansi)
 	return re.ReplaceAllString(s, "")
--- a/internal/logs/fetcher.go
+++ b/internal/logs/fetcher.go
@@ -70,11 +70,10 @@ func (f *FetcherForTestingPods) Logs(result oct.TestResult) (string, error) {
 		}
 	}
 
-	return strip(logs.String()), nil
+	return stripANSI(logs.String()), nil
 }
 
-// strip removes ANSI escape sequences from the input string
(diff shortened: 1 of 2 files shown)


internal/logs/fetcher.go
```diff
@@ -67,5 +70,11 @@ func (f *FetcherForTestingPods) Logs(result oct.TestResult) (string, error) {
 		}
 	}
 
-	return logs.String(), nil
+	return strip(logs.String()), nil
+}
+
+// strip removes ANSI escape sequences from the input string
+func strip(s string) string {
```
clebs, line 77: Why not call the function `stripANSI` and drop the comment?

Also a unit test for the function would be great.

Ressetkk: Yep I have updated the code based on your review

internal/logs/fetcher_test.go
```diff
@@ -63,6 +62,32 @@ func Test(t *testing.T) {
 	}
 }
 
+func Test_stripANSI(t *testing.T) {
+	testCases := []struct {
```
clebs, line 66: Could you please add `t.Parallel()` as the first line of the test so that it runs concurrently to others?
This reduces test execution time significantly.

Ressetkk: Done!

clebs (approved):

Issue #431, opened 2020-04-14
Consistent flags naming


a-thaler:
**Description**
Please introduce a proper naming strategy for flags of a CLI command and apply it. At the moment it is inconsistent, see for example `kyma provision minikube --disk-size --vm-driver --hypervVirtualSwitch` (the flag for the switch name is using CamelCase while the others do not)

**Reasons**

Developer experience
**Attachments**

a-thaler:
I vote to use my-resources notation instead of camelCase as it is used in minikube and kubectl

PR #650, opened 2020-11-17
Consistent flag naming

shorim:
**Description**

- avoid using camelCase naming for flags in order to have consistency in the naming style
- `noWait`, `tlsCert` , `tlsKey` , `FallbackLevel` and `hypervVirtualSwitch` flags are changed to `no-wait`, `tls-cert`, `tls-key`, `fallback-level` and `hyperv-virtual-switch` respectively.

**Related issue(s)**

#431

### DIFF
--- a/cmd/kyma/install/cmd.go
+++ b/cmd/kyma/install/cmd.go
@@ -79,10 +79,10 @@ The standard installation uses the minimal configuration. The system performs th
 		Aliases: []string{"i"},
 	}
 
-	cobraCmd.Flags().BoolVarP(&o.NoWait, "noWait", "n", false, "Flag that determines if the command should wait for Kyma installation to complete.")
+	cobraCmd.Flags().BoolVarP(&o.NoWait, "no-wait", "n", false, "Determines if the command should wait for Kyma installation to complete.")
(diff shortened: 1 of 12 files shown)


clebs (approved):
LGTM!

As follow up, I would ask you to check if there are any changes needed in the kyma docs, specially under: https://kyma-project.io/docs/#installation-installation

suleymanakbas91 (commented):
In addition to docs, we need to adapt test-infra scripts as well. Otherwise, as soon as we create a new stable binary, they will start failing. Also, this should be documented in the release changelog as it is breaking.

shorim:
@clebs I have checked the kyma docs and only found that only `tlsCert` and `tlsKey` were used there. Here is the PR: https://github.com/kyma-project/kyma/pull/9988

shorim:
@suleymanakbas91 I have searched the test-infra repo for the usage of any of the changed 5 flags. Only found `tlsCert` and `tlsKey` used in multiple scripts. Here is the PR: https://github.com/kyma-project/test-infra/pull/2966

alexandra-simeonova (approved):

PR #687, opened 2020-12-30
Clean logs for alpha deploy cmd

tobiscr:
**Description**

Add step-based output for alpha deploy/undeploy command. Add verbose mode.

**Related issue(s)**

### DIFF
--- a/cmd/kyma/alpha/deploy/cmd.go
+++ b/cmd/kyma/alpha/deploy/cmd.go
@@ -95,12 +95,13 @@ func (cmd *command) Run() error {
 		Log:                           cmd.getLogFunc(),
 	}
 
+	var updateCh chan deployment.ProcessUpdate
 	if cmd.opts.Verbose {
 		defer log.Println("Kyma deployed!")
 	} else {
 		// Start rendering async CLI UI
 		asyncUI := asyncui.AsyncUI{StepFactory: &cmd.Factory}
-		updateCh := asyncUI.Start()
+		updateCh = asyncUI.Start()
 		defer asyncUI.Stop()
 	}
 
--- a/cmd/kyma/alpha/deploy/cmd.go
+++ b/cmd/kyma/alpha/deploy/cmd.go
@@ -101,7 +101,10 @@ func (cmd *command) Run() error {
 	} else {
 		// Start rendering async CLI UI
 		asyncUI := asyncui.AsyncUI{StepFactory: &cmd.Factory}
-		updateCh = asyncUI.Start()
+		updateCh, err = asyncUI.Start()
+		if err != nil {
+			return err
+		}
 		defer asyncUI.Stop() // stop receiving update-events and wait until UI rendering is finished
 	}
 
--- a/cmd/kyma/alpha/deploy/cmd.go
+++ b/cmd/kyma/alpha/deploy/cmd.go
@@ -46,7 +46,6 @@ func NewCmd(o *Options) *cobra.Command {
 	cobraCmd.Flags().DurationVarP(&o.QuitTimeout, "quit-timeout", "", 1200*time.Second, "Time after which the deployment is aborted. Worker goroutines may still be working in the background. This value must be greater than the value for cancel-timeout.")
 	cobraCmd.Flags().DurationVarP(&o.HelmTimeout, "helm-timeout", "", 360*time.Second, "Timeout for the underlying Helm client.")
 	cobraCmd.Flags().IntVar(&o.WorkersCount, "workers-count", 4, "Number of parallel workers used for the deployment.")
-	cobraCmd.Flags().BoolVarP(&o.Verbose, "verbose", "v", false, "Enable verbose mode.")
 	return cobraCmd
 }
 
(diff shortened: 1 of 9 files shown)


cmd/kyma/alpha/deploy/cmd.go
```diff
@@ -89,10 +92,23 @@ func (cmd *command) Run() error {
...
 		BackoffMaxElapsedTimeSeconds:  60 * 5,
-		Log:                           log.Printf,
+		Log:                           cmd.getLogFunc(),
+	}
+
+	var updateCh chan deployment.ProcessUpdate
+	if cmd.opts.Verbose {
+		defer log.Println("Kyma deployed!")
+	} else {
+		// Start rendering async CLI UI
```
skhalash, line 102: Imho the function names are descriptive enough, which makes those comments redundant

cmd/kyma/alpha/deploy/cmd.go
```diff
@@ -120,3 +134,12 @@ func (cmd *command) validateFlags() error {
 
 	return nil
 }
+
+func (cmd *command) getLogFunc() func(format string, v ...interface{}) {
```
skhalash, line 138: Would it be possible to extract `getLogFunc()` to be reused by both `deploy` and `uninstall` commands?

cmd/kyma/alpha/uninstall/cmd.go
```diff
@@ -78,10 +80,23 @@ func (cmd *command) Run() error {
...
+		defer log.Println("Kyma uninstalled!")
+	} else {
+		// Start rendering async CLI UI
+		asyncUI := asyncui.AsyncUI{StepFactory: &cmd.Factory}
+		updateCh, err = asyncUI.Start()
+		if err != nil {
+			return err
+		}
+		defer asyncUI.Stop() // stop receiving update-events and wait until UI rendering is finished
+	}
```
skhalash, line 97: Hey, could you explain the logic behind it? Why is asyncui turned on only when the verbose flag is not set?

pkg/asyncui/asyncui.go
```diff
@@ -0,0 +1,143 @@
...
+	undeployComponentsPhaseMsg    string = "Undeploying Kyma"
+	deployComponentMsg            string = "Deploying component '%s'"
+)
+
+// AsyncUI renders the CLI ui based on receiving events
+type AsyncUI struct {
+	// used to create UI steps
+	StepFactory StepFactory
+	// channel provided by caller to retrieve errors
+	Errors chan<- error
```
skhalash, line 31: Do those 2 struct fields have to be exposed? I guess, they could be kept internal?

pkg/asyncui/asyncui.go
```diff
@@ -0,0 +1,143 @@
...
+	}()
+
+	return ui.updates, nil
+}
+
+func (ui *AsyncUI) sendError(err error) {
+	if err != nil && ui.Errors != nil {
+		ui.Errors <- err
+	}
+}
```
skhalash, line 82: As far as I see, this error delivery mechanism is not yet used (and also not tested). Maybe we can spit until we actually need it. What would you say?

skhalash (changes requested):

cmd/kyma/alpha/deploy/cmd.go
```diff
@@ -44,6 +46,7 @@ func NewCmd(o *Options) *cobra.Command {
 	cobraCmd.Flags().DurationVarP(&o.QuitTimeout, "quit-timeout", "", 1200*time.Second, "Time after which the deployment is aborted. Worker goroutines may still be working in the background. This value must be greater than the value for cancel-timeout.")
 	cobraCmd.Flags().DurationVarP(&o.HelmTimeout, "helm-timeout", "", 360*time.Second, "Timeout for the underlying Helm client.")
 	cobraCmd.Flags().IntVar(&o.WorkersCount, "workers-count", 4, "Number of parallel workers used for the deployment.")
+	cobraCmd.Flags().BoolVarP(&o.Verbose, "verbose", "v", false, "Enable verbose mode.")
```
suleymanakbas91, line 49: There is already a verbose flag on kyma level https://github.com/kyma-project/cli/blob/master/cmd/kyma/kyma.go#L52. Could you please reuse it instead of creating a new one?

suleymanakbas91 (changes requested):

cmd/kyma/alpha/uninstall/cmd.go
tobiscr: Sure, the reason is that the verbose-mode is just enabling the logging of the underneath executed Hydroform-API. The CLI should then be quiet to avoid redundant information shown to end-users.

pkg/asyncui/asyncui.go
tobiscr: Good point, thanks for bringing it up. The purpose of this field is to provide users of the Async-UI the option to retrieve errors / failures. Right now, the failures are just leading to failing steps but not communicated up to the caller of the Async-UI.

But it's worth to think whether this is:

- required at all
- needs further explanations (e.g. enhancing the comment or renaming the field name...)

skhalash (approved):

skhalash (approved):

suleymanakbas91 (approved):

Issue #690, opened 2021-01-13
New installation: support overrides provided as file and via CLI parameters


tobiscr:
**Description**

The new deployment command has to enable the user to provide override parameters in a file and as CLI parameters.

E.g. `-overridesFile pathToFile.yaml -override key1=value1 -override key2=value2`

**Reasons**

Overrides should be flexible configurable by customers by supporting file and additionally single override values as CLI parameter.

**Attachments**
-

PR #694, opened 2021-01-14
Support override flags

tobiscr:
**Description**

User can define overrides as file and via CLI parameters.

**Related issue(s)**
#690

### DIFF
--- a/cmd/kyma/alpha/deploy/cmd.go
+++ b/cmd/kyma/alpha/deploy/cmd.go
@@ -13,7 +13,7 @@ import (
 	"github.com/kyma-project/cli/internal/clusterinfo"
 	"github.com/kyma-project/cli/internal/kube"
 	"github.com/kyma-project/cli/pkg/asyncui"
-
+	"github.com/magiconair/properties"
 	"github.com/spf13/cobra"
 
 	"github.com/kyma-incubator/hydroform/parallel-install/pkg/components"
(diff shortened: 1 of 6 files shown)


cmd/kyma/alpha/deploy/cmd.go
```diff
@@ -13,7 +13,7 @@ import (
 	"github.com/kyma-project/cli/internal/clusterinfo"
 	"github.com/kyma-project/cli/internal/kube"
 	"github.com/kyma-project/cli/pkg/asyncui"
-
+	"github.com/magiconair/properties"
```
tobiscr, line 16: @clebs :; not sure if it's worth to use a properties-parser for validating hte CLI parameters. Maybe better to extract key/value pairs by our own. What do you think?

kazydek (approved):

cmd/kyma/alpha/deploy/cmd.go
clebs: Sounds good to me, it saves us some parsing and might be more helpful in the future, since it can recursively parse and supports struct tags.
Also a good point is, this package has no dependencies on its go.mod file (otherwise I would have advised against it).

tobiscr: 👍

clebs (approved):

Issue #697, opened 2021-01-15
New Installation: fix DNS resolution when using *.local.kyma.dev on local cluster


tobiscr:
**Description**

The K8s DNS resolution for `*.local.kyma.dev`domain resolves to `127.0.0.1` also if the requested target is a service running inside of the local Kubernetes cluster.

**Reasons**

Using the public resolvable `local.kyma.dev` domain name makes the local Kyma installation convenient as the local DNS resolver needs not to be modified (e.g. by adding host entries to `/etc/hosts` file). The Kubernetes DNS system has to be configured to resolve locally running services (e.g. `dex.local.kyma.dev`) to the K8s internal service IP instead delegating requests to external DNs resolvers.

It has to be defined how such fixes will be applied as this is a change that is coupled to the Kyma domain name and has no relation to a component or the underneath cluster.

**Attachments**

PR #699, opened 2021-01-15
New installation: Configure coreDNS when using kyma local dev domain

tobiscr:
**Description**

Configure coreDNS to properly resolve local Kyma dev domain names.

**Related issue(s)**
#697


cmd/kyma/alpha/deploy/localdomain.go
```diff
@@ -0,0 +1,72 @@
...
+)
+
+// LocalKymaDevDomain contains the default domain name if Kyma is locally deployed
+const LocalKymaDevDomain = "local.kyma.dev"
+
+// KymaLocalDomain configures the local kyma domain in Kubernetes DNS
+type KymaLocalDomain struct {
+	coreDNSConfig []byte
+	kubeClient    kubernetes.Interface
+}
```
skhalash, line 18: In my opinion, the name `KymaLocalDomain` does not really reflect what this type is doing (and is too similar to the constant `LocalKymaDevDomain` above). Maybe something like `LocalDomainDNSConfigurer`?

cmd/kyma/alpha/deploy/localdomain.go
```diff
@@ -0,0 +1,72 @@
...
+
+// New creates a new Kyma local domain instance
```
skhalash, line 20: Could you please document what's exactly happening here?

skhalash (changes requested):
Looks good! Just a few cosmetic comments.

cmd/kyma/alpha/deploy/localdomain.go
tobiscr: Agree, the name is not self-expressive enough. I adjusted it to "DNSConfigurer".

cmd/kyma/alpha/deploy/localdomain.go
tobiscr: Good point, as it's a not self-explaining change. Added more documentation to it.

skhalash (approved):

cmd/kyma/alpha/deploy/dnsconfigurer.go
```diff
@@ -0,0 +1,77 @@
...
+// NewDNSConfigurer creates a new DNS configurer instance
+func NewDNSConfigurer(kubeClient kubernetes.Interface) *DNSConfigurer {
+	return &DNSConfigurer{
+		kubeClient: kubeClient,
+	}
+}
+
+// ConfigureCoreDNS patches the Kubernetes CoreDNS to support the local kyma domain.
+// It adds the rewrite plugin of CoreDNS to delegate requests to *.local.kyma.dev to the Istio Ingress Gateway.
+func (kld *DNSConfigurer) ConfigureCoreDNS() error {
```
clebs, line 33: Do we really need a type with a single method? Or can this be just a package function?
If there are plans to make it configurable and with different utilities for DNS stuff I could see a type being useful, but at the moment it just adds boilerplate without benefit.

tobiscr: Good point, I usually prefer types to be a bit more flexible how to handle data. But agree, in this case it unnecessary code.

tobiscr: FIXED

clebs: It is always possible to turn this function into a method should we need it, but for now it does not hold any data or state.

clebs (approved):

cmd/kyma/alpha/deploy/dnsconfigurer.go
tobiscr: agree

PR #717, opened 2021-01-23
Run functions locally

pPrecel:
**Description**

- Add new command (`kyma run function`)
- Implement ability of running functions locally
- Upgrade the method of the finalization  (after `ctrl+c`)

**Related issue(s)**

https://github.com/kyma-project/kyma/issues/9508
https://github.com/kyma-incubator/hydroform/pull/186


internal/cli/finalizer.go
```diff
@@ -0,0 +1,54 @@
...
+
+import (
+	"fmt"
+	"os"
+	"os/signal"
+	"syscall"
+	"time"
+)
+
+const timeout = 5
```
clebs, line 11: Please add the unit of the timeout here, `5` by itself doesn't mean anything.
```suggestion
const timeout = 5 * time.Second
```

internal/cli/finalizer.go
```diff
@@ -0,0 +1,54 @@
...
+
+type Finalizer struct {
+	notify func(c chan<- os.Signal, sig ...os.Signal)
+	exit   func(int)
+	funcs  []func()
+}
+
+func NewFinalizer() *Finalizer {
+	return &Finalizer{
+		notify: signal.Notify,
```
clebs, line 21: This field is not needed, you can just call `signal.Notify()` directly in the relevant place.

internal/cli/finalizer.go
```diff
@@ -0,0 +1,54 @@
...
+		exit:   os.Exit,
```
clebs, line 22: This field is not needed, you can just call `os.Exit()` directly in the relevant place.

internal/cli/finalizer.go
```diff
@@ -0,0 +1,54 @@
...
+	}
+}
+
+func (f *Finalizer) Add(function func()) {
+	f.funcs = append(f.funcs, function)
+}
+
+func (f *Finalizer) SetupCloseHandler() {
```
clebs, line 30: Why have a separate function to activate the finaliser? I would expect this code to be part of `NewFinalizer()`.
The constructor should deliver a ready to use object

internal/cli/finalizer.go
```diff
@@ -0,0 +1,54 @@
...
+			lastChan <- struct{}{}
+		}()
+
+		go func() {
+			time.Sleep(time.Second * timeout)
+			lastChan <- struct{}{}
+		}()
+
+		<-lastChan
+		fmt.Printf("\r- Signal '%v' received from Terminal. Exiting...\n ", sig)
```
clebs, line 51: Shouldn't this message be printed in line 35? Finalizer functions could take a while to complete, giving the user the impression that the CLI froze.

cmd/main.go
```diff
@@ -1,34 +1,20 @@
...
 
 	_ "k8s.io/client-go/plugin/pkg/client/auth/gcp"
 )
 
 func main() {
-	setupCloseHandler()
-	command := kyma.NewCmd(cli.NewOptions())
+	fin := cli.NewFinalizer()
+	fin.SetupCloseHandler()
+	command := kyma.NewCmd(cli.NewOptions(fin))
```
clebs, line 14: Since there is only one initialisation for the finalizer and no dynamic fields at this point in time, maybe we can move this code to `cli.NewOptions()`.

internal/cli/options.go
```diff
@@ -10,9 +10,12 @@ type Options struct {
 	Verbose bool
 	step.Factory
 	KubeconfigPath string
+	Finalizer      *Finalizer
```
clebs, line 13: From the consumer perspective this contains 0 or more finalizers, not just 1. I would call the field `Finalizers`.
That way, when adding, it looks more natural: `opts.Finalizers.Add(...)`.

internal/cli/finalizer_test.go
```diff
@@ -0,0 +1,205 @@
+package cli
```
clebs, line 1: Please add `t.Parallel()` as the first line of tests whenever possible.

go.mod
```diff
@@ -5,48 +5,45 @@ go 1.14
 replace (
 	github.com/hashicorp/consul v0.0.0-20171026175957-610f3c86a089 => github.com/hashicorp/consul/sdk v0.7.0
 	github.com/hashicorp/consul/api v1.3.0 => github.com/hashicorp/consul/api v0.0.0-20191112221531-8742361660b6
+	github.com/kyma-incubator/hydroform/function => github.com/pPrecel/hydroform/function v0.0.0-20210206085647-39a02f919fe9
```
clebs, line 8: Please remove this replace directive.

go.mod
```diff
@@ -5,48 +5,45 @@ go 1.14
...
 	github.com/blang/semver/v4 v4.0.0
 	github.com/briandowns/spinner v1.12.0
 	github.com/containerd/cgroups v0.0.0-20201119153540-4cbc285b3327 // indirect
-	github.com/containerd/containerd v1.4.2 // indirect
-	github.com/containerd/continuity v0.0.0-20200710164510-efbc4488d8fe // indirect
 	github.com/coreos/etcd v3.3.13+incompatible // indirect
 	github.com/daviddengcn/go-colortext v1.0.0
-	github.com/docker/cli v0.0.0-20200130152716-5d0cf8839492
-	github.com/docker/docker v1.4.2-0.20200203170920-46ec8731fbce
+	github.com/docker/cli v20.10.3+incompatible
```
clebs, line 24: Both docker dependencies are heavily used by the `kyma install` command when installing from local sources into a remote cluster.
Did you test that it still works?

clebs (changes requested):

internal/cli/finalizer.go
pPrecel: 👍

internal/cli/finalizer.go
pPrecel: This function included in the struct body because testing purposes. I'm replacing this method with my fixed method during tests.

internal/cli/finalizer.go
pPrecel: This function included in the struct body because testing purposes. I'm replacing this method with my fixed method during tests.

internal/cli/finalizer.go
pPrecel: It is ready to use ;) And you can use it by the `SetupCloseHandler` method. In my opinion, both solutions are acceptable.

internal/cli/finalizer.go
pPrecel: Hmm you are right. I will change it

cmd/main.go
pPrecel: Finalization of the program is an important step in the life cycle of the application, exactly like running the main cmd line loop. I think that hiding it in the `cli.NewOptions()` method is not the best way to do ;) But of course, I can do this

internal/cli/options.go
pPrecel: good point!

internal/cli/finalizer_test.go
pPrecel: 👍

go.mod
pPrecel: I can't because my second PR is still in review. I will change it later (before merge)

go.mod
pPrecel: Good question. Maybe first I should try to migrate to the previous version of these dependencies

internal/cli/options.go
pPrecel: :)

internal/cli/finalizer.go
clebs: Ah I see... that is a good point. In that case I guess we can keep it to mock sys calls.

internal/cli/finalizer.go
clebs: If I call `NewFinalizer()`, I always have to call `SetupCloseHandler()` afterwards, there is no use case without calling the second function.
So we might as well simplify the contract and need only to call the constructor.

go.mod
clebs: I thought so yes, just a reminder so that we do not forget.

go.mod
clebs: It is fine to upgrade them, but we need to be sure nothing breaks.

cmd/main.go
clebs: Ok yes, having it explicitly can be a good reason to have it in the `main`.

Still it bothers me a little that it is a constructor parameter that might not be necessary many times (calling a constructor with `nil` params usually is a smell)

go.mod
pPrecel: Ok I will check this ;)

internal/cli/finalizer.go
pPrecel: But this will complicate my tests. After this PR I will take another issue around the finalizer (Issue is not created yet). Can we talk about it during next review?

cmd/main.go
pPrecel: Ok. I will move this to the NewOptions method ;)

go.mod
pPrecel: I tested provisioning and installation. Everything works for me

docs/gen-docs/kyma_run_function.md
```diff
@@ -0,0 +1,40 @@
...
+
+```bash
+kyma run function [flags]
+```
+
+## Options
+
+```bash
+      --containerName string   The name of the created container.
+  -d, --detach                 Change this flag to true if you don't want to follow the container logs after run'.
```
kazydek, line 19:
```suggestion
  -d, --detach                 Change this flag to "true" if you don't want to follow the container logs after running the Function.
```

docs/gen-docs/kyma_run_function.md
```diff
@@ -0,0 +1,40 @@
...
+  -e, --env stringArray        The system environments witch which the container will be run.
```
kazydek, line 20:
```suggestion
  -e, --env stringArray        The system environments the container will be run with.
```

docs/gen-docs/kyma_run_function.md
```diff
@@ -0,0 +1,40 @@
...
+  -f, --filename string        Full path to the config file.
+      --imageName string       Full name with tag of the container.
```
kazydek, line 22:
```suggestion
      --imageName string       Full name with the container's tag.
```

docs/gen-docs/kyma_run_function.md
```diff
@@ -0,0 +1,40 @@
...
+  -p, --port string            The port on which the container will be exposed. (default "8080")
```
kazydek, line 23:
```suggestion
  -p, --port string            The port on which the container will be exposed. (default "8080").
```

docs/gen-docs/kyma_run_function.md
kazydek: Or: `Full name with the tag of the container.` but I guess the above version is shorter.

docs/gen-docs/kyma_run_function.md
pPrecel: `(default "8080")` documentation generator adds this dot :D

docs/gen-docs/kyma_run_function.md
pPrecel: I removed this flag

docs/gen-docs/kyma_run_function.md
pPrecel: I removed this flag

docs/gen-docs/kyma_run_function.md
pPrecel: Done, thanks!

docs/gen-docs/kyma_run_function.md
```diff
@@ -0,0 +1,40 @@
...
+  -h, --help                Displays help for the command.
+      --kubeconfig string   Specifies the path to the kubeconfig file. By default, Kyma CLI uses the KUBECONFIG environment variable or "/$HOME/.kube/config" if the variable is not set.
+      --non-interactive     Enables the non-interactive shell mode.
+  -v, --verbose             Displays details of actions triggered by the command.
+```
+
+## See also
+
+* [kyma run](#kyma-run-kyma-run)	 - 
+
```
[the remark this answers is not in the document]
pPrecel: Yeaaah :D Done, thanks

docs/gen-docs/kyma_run_function.md
```diff
@@ -0,0 +1,40 @@
+---
+title: kyma run function
+---
+
+Run functions locally.
+
+## Synopsis
+
+Use this command to run function in docker from local sources.
```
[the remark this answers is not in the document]
pPrecel: Done! Thanks!

docs/gen-docs/kyma_run.md
```diff
@@ -0,0 +1,25 @@
...
+  -h, --help                Displays help for the command.
+      --kubeconfig string   Specifies the path to the kubeconfig file. By default, Kyma CLI uses the KUBECONFIG environment variable or "/$HOME/.kube/config" if the variable is not set.
+      --non-interactive     Enables the non-interactive shell mode.
+  -v, --verbose             Displays details of actions triggered by the command.
+```
+
+## See also
+
+* [kyma](#kyma-kyma)	 - Controls a Kyma cluster.
+* [kyma run function](#kyma-run-function-kyma-run-function)	 - Run functions locally.
```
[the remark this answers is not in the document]
pPrecel: Done done 👍

docs/gen-docs/kyma_run.md
```diff
@@ -0,0 +1,25 @@
+---
+title: kyma run
+---
+
+
+
+## Synopsis
+
+
```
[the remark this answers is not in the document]
pPrecel: yup, thanks a lot

docs/gen-docs/kyma_run_function.md
kazydek: Sorry, my bad.

kazydek (dismissed):
🎉

cmd/kyma/run/function/function.go
```diff
@@ -0,0 +1,140 @@
...
+		Use:   "function",
+		Short: "Runs Functions locally.",
+		Long:  `Use this command to run a Function in Docker from local sources.`,
+		RunE: func(cmd *cobra.Command, args []string) error {
+			return c.Run()
+		},
+	}
+
+	cmd.Flags().StringVarP(&o.Filename, "filename", "f", "", `Full path to the config file.`)
+	cmd.Flags().StringVarP(&o.Dir, "sourceDir", "d", "", `Full path to the folder with the source code.`)
```
m00g3n, line 40: Can we change `sourceDir` to `source-dir`?

cmd/kyma/run/function/function.go
```diff
@@ -0,0 +1,140 @@
...
+	cmd.Flags().StringVar(&o.ContainerName, "containerName", "", `The name of the created container.`)
```
m00g3n, line 41: Can we change `containerName` to `container-name`?

docs/gen-docs/kyma_run_function.md
```diff
@@ -0,0 +1,38 @@
...
+```
+
+## Options
+
+```bash
+      --containerName string   The name of the created container.
+      --detach                 Change this flag to "true" if you don't want to follow the container logs after running the Function.
+  -f, --filename string        Full path to the config file.
+  -p, --port string            The port on which the container will be exposed. (default "8080")
+  -d, --sourceDir string       Full path to the folder with the source code.
```
m00g3n, line 22: If we change the `sourceDir` to `source-dir` we need to update docs.

docs/gen-docs/kyma_run_function.md
```diff
@@ -0,0 +1,38 @@
...
+Use this command to run a Function in Docker from local sources.
+
+```bash
+kyma run function [flags]
+```
+
+## Options
+
+```bash
+      --containerName string   The name of the created container.
```
m00g3n, line 18: If we change the `containerName` to `container-name` we need to update docs.

internal/cli/finalizers.go
```diff
@@ -0,0 +1,58 @@
...
+
+	return fin
+}
+
+func (f *Finalizers) Add(function func()) {
+	f.funcs = append(f.funcs, function)
+}
+
+func (f *Finalizers) setupCloseHandler() {
+	c := make(chan os.Signal, 2)
```
m00g3n, line 34: Do we need buffer of 2 here?

cmd/kyma/run/run.go
```diff
@@ -0,0 +1,19 @@
...
+	"github.com/kyma-project/cli/internal/cli"
+	"github.com/spf13/cobra"
+)
+
+//NewCmd creates a new run command
+func NewCmd(o *cli.Options) *cobra.Command {
+	cmd := &cobra.Command{
+		Use:   "run",
+		Short: "Runs resources locally.",
+		Long:  `Use this command to run resources from the local configuration.`,
```
clebs, line 14: This might not be true for all subcommands. First level commands should have a very broad documentation since we will be adding many other subcommands in the future.

go.mod
```diff
@@ -5,6 +5,8 @@ go 1.14
 replace (
 	github.com/hashicorp/consul v0.0.0-20171026175957-610f3c86a089 => github.com/hashicorp/consul/sdk v0.7.0
 	github.com/hashicorp/consul/api v1.3.0 => github.com/hashicorp/consul/api v0.0.0-20191112221531-8742361660b6
+	// Remove this line before merge
+	github.com/kyma-incubator/hydroform/function => github.com/pPrecel/hydroform/function v0.0.0-20210222074129-f7f7a52ff8a0
```
clebs, line 9: Still can't approve with this line in the mod file

clebs (changes requested):

cmd/kyma/run/run.go
kazydek: @clebs This long description already seems quite generic - which part seems not to be true and how would you change it?

clebs: The part that might not be true is `local configuration`.

clebs (approved):

kazydek (approved):

PR #726, opened 2021-01-26
Adding Stas and Tobi as codeowner for CLI

tobiscr:
**Description**
Adding @tobiscr  and @skhalash as Codeowners for CLI

**Related issue(s)**


tobiscr:
Change was already applied by @suleymanakbas91

PR #737, opened 2021-02-01
Remove DNS patch - moved to installer lib

tobiscr:
**Description**

To be independent from CLI calls the DNS patch for `local.kyma.dev` will be moved to installer library and become part of the certificate handling.

**Related issue(s)**


strekm (approved):
/meow

suleymanakbas91 (approved):

Issue #755, opened 2021-02-15
New installation: Flag to toggle atomic upgrades


clebs:
**Description**

In a normal upgrade, if components can't be upgraded, the installation library will rollback. This can be problematic in non-productive clusters if we wish to debug the issues.

Add a flag to the CLI to toggle upgrade rollbacks:
`kyma alpha deploy --atomic=false` -> This will disable rollbacks when upgrading.

**Reasons**

Allow better debugging capabilities, rollbacks need to be disabled on demand.

**Attachments**
Depends on the support for the flag in Hydroform: https://github.com/kyma-incubator/hydroform/issues/195

tobiscr:
The PR for CLI requires the changes in PR https://github.com/kyma-incubator/hydroform/pull/198 as we have to update the Hydroform-API before this PR can be merged.

a-thaler:
Flag is not effective and even worse by default no rollback seem to happen, re-opened

tobiscr:
Counter part on hydroform side is https://github.com/kyma-incubator/hydroform/pull/209

tobiscr:
Please be aware to use this notation to disable the atomic flag: `--atomic=false` 

Tested it also with `--atomic false` and the atomic-flag was still set to `true`.

PR #757, opened 2021-02-16
Add atomic-deployment flag

tobiscr:
**Description**
Add support for atomic deployments.

**Related issue(s)**
https://github.com/kyma-project/cli/issues/755

**Depends on**
https://github.com/kyma-incubator/hydroform/pull/198

### DIFF
--- a/docs/gen-docs/kyma_alpha_deploy.md
+++ b/docs/gen-docs/kyma_alpha_deploy.md
@@ -15,6 +15,7 @@ kyma alpha deploy [flags]
 ## Options
 
 ```bash
+  -a, --atomic                    Use atomic deployment (rollback any component which could not be successfully installed). (default true)
       --cancel-timeout duration   Time after which the workers' context is canceled. Pending worker goroutines (if any) may continue if blocked by a Helm client. (default 15m0s)
   -c, --components string         Path to the components file. (default "workspace/installation/resources/components.yaml")
   -d, --domain string             Domain used for installation. (default "local.kyma.dev")
--- a/cmd/kyma/alpha/delete/cmd.go
+++ b/cmd/kyma/alpha/delete/cmd.go
@@ -39,7 +39,7 @@ func NewCmd(o *Options) *cobra.Command {
 		Aliases: []string{"d"},
 	}
 
-	cobraCmd.Flags().DurationVarP(&o.CancelTimeout, "cancel-timeout", "", 900*time.Second, "Time after which the workers' context is canceled. Pending worker goroutines (if any) may continue if blocked by a Helm client.")
+	cobraCmd.Flags().DurationVarP(&o.CancelTimeout, "cancel-timeout", "", 900*time.Second, "Time after which the workers' context is canceled. Any pending worker goroutines that are blocked by a Helm client will continue.")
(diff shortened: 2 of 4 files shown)


docs/gen-docs/kyma_alpha_deploy.md
```diff
@@ -15,6 +15,7 @@ kyma alpha deploy [flags]
 ## Options
 
 ```bash
+  -a, --atomic                    Use atomic deployment (rollback any component which could not be successfully installed). (default true)
```
NHingerl, line 18:
```suggestion
  -a, --atomic                    Use atomic deployment, which rolls back any component that could not be installed successfully (default: true).
```

docs/gen-docs/kyma_alpha_deploy.md
```diff
@@ -15,6 +15,7 @@ kyma alpha deploy [flags]
...
       --cancel-timeout duration   Time after which the workers' context is canceled. Pending worker goroutines (if any) may continue if blocked by a Helm client. (default 15m0s)
```
NHingerl, line 19: Does the timeout affect several workers or just one worker? If just one, use "worker's" instead of "worker'"

```suggestion
      --cancel-timeout duration   Time after which the workers' context is canceled (default: 15m0s). Any pending worker goroutines that are blocked by a Helm client will continue.
```

docs/gen-docs/kyma_alpha_deploy.md
```diff
@@ -15,6 +15,7 @@ kyma alpha deploy [flags]
...
   -c, --components string         Path to the components file. (default "workspace/installation/resources/components.yaml")
```
NHingerl, line 20:
```suggestion
  -c, --components string         Path to the components file (default: "workspace/installation/resources/components.yaml")
```

docs/gen-docs/kyma_alpha_deploy.md
```diff
@@ -15,6 +15,7 @@ kyma alpha deploy [flags]
...
   -d, --domain string             Domain used for installation. (default "local.kyma.dev")
```
NHingerl, line 21:
```suggestion
  -d, --domain string             Domain used for installation (default: "local.kyma.dev")
```

NHingerl (changes requested):
Not quite sure about my 2nd sentence for the timeout, pls check if I got it right.

docs/gen-docs/kyma_alpha_deploy.md
tobiscr: the timeout affects all workers

docs/gen-docs/kyma_alpha_deploy.md
tobiscr: Good input, but we need to handled that decoupled form this PR as it requires a change in the doc-generator logic.

tobiscr:
**TL;DR**
This PR will be discontinued and is now part of #763 .

**Long version**
We had a major change in the Hydroform API (logging was changed to use ZAP) which requires a bigger change on CLI side. To avoid duplicated work, I will close this PR and merge its changes into PR #763. This PR includes all changes to integrate the CLI with the latest Hydroform API.

@NHingerl : thanks for your valuable feedback - is it also added to PR #763. Please review the updates there.

Issue #613, opened 2020-10-14
Auto-generated names for functions


pPrecel:
**Description**

Add functionality of auto-generated default names of functions and projects after using the `kyma init function` command.

**Reasons**

Because the default name of generated functions is `first-function` we want to allow users to deploy more than one function in a namespace easily.

**Attachments**

[Suggestion from the review](https://github.com/kyma-project/cli/pull/606#discussion_r500957426)

PR #775, opened 2021-02-24
Add auto generated name for function

rJankowski93:
**Description**

- add generated name for function
- replace custom var with string
- fix test

**Related issue(s)**

#613


cmd/kyma/init/function/function.go
```diff
@@ -38,7 +39,7 @@ Use the flags to specify the initial configuration for your Function or to choos
 		},
 	}
 
-	cmd.Flags().StringVar(&o.Name, "name", defaultName, `Function name.`)
+	cmd.Flags().StringVar(&o.Name, "name", generateRandomFunctionName(), `Function name.`)
```
clebs, line 42: It would be better to only generate a name if the user is not providing the flag because random generation is costly.

pPrecel: This PR is still `Work In Progress` ;) We know that, but thanks for your suggestions

clebs: I see, in that case please create `Draft` pull requests so that reviewers are not being notified.

cmd/kyma/init/function/opts.go
```diff
@@ -37,6 +40,11 @@ func (o *Options) setDefaults(defaultNamespace string) (err error) {
 		}
 	}
 
+	if o.Name == "" {
+		rand.Seed(time.Now().UnixNano())
```
clebs, line 44: Seeding the RNG is only necessary if you want different pseudo-random sequences in each run, which is probably not needed here.

If you still think this is necessary though, I would expect the seeding to happen inside `GenerateName`. The consumer having to do it, leaks implementation details of `GenerateName`.

rJankowski93: I think this line is necessary, so I moved this to implementation GenerateName

clebs: It that case, this PR can be merged after https://github.com/kyma-incubator/hydroform/pull/217 and updating the `go.mod` file.

rJankowski93: Yes, of course, but I'm waiting for your review because if you will have another comments I will have to create new PR to hydroform.

clebs: AH I see, I do not have any further comments. As I said in the previous comment, as soon as the other PR is merged and the go.mod file updated here, this can be merged.

rJankowski93: Done

clebs (approved):

kazydek (approved):

PR #790, opened 2021-03-02
Fix minikube ip for docker driver

DimitarPetrov:
**Description**

`minikube ip` returns 127.0.0.1 in the case of docker driver (by design) which is not valid inside the cluster as it represents localhost for the container. Therefore test-definitions fail on minikube with docker driver.


clebs (approved):

clebs:
Hi @DimitarPetrov,

thanks for your PR! I have tested and merged it.

PR #796, opened 2021-03-08
Allow to skip kyma cert import command

kwiatekus:
**Description**

- Allow to skip import cert part for CI use case

### DIFF
--- a/cmd/kyma/alpha/deploy/cmd.go
+++ b/cmd/kyma/alpha/deploy/cmd.go
@@ -145,8 +145,8 @@ func (cmd *command) Run() error {
 	cmd.duration = time.Since(start)
 
 	// import certificate
-	qImportCertsStep := cmd.NewStep("Import Kyma certificate")
-	if !cmd.avoidUserInteraction() && qImportCertsStep.PromptYesNo("Import Kyma certificate locally? ") {
+	qImportCertsStep := cmd.NewStep("Install Kyma certificate locally")
+	if !cmd.avoidUserInteraction() && qImportCertsStep.PromptYesNo("Should the Kyma certificate be installed locally") {
 		if err := cmd.importCertificate(); err != nil {
 			return err
 		}
--- a/cmd/kyma/alpha/deploy/cmd.go
+++ b/cmd/kyma/alpha/deploy/cmd.go
@@ -54,6 +54,7 @@ func NewCmd(o *Options) *cobra.Command {
 
 	cobraCmd.Flags().StringVarP(&o.WorkspacePath, "workspace", "w", defaultWorkspacePath, "Path used to download Kyma sources.")
 	cobraCmd.Flags().BoolVarP(&o.Atomic, "atomic", "a", true, "Use atomic deployment, which rolls back any component that could not be installed successfully.")
+	cobraCmd.Flags().BoolVarP(&o.SkipCertImport, "skip-cert-import", "", false, "Set to true if you want to skip importing kyma cert to the host machine.")
 	cobraCmd.Flags().StringVarP(&o.ComponentsFile, "components", "c", defaultComponentsFile, "Path to the components file.")
 	cobraCmd.Flags().StringSliceVarP(&o.OverridesFiles, "values-file", "f", []string{}, "Path to a JSON or YAML file with configuration values.")
 	cobraCmd.Flags().StringSliceVarP(&o.Overrides, "value", "", []string{}, "Set a configuration value (e.g. --value component.key='the value').")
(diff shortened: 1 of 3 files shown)


clebs (dismissed):

docs/gen-docs/kyma_alpha_deploy.md
```diff
@@ -22,6 +22,7 @@ kyma alpha deploy [flags]
       --helm-timeout duration     Timeout for the underlying Helm client. (default 6m0s)
   -p, --profile string            Kyma deployment profile. If not specified, Kyma is installed with the default chart values. The supported profiles are: "evaluation", "production".
       --quit-timeout duration     Time after which the deployment is aborted. Worker goroutines may still be working in the background. This value must be greater than the value for cancel-timeout. (default 20m0s)
+      --skip-cert-import          Set to true if you want to skip importing kyma cert to the host machine.
```
NHingerl, line 25: Pls confirm specifically how this value must be set - I assume similar to the --atomic flag? "Set `--atomic=false` to..."

--

```suggestion
      --skip-cert-import          Set `--skip-cert-import=true" to skip importing the Kyma certificate to the host machine.
```

(is it "the" or "a" certificate?

NHingerl (changes requested):
clarify command syntax

cmd/kyma/alpha/deploy/cmd.go
```diff
@@ -145,8 +146,10 @@ func (cmd *command) Run() error {
 	cmd.duration = time.Since(start)
 
 	// import certificate
-	if err := cmd.importCertificate(); err != nil {
-		return err
+	if !cmd.opts.SkipCertImport {
```
a-thaler, line 149: Why we use a new flag for that? For me that is the regular CI or non-interactive scenario, so please re-use the existing flags we have.

docs/gen-docs/kyma_alpha_deploy.md
clebs: To clarify, boolean flags can be set in 2 different ways:
- To set it to true, both `--atomic` and `--atomic=true` are equivalent.
- To set it t false, `--atomic=false` and not passing the flag at all are equivalent.

cmd/kyma/alpha/deploy/cmd.go
```diff
@@ -145,8 +145,14 @@ func (cmd *command) Run() error {
 	cmd.duration = time.Since(start)
 
 	// import certificate
-	if err := cmd.importCertificate(); err != nil {
-		return err
+	qImportCertsStep := cmd.NewStep("Install Kyma certificate locally")
+	if !cmd.avoidUserInteraction() && qImportCertsStep.PromptYesNo("Should the Kyma certificate be installed locally") {
```
clebs, line 149:
```suggestion
	if !cmd.avoidUserInteraction() && qImportCertsStep.PromptYesNo("Should the Kyma certificate be installed locally?") {
```

cmd/kyma/alpha/deploy/cmd.go
```diff
@@ -145,8 +145,14 @@ func (cmd *command) Run() error {
...
+		if err := cmd.importCertificate(); err != nil {
+			return err
+		}
+		qImportCertsStep.Success()
+	} else {
+		qImportCertsStep.Failure()
```
clebs, line 155: Why should the step fail if the user choses not to import the certificate or we are in non interactive mode?
Failing means something went wrong.

cmd/kyma/alpha/deploy/cmd.go
```diff
@@ -145,8 +145,14 @@ func (cmd *command) Run() error {
 	cmd.duration = time.Since(start)
 
 	// import certificate
-	if err := cmd.importCertificate(); err != nil {
-		return err
+	qImportCertsStep := cmd.NewStep("Install Kyma certificate locally")
```
clebs, line 148: It would make sense to move this step inside of `cmd.importCertificate()` and print manual import instructions in case the user choses to not automatically do it.

clebs (changes requested):

cmd/kyma/alpha/deploy/cmd.go
```diff
@@ -415,6 +414,11 @@ func (cmd *command) printSummary(o deployment.Overrides) error {
 }
 
 func (cmd *command) importCertificate() error {
+	if !cmd.approveImportCertificate() {
+		//no approval given: stop import
+		return nil
```
clebs, line 419: Please print manual import instructions here. The user might want to import the certificate himself later.
Just need to call `cert.Instructions()`

tobiscr: aaah .. good point. Will add that.

clebs (approved):

PR #823, opened 2021-03-29
feat(make): added ARM build targets

### LINKED ISSUE #822: Build targets for ARM
**Description**

Adding new build targets covering ARM platforms (`Windows`, `Linux`, `MacOS`).

**Reasons**

It would be a good idea, especially after increasing number of ARM devices appearing on the market in recent and upcoming months - e.g. laptops with *Apple M1* or *Microsoft SQ2*.

**Attachments**

`None`

Shanduur:
**Description**

- Added build targets for ARM:
    - `GOOS=windows GOARCH=arm64`
    - `GOOS=darwin GOARCH=arm64`
    - `GOOS=linux GOARCH=arm64`

**Related issue(s)**

Resolves #822

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -26,17 +26,33 @@ validate:
 .PHONY: build
 build: build-windows build-linux build-darwin
 
+.PHONY: build-experimental
+build-experimental: build-windows-arm build-linux-arm build-darwin-arm
+
 .PHONY: build-windows
 build-windows:
 	CGO_ENABLED=0 GOOS=windows GOARCH=amd64 go build -o ./bin/kyma.exe $(FLAGS) ./cmd
+	
+.PHONY: build-windows-arm
+build-windows-arm:
+	CGO_ENABLED=0 GOOS=windows GOARCH=arm64 go build -o ./bin/kyma.exe $(FLAGS) ./cmd
 
 .PHONY: build-linux
 build-linux:
 	CGO_ENABLED=0 GOOS=linux GOARCH=amd64 go build -o ./bin/kyma-linux $(FLAGS) ./cmd
 
+.PHONY: build-linux-arm
+build-linux-arm:
+	CGO_ENABLED=0 GOOS=linux GOARCH=arm64 go build -o ./bin/kyma-linux-arm $(FLAGS) ./cmd
+
 .PHONY: build-darwin
 build-darwin:
 	CGO_ENABLED=0 GOOS=darwin GOARCH=amd64 go build -o ./bin/kyma-darwin $(FLAGS) ./cmd
+
+.PHONY: build-darwin-arm
+build-darwin-arm:
+	go version | grep '1.16'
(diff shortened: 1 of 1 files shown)


clebs:
Hi @Shanduur,

thanks for your PR! I just tested it and it seems the goal `make build-windows-arm` is not correct.
`GOARCH` should be `arm` instead of `arm64`.

Shanduur:
Thanks for pointing this out @clebs, I was sure I tested that. I pushed fix for this target.

Shanduur:
Whoops, I think I messed this up a bit, with exiting with exit code 1, if Go Version is lower than 1.16. Should this just skip that build target?

Makefile
```diff
@@ -24,20 +24,33 @@ validate:
...
+build-linux-arm:
+	CGO_ENABLED=0 GOOS=linux GOARCH=arm64 go build -o ./bin/kyma-linux-arm $(FLAGS) ./cmd
+
 .PHONY: build-darwin
 build-darwin:
 	CGO_ENABLED=0 GOOS=darwin GOARCH=amd64 go build -o ./bin/kyma-darwin $(FLAGS) ./cmd
 
+.PHONY: build-darwin-arm
+build-darwin-arm:
+	go version | grep '1.16'
```
clebs, line 51: Hi @Shanduur, the build for ARM on Mac OSX is not working because of this grep. Our CI/CD is running on go 1.15.7 and thus it fails.

Is ARM support for OSX only available on g 1.16?

clebs:
Hi @Shanduur, if ARM support works on OSX on go 1.15 then yes, removing that check should fix this.

Shanduur:
As stated in the [relase blog post](https://blog.golang.org/go1.16) target for ARM MacOS is set to `darwin/arm64` since 1.16. Previously this target corresponded to iOS devices, so it will compile only on Macs, and it won't run on Apple M1 devices. I thought about adding MacOS as experimental build target, at least until you update the version in *CI* and *go.mod*:
```make
.PHONY: build-experimental
build-experimental: build-darwin-arm

.PHONY: build-darwin-arm
build-darwin-arm:
        CGO_ENABLED=0 GOOS=darwin GOARCH=arm64 go build -o ./bin/kyma-darwin-arm $(FLAGS) ./cmd
```

Makefile
Shanduur: Unfortunately on Go <1.16 outside MacOS this will result with:
```
/home/user/.go/pkg/tool/linux_amd64/link: running gcc failed: exit status 1
/tmp/go-link-271834784/go.o: file not recognized: file format not recognized
collect2: error: ld returned 1 exit status

make: *** [Makefile:54: build-darwin-arm] Error 2
```

clebs: As discussed, ARM support on Mac OSX is commented out until our CI/CD runs on go 1.16.

clebs (approved):

PR #831, opened 2021-03-31
Migrate the CLI repo from master to main

clebs:
**Description**

- Migrate the CLI repo from master branch to main


suleymanakbas91 (changes requested):
I think we should also change the source target `master` to `main`.

clebs:
Hi @suleymanakbas91,

that change will happen next week when the https://github.com/kyma-project/kyma repo is migrated.
At the moment kyma source is still master.

suleymanakbas91 (approved):

Issue #771, opened 2021-02-23
Allow exposing functions via kyma CLI


kwiatekus:
**Description**

As a CLI user I want to expose my function so that it is accessible via HTTP
 - [x] Allow to define api rules as part of function manifest ( config.yaml )
 - [x] The API rule spec should allow for `allow` and `jwt` access strategies
 - [x] API rules should be deployed in the runtime after `apply` command
 - [x] API rules should be read from the runtime and written to the manifest file after `sync` comannd
 - [ ] STRETCH add a new `expose` command that would modify the manifest file

**Reasons**
ATM it's not possible to expose function via kyma CLI
Users have to add api rules via  console UI or kubectl.

kwiatekus:
```yaml
name: my-function
namespace: my-namespace
runtime: nodejs12
source:
    sourceType: inline
    sourcePath: ...
triggers:
    - version: v1
      source: ec-mock
      type: order.deliverysent
apirules:
      .........      
  ```

PR #838, opened 2021-04-06
Allow to define api rules as part of function manifest

pPrecel:
**Description**

- As in the title

**Related issue(s)**

Old PR: https://github.com/kyma-project/cli/pull/813
https://github.com/kyma-project/cli/issues/771

### DIFF
--- a/cmd/kyma/apply/function/function.go
+++ b/cmd/kyma/apply/function/function.go
@@ -89,6 +89,11 @@ func (c *command) Run() error {
 		return err
 	}
 
+	apiRule, err := resources.NewApiRule(configuration)
+	if err != nil {
+		return err
+	}
+
 	if configuration.Source.Type == workspace.SourceTypeGit {
 		gitRepository, err := resources.NewPublicGitRepository(configuration)
 		if err != nil {
@@ -102,6 +107,7 @@ func (c *command) Run() error {
 		[]operator.Operator{
 			operator.NewSubscriptionOperator(client.Resource(operator.GVRSubscription).Namespace(configuration.Namespace),
 				configuration.Name, configuration.Namespace, subscriptions...),
+			operator.NewGenericOperator(client.Resource(operator.GVKApiRule).Namespace(configuration.Namespace), apiRule),
 		},
 	)
 
--- a/cmd/kyma/apply/function/function.go
+++ b/cmd/kyma/apply/function/function.go
@@ -16,6 +16,7 @@ import (
 	"github.com/pkg/errors"
 	"github.com/spf13/cobra"
 	"gopkg.in/yaml.v2"
+	v1 "k8s.io/apimachinery/pkg/apis/meta/v1"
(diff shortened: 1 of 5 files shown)


cmd/kyma/apply/function/function.go
```diff
@@ -122,6 +131,22 @@ func (c *command) Run() error {
 	return mgr.Do(ctx, options)
 }
 
+func (c *command) kymaHostAddress() string {
+	var url string
+	vs, err := c.K8s.Istio().NetworkingV1alpha3().VirtualServices("kyma-system").Get(context.Background(), "apiserver-proxy", v1.GetOptions{})
```
clebs, line 136: As far as I know, `apiserver-proxy` is no longer going to be available soon and this code will stop working.
There are many alternatives to obtain the kyma host from; I would suggest getting it from `kyma-gateway` like this (kubectl example):
`kubectl get gateway kyma-gateway -n kyma-system -o yaml`

cmd/kyma/apply/function/function.go
```diff
@@ -122,6 +131,22 @@ func (c *command) Run() error {
...
+	switch {
+	case err != nil:
+		fmt.Printf("Unable to read the Kyma host URL due to error: %s. \n%s\r\n", err.Error(),
```
clebs, line 139: Please do not print errors and then continue execution as if nothing happened. This will then fail further down the line and the printed output will be confusing and make it very hard to debug what is really going on.

clebs (changes requested):

cmd/kyma/apply/function/function.go
pPrecel: This function represents additional and optional functionality. Cli may work fine without obtained the kyma address. But you are right that debugging (mostly output log) may be harder to understand. So I will add another Step to log this error as a warning during the working step

cmd/kyma/apply/function/function.go
pPrecel: I will investigate this. Thanks for this information! In this case, I'm getting the address from the apiserver-proxy virtualservice because this service is related to apiRules that this PR is about

cmd/kyma/apply/function/function.go
pPrecel: [Done](https://github.com/kyma-project/cli/pull/838/files#diff-eeb658a08e2c3a583900fa0a93f72515685f64d39eff93dd12c434d50ed04489R99-R102)

cmd/kyma/apply/function/function.go
clebs: Yes it is not wrong, the issue is that `apiserver-proxy` will disappear soon and this code will fail.

We can just make it correctly long term now instead of waiting for it to fail.
It is only changing this one line with the gateway and we are good to go, the rest is all fine.

pPrecel: Done. Thanks a lot!

clebs (approved):

Issue #826, opened 2021-03-29
New Installation: Allowing single component deployment


tobiscr:
**Description**

To improv the development speed it has to be possible to install of one or more defined components deployment. 

The CLI has to support an option to provide a list of components (as arguments, e.g. `--component abc --component xyz@namespace`).

**Reasons**

Improve development speed and offer a more flexible way how to install custom Kyma components beside using a file (component-list).

**Attachments**

a-thaler:
Adjust examples at https://kyma-project.io/docs/main/cli/overview/#details-kyma-cli-alpha-command-usage-examples please

PR #851, opened 2021-04-12
Component flag usage example

shorim:
**Description**

- modify usage example to use `--components-file` flag  instead of  `--components` flag.
- add usage example for the new add flag `--component`.

**Related issue(s)**

#826

### DIFF
--- a/docs/03-03-examples-alpha.md
+++ b/docs/03-03-examples-alpha.md
@@ -61,7 +61,7 @@ For example, to install Kyma from a specific version, such as `1.19.1`, run:
 - To deploy Kyma with only specific components, run:
 
   ```
-  kyma alpha deploy --components {COMPONENTS_FILE_PATH}
+  kyma alpha deploy --components-file {COMPONENTS_FILE_PATH}
   ```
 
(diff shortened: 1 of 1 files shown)


docs/03-03-examples-alpha.md
```diff
@@ -78,6 +78,18 @@ For example, to install Kyma from a specific version, such as `1.19.1`, run:
     - name: "dex"
   ```
 
+- Alternatively, you can specify single components instead of a file:
+  
+  ```
+  kyma alpha deploy --component {COMPONENT_NAME@NAMESPACE}
+  ```
+
+  If no namespace is provided, then the default namespace is used. For example, to install `testing` component in the default namespace and `application-connector` component in `kyma-integration` namespace, run:
```
alexandra-simeonova, line 87:
```suggestion
  If no Namespace is provided, then the default Namespace is used. For example, to install the `testing` component in the default Namespace and the `application-connector` component in the `kyma-integration` Namespace, run:
```

alexandra-simeonova (approved):

tobiscr (approved):

Issue #787, opened 2021-03-01
New installation: support further k3d specific command parameters in CLI


tobiscr:
**Description**

The `alpha provision k3d` supports just a subset of available K3d parameters. Users have to be able to pass further parameters to K3d, independent to the options offered by the Kyma CLI.

Example:

    kyma alpha provision k3s --k3d-args "--a-k3d-parameter" --k3d-args "--another-k3d-parameter"

**Reasons**

Allow customers to define further k3d parameters.

**Attachments**

PR #856, opened 2021-04-14
K3d further parameters

shorim:
**Description**

- allow users to add further parameters through the `--k3d-arg` flag which are passed to the k3d provisioning command.
- add `--agent-arg` flag which mirrors the `--k3s-agent-arg` flag in the k3d provisioning command.

**Related issue(s)**

#787

### DIFF
--- a/internal/k3s/k3s.go
+++ b/internal/k3s/k3s.go
@@ -127,7 +127,7 @@ func ClusterExists(verbose bool, clusterName string) (bool, error) {
 }
 
 //StartCluster starts a cluster
-func StartCluster(verbose bool, timeout time.Duration, clusterName string, workers int, serverArgs []string) error {
+func StartCluster(verbose bool, timeout time.Duration, clusterName string, workers int, serverArgs []string, k3dArgs []string) error {
 	cmdArgs := []string{
 		"cluster", "create", clusterName,
 		"--kubeconfig-update-default",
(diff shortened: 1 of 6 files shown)


internal/k3s/k3s.go
```diff
@@ -148,6 +148,14 @@ func StartCluster(verbose bool, timeout time.Duration, clusterName string, worke
 		cmdArgs = append(cmdArgs, srvArg)
 	}
 
+	//add further k3d args which are not offered by the Kyma CLI flags
+	for _, k3dArg := range k3dArgs {
+		if k3dArg == "" {
```
clebs, line 153: Maybe it would be interesting to see what happens here if someone passes something like:
`--k3d-arg="     "`

shorim: ah good catch.
It will then return an error to the user `FATA[0000] accepts between 0 and 1 arg(s), received 2 `.
We can do one of the 2 things:
1- either we even remove the current check `if k3dArg == ""` and let the `k3d` command return an error to the user whenever he passes an empty flag
2- or we could adapt our check and make it trim trailing and leading spaces and then check if the string is empty

clebs: I was having the option 1 in mind, since it will transparently tell exactly what arguments are wrong.

internal/k3s/k3s.go
```diff
@@ -141,18 +141,18 @@ func StartCluster(verbose bool, timeout time.Duration, clusterName string, worke
...
-			continue
-		}
 		cmdArgs = append(cmdArgs, "--k3s-server-arg")
 		cmdArgs = append(cmdArgs, srvArg)
 	}
 
+	//add agent args
+	for _, agentArg := range agentArgs {
+		cmdArgs = append(cmdArgs, "--k3s-agent-arg")
+		cmdArgs = append(cmdArgs, agentArg)
```
clebs, line 151:
```suggestion
		cmdArgs = append(cmdArgs, "--k3s-agent-arg", agentArg)
```

clebs (approved):

alexandra-simeonova (approved):

PR #870, opened 2021-04-22
Improve defaulting of sourcePath in the function config

### LINKED ISSUE #828: Improve defaulting of sourcePath in the function config
**Description**

Now, output from the `kyma init function` looks like this:
```
workspaceVersion: ""
name: function-sharp-karolina9
namespace: default
runtime: nodejs12
source:
    sourceType: inline
    sourcePath: /tmp/tmp
```
We don't need to always show user property `source.sourcePath` because it's always `$PWD` by default. We can change it to `.` or just remove it from generated config. In the second option, the user can always override the default `sourcePath` by adding it to the config, but the default value for this property should be a folder where `config.yaml` exists. 

My proposition:
- Remove `source.sourcePath` from the config template by default
- Default value (if there is no value for `source.sourcePath` in the `config.yaml`) is folder where the `config.yaml` exists
- User can override this value by adding a custom path (to the `source.sourcePath`)

**Reasons**

Better user experience. Why is the default value for this field in the config? I think that the user doesn't need to understand how kymaCli apply resources to the cluster and he doesn't need to see default values in config which can be confusing for him. But he still should be possible to set his own value, for example, if he locates the code of his functions in another folder than config.

igwejk:
*Description*

- Remove source.sourcePath from the config template by default
- Default value (if there is no value for source.sourcePath in the config.yaml) is folder where the config.yaml exists
- User can override this value by adding a custom path (to the source.sourcePath)

Fixes #828

### DIFF
--- a/cmd/kyma/init/function/opts.go
+++ b/cmd/kyma/init/function/opts.go
@@ -49,7 +49,6 @@ func (o *Options) setDefaults(defaultNamespace string) (err error) {
 		o.RepositoryName = o.Name
 	}
 
-	setIfZero(&o.SourcePath, o.Dir)
 	setIfZero(&o.Namespace, defaultNamespace)
 	return
 }


pPrecel (approved):
Looks great! Thanks for your contribution

clebs (approved):

igwejk:
Thanks for the reviews @pPrece and @clebs.

Issue #854, opened 2021-04-14
New Installation: workspace directory name and location should be changed


tobiscr:
**Description**

The default location of the `workspace` directory is the current working directly. This can cause unnecessary permission issues. Also the name `workspace` is not the best choice, as it  is a common name and sometimes used by users to store their work related data.

The workspace directory should be 

1. located in a place where no write-permissions are expected (e.g. `$HOME`)
2. be place either in a sub-directory (e.g.  `$HOME/.kyma-cli/ws` ) or using a name which has a low risk to collide with existing directories (e.g. `kyma-cli-ws`)

**Reasons**

Users were complaining the the workspace folder location is colliding with their existing folder structure.

**Attachments**

clebs:
Hi @tobiscr,
this is a good improvement. The CLI already has a `~/.kyma` folder that we can use.
See:
https://github.com/kyma-project/cli/blob/8234ebcf09634b8df687801530689bb85f701c4e/internal/files/files.go#L12

PR #874, opened 2021-04-22
Change default workspace for deployment

shorim:
**Description**

- Change default workspace for deployment from `workspace` to `$HOME/.kyma/sources`. `$HOME/.kyma` directory is already used for provisioning terraform clusters. So. it is a good idea to reuse it here and also the directory `workspace` was sometimes colliding with existing directories from the users.
- In case there was an error while trying to get the home directory of the user, then the direcory `.kyma-sources` is used as a fallback.

**Related issue(s)**

#854

### DIFF
--- a/cmd/kyma/alpha/deploy/opts.go
+++ b/cmd/kyma/alpha/deploy/opts.go
@@ -11,6 +11,7 @@ import (
 
 	"github.com/kyma-incubator/hydroform/parallel-install/pkg/download"
 	"github.com/kyma-project/cli/internal/cli"
+	"github.com/kyma-project/cli/internal/files"
 )
 
 const (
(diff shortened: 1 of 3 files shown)


NHingerl (dismissed):

cmd/kyma/alpha/deploy/opts.go
```diff
@@ -165,3 +172,11 @@ func (o *Options) pathExists(path string, description string) error {
 func (o *Options) workspaceTmpDir() string {
 	return filepath.Join(o.WorkspacePath, "tmp")
 }
+
+func getDefaultWorkspacePath() string {
+	kymaHome, err := files.KymaHome()
+	if err != nil {
+		return filepath.Join(".kyma-sources")
```
clebs, line 179: Is there a reason to use `filepath.Join()` here?
When the parameter list is < 2 it does nothing.

shorim: yes, it is not needed here. I removed it.

clebs (approved):

NHingerl (approved):

PR #883, opened 2021-04-29
Allow #836 CLI patch releases

### LINKED ISSUE #836: Allow doing patch releases from CLI without Kyma release
**Description**

Right now Kyma CLI releases are bound to Kyma releases because of the default Kyma version set in the [Homebrew formula](https://github.com/Homebrew/homebrew-core/blob/master/Formula/kyma-cli.rb#L21). That's why we cannot do hot fixes on CLI and release a new patch version if there is no new version being released from Kyma. We should find another way to set the default Kyma version so that we can release Kyma CLI patches without releasing Kyma.

**Acceptance Criteria**
- Allow releasing Kyma CLI patches without releasing Kyma.

elchead:
Description
new PR: squashed commit history

bind Kyma version to CLI when explicitly stating VERSION
get latest patch version from Prow registry
Related issue(s)
Fixes #836

image
image


clebs (dismissed):

clebs (dismissed):

elchead:
@superojla , @mmitoraj, @majakurcius: Could some of you please look over the issue for it to be included in the next release? Thanks :)

cmd/kyma/install/cmd.go
```diff
@@ -40,7 +38,7 @@ func NewCmd(o *Options) *cobra.Command {
 		Use:   "install",
 		Short: "Installs Kyma on a running Kubernetes cluster.",
 		Long: `Use this command to install Kyma on a running Kubernetes cluster.
-
+		
 ### Detailed description
```
NHingerl, line 42: Do we have "vague description" too? What does "detailed" mean, is this word really needed?

cmd/kyma/install/cmd.go
```diff
@@ -54,15 +52,15 @@ The standard installation uses the minimal configuration. The system performs th
 
 1. Deploys and configures the Kyma Installer. At this point, steps differ depending on the installation type.
 
-   When you install Kyma locally ` + "**from release**" + `, the system:
+When you install Kyma locally ` + "**from release**" + `, the system:
 
-   1. Fetches the latest or specified release along with configuration.
+1. Fetches the latest or specified release along with configuration.
```
NHingerl, line 57:
```suggestion
1.    Fetches the latest or specified release along with configuration.
```

I believe we need the indentation for proper formatting.

cmd/kyma/install/cmd.go
```diff
@@ -54,15 +52,15 @@ The standard installation uses the minimal configuration. The system performs th
```
NHingerl, line 52: Do I understand correctly, the system performs only two "main" steps, but depending on the installation type, main step 1 has two different sets of 6 substeps, of which the last 3 are identical for both types?  Can we elevate those last 3 substeps (overrides, admin pw, minikube patch) to main step level and only keep the ones that are different below the introduction "the following steps differ..."

NHingerl (dismissed):

clebs (approved):

elchead:
@NHingerl could you remove your required changes? I made the changes but it still seems to block

clebs (approved):

PR #893, opened 2021-05-17
Output permission error for k3s provision

elchead:
**Description**

For linux root access is required to reserve port 80. Improve error output
- ...
- ...
- ...

**Related issue(s)**

### DIFF
--- a/cmd/kyma/alpha/provision/k3s/cmd.go
+++ b/cmd/kyma/alpha/provision/k3s/cmd.go
@@ -84,9 +84,10 @@ func (c *command) verifyK3sStatus() error {
 			s.Failure()
 			return err
 		}
-	} else if c.portAllocated(80) || c.portAllocated(443) {
-		s.Failuref("Port 80 or 443 are already in use. Please stop the allocating service and try again.")
-		return fmt.Errorf("Port 80 or 443 are already in use")
+	} else if err := c.portAllocated(80, 443); err != nil {
+		errorMsg := fmt.Sprintf("Port 80 or 443 cannot be allocated:%s", err)
+		s.Failure()
+		return fmt.Errorf(errorMsg)
 	}
 
 	s.Successf("K3s status verified")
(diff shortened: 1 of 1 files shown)


cmd/kyma/alpha/provision/k3s/cmd.go
```diff
@@ -114,12 +115,17 @@ func (c *command) deleteExistingK3sCluster() error {
 }
 
 //Check if a port is allocated
-func (c *command) portAllocated(port int) bool {
-	con, err := net.Listen("tcp", fmt.Sprintf(":%d", port))
-	if con != nil {
-		con.Close()
+func (c *command) portAllocated(ports ...int) error {
```
clebs, line 118: I would rename this function to something different if it no longer returns a boolean.
For example `allocatePorts`.

cmd/kyma/alpha/provision/k3s/cmd.go
```diff
@@ -84,9 +84,10 @@ func (c *command) verifyK3sStatus() error {
 			s.Failure()
 			return err
 		}
-	} else if c.portAllocated(80) || c.portAllocated(443) {
-		s.Failuref("Port 80 or 443 are already in use. Please stop the allocating service and try again.")
-		return fmt.Errorf("Port 80 or 443 are already in use")
+	} else if err := c.portAllocated(80, 443); err != nil {
+		errorMsg := fmt.Sprintf("Port 80 or 443 cannot be allocated:%s", err)
+		s.Failure()
+		return fmt.Errorf(errorMsg)
```
clebs, line 90:
```suggestion
		s.Failure()
		return errors.Wrap(err, "Port 80 or 443 cannot be allocated")
```

clebs (changes requested):

cmd/kyma/alpha/provision/k3s/cmd.go
```diff
@@ -114,12 +115,17 @@ func (c *command) deleteExistingK3sCluster() error {
...
-		con.Close()
+func (c *command) portAllocated(ports ...int) error {
+	for _, port := range ports {
+		con, err := net.Listen("tcp", fmt.Sprintf(":%d", port))
+		if err != nil {
+			return err
+		}
+		if con != nil {
+			con.Close()
+		}
```
clebs, line 126: In case of error connection would stay open!
```suggestion
		con, err := net.Listen("tcp", fmt.Sprintf(":%d", port))
		defer con.Close()
		if err != nil {
			return err
		}
```

PR #909, opened 2021-06-02
Set fixed k3s image version

### LINKED ISSUE #907: Default k3s version for provision command
**Description**
The current "provision k3s" command uses the "latest" k3s version which is using the docker image with "latest" tag. That is quite indeterministic, old CLI installations might use an outdated k3s version, new installation will use the newest which might not be supported yet by kyma.

Goal: Have a configurable default which will set the recommended k8s version for the current kyma version
Criterias:
- Using a specific CLI version should result always in the same k3s major/minor version used for provisioning k3s, no matter when the CLI was installed
- there is a flag to set the k8s/k3s version, overwriting the default value

Hint: `k3d cluster create kyma --image rancher/k3s:v1.20.4-k3s1` will install k3s in a specific version

**Reasons**

**Attachments**

Recently updated k8s versions for the other provisioning commands: https://github.com/kyma-project/cli/pull/908

chrkl:
**Description**

- Use fixed k3s container image version and add CLI flag to override it. The k3s provisioning was always using the latest image before.

**Related issue(s)**

Fixes #907

### DIFF
--- a/docs/gen-docs/kyma_alpha_provision_k3s.md
+++ b/docs/gen-docs/kyma_alpha_provision_k3s.md
@@ -15,12 +15,13 @@ kyma alpha provision k3s [flags]
 ## Flags
 
 ```bash
-  -a, --agent-arg strings    One or more arguments passed to the k3s agent command on agent nodes (e.g. --agent-arg='--alsologtostderr')
-      --k3d-arg strings      One or more arguments passed to the k3d provisioning command (e.g. --k3d-arg='--no-rollback')
-      --name string          Name of the Kyma cluster (default "kyma")
-  -s, --server-arg strings   One or more arguments passed to the Kubernetes API server (e.g. --server-arg='--alsologtostderr')
-      --timeout duration     Maximum time for the provisioning. If you want no timeout, enter "0". (default 5m0s)
-      --workers int          Number of worker nodes (k3s agents) (default 1)
+  -a, --agent-arg strings     One or more arguments passed to the k3s agent command on agent nodes (e.g. --agent-arg='--alsologtostderr')
+      --k3d-arg strings       One or more arguments passed to the k3d provisioning command (e.g. --k3d-arg='--no-rollback')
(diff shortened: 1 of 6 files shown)


a-thaler:
Shouldn't we simplify it a bit for the user and only provide the option of setting a kubernetes version like we do in the other provisioning commands? Of course it is much more explicit to expose the full image but it is not consistent with other commands and makes it more complicated as people would need to understand what it means..

clebs (approved):

docs/gen-docs/kyma_alpha_provision_k3s.md
```diff
@@ -15,12 +15,13 @@ kyma alpha provision k3s [flags]
...
 ```bash
-  -a, --agent-arg strings    One or more arguments passed to the k3s agent command on agent nodes (e.g. --agent-arg='--alsologtostderr')
-      --k3d-arg strings      One or more arguments passed to the k3d provisioning command (e.g. --k3d-arg='--no-rollback')
-      --name string          Name of the Kyma cluster (default "kyma")
-  -s, --server-arg strings   One or more arguments passed to the Kubernetes API server (e.g. --server-arg='--alsologtostderr')
-      --timeout duration     Maximum time for the provisioning. If you want no timeout, enter "0". (default 5m0s)
-      --workers int          Number of worker nodes (k3s agents) (default 1)
+  -a, --agent-arg strings     One or more arguments passed to the k3s agent command on agent nodes (e.g. --agent-arg='--alsologtostderr')
+      --k3d-arg strings       One or more arguments passed to the k3d provisioning command (e.g. --k3d-arg='--no-rollback')
+  -k, --kube-version string   Kubernetes version of the cluster. (default "1.20.6")
```
majakurcius, line 20:
```suggestion
  -k, --kube-version string   Kubernetes version of the cluster (default "1.20.6")
```
Please remove the comma to be consistent with the other lines :)

majakurcius (approved):

PR #965, opened 2021-08-30
Don't fail if override file is not present

elchead:
**Description**
The CLI should skip if the override file is not present issuing a warning

- bump hydroform version
- ...
- ...

**Related issue(s)**

Require prior merge https://github.com/kyma-project/hydroform/pull/338

### DIFF
--- a/cmd/kyma/deploy/cmd.go
+++ b/cmd/kyma/deploy/cmd.go
@@ -213,7 +213,7 @@ func (cmd *command) Run() error {
 	overridesStep := cmd.NewStep("Applying Kyma2 overrides")
 	if err := overrides.AddFile(filepath.Join(cmd.opts.WorkspacePath, kyma2OverridesPath)); err != nil {
 		if errors.Is(err, fs.ErrNotExist) {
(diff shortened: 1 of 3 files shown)


cmd/kyma/deploy/cmd.go
```diff
@@ -209,9 +210,15 @@ func (cmd *command) Run() error {
 	}
 
 	// TODO remove this block when default component values are migrated to kyma 2.0
+	overridesStep := cmd.NewStep("Applying Kyma2 overrides")
 	if err := overrides.AddFile(filepath.Join(cmd.opts.WorkspacePath, kyma2OverridesPath)); err != nil {
-		return errors.Wrap(err, "Could not add overrides for Kyma 2.0")
+		if errors.Is(err, fs.ErrNotExist) {
+			overridesStep.LogErrorf("Kyma2 override path not found but continuing: %s", err)
```
a-thaler, line 216: It should be logged only in verbose mode as it is nothing important if it cannot be found. The overall goal is to get rid of the file

JeremyHarisch (approved):

PR #972, opened 2021-09-02
Alpha deploy reconciler

lindnerby:
**Description**

- Moved the functionality needed for `kyma alpha deploy` to internal CLI packages: `coredns`, `overrides`, `k3d`, `garener`
- Implement a new reconciler-based `kyma alpha deploy`

**How to test it on a k3d cluster**
```
brew install istioctl
make build-darwin
./bin/kyma-darwin provision k3d
ISTIOCTL_PATH=/usr/local/bin/istioctl ./bin/kyma-darwin alpha deploy
```

**Related issue(s)**


internal/coredns/coredns.go
```diff
@@ -0,0 +1,247 @@
...
+		if isK3s {
+			domainName = coreDNSLocalDomainName
+		} else {
+			domainName = coreDNSRemoteDomainName
+		}
+		patches["Corefile"], err = generateCorefile(domainName)
+		if err != nil {
+			return nil, err
+		}
+	}
```
jeremyharisch, line 165: I think it would be better to first check if its a K3d cluster, before trying to fetch the Gardener domain, to save up some calls.
Furthermore, we changed the naming from `K3s` to `K3d` in most part of Kyma (also in CLI command), thus we should also change it in this new command to `K3d`.

skhalash: Agree, but since it's untested functionality (there are literally 0 tests) we just moved from the parallel-install pkg, I'd not refactor the logic. I'll apply the renaming.

alexandra-simeonova (dismissed):

alexandra-simeonova (approved):

rakesh-garimella (approved):

Issue #934, opened 2021-07-01
New dashboard command to simplify access to busola


a-thaler:
**Description**
With [kyma 2](https://github.com/kyma-project/kyma/issues/11337) there will be a new central console as a managed service. That can be used for any non-local cluster and it will be great to get the URL in an easy way at hands after kyma got installed.
In the local setup, you need to run your own instance in order to have a dashboard.

Goal: Have a new command simplifying both scenarios of getting a kyma dashboard in local and non-local setup

Acceptance Criteria:
- There is a dedicated command (`kyma dashboard`) to open the central dashboard in your browser. The central dashboard is hosted at https://dashboard.kyma-project.io/
- The command also prints the URL to be opened before it starts opening the browser
- If the user wants it (prompt required which can be silenced, the current `kubecontext` should be directly used to connect the dashboard to your cluster. See https://pages.github.tools.sap/kyma/busola-url-creator/. We could also upload automatically but reject kubeconfigs which are not OIDC based
- The command supports a flag (--local) to run the dashboard local via docker and then opens the local URL in the browser, again connecting the dashboard with the cluster immediately
- The local mode should be blocking and needs to be interrupted in order to terminate, similar to the "kubectl proxy" command
- `kyma deploy` should mention the central dashboard URL in the summary but also should promote the new command to quickly run/open/connect the dashboard
- By default a fixed port should be used for the local startup, having a CLI flag to change it
- If there is a docker container running with the same name, fail gracefully

**Reasons**
Make it simpler to use the dashboard with a fresh cluster

**Attachments**

```
docker run --rm -p 3001:3001 busola/local
```

Sawthis:
Correct command to run Busola locally:
```
docker run --rm -p 3001:3001 eu.gcr.io/kyma-project/busola
```

skhalash:
@a-thaler Should we just implement it as a stable `kyma dasbhoard` command, since `kyma deploy` is already stable?

a-thaler:
@skhalash yes

Furthermore we discovered that local busola on docker cannot access local k3d when running with Docker Desktop. Busola team will add a env variable to make it working. The CLI would need to detect the specific environment and set the env variable

PR #999, opened 2021-09-30
Add dashboard command

dennis-ge:
**Description**

Added a new `dashboard` command that automatically opens the Kyma dashboard ([https://github.com/kyma-project/busola](https://github.com/kyma-project/busola)). As requested, I added a flag to also run the dashboard locally via Docker. 

**Related issue**
Implements #934


internal/docker/docker.go
```diff
@@ -0,0 +1,101 @@
+package docker
```
dennis-ge, line 1: This contents of this file are mostly from the [hydroform/function](https://github.com/kyma-project/hydroform/blob/main/function/pkg/docker/run.go) package which is used in the `kyma run function` command. I could not use this package alone because a command  is executed and a directory is mounted  for each container, which is not necessary for the dashboard command. Hence, I created this file with a slightly modified functionality.

lindnerby: Perfect 👍

lindnerby:
Tests are missing. At least for the new command we need some.

lindnerby (approved):

NHingerl (approved):

lindnerby (approved):

PR #1029, opened 2021-10-11
Command to add certificates to local storage

### LINKED ISSUE #335: Commands for local setup
**Description**
The "kyma install" command is covering three aspects currently, installation, watch install, setup local infra.
The three parts should be executable independently. For the watch there is https://github.com/kyma-project/cli/issues/41 already, the local setup is missing yet.

The motivation is: my colleague sends me a kubeconfig to an existing remote cluster. In order to debug it I need to setup helm with certificates, install xip certificate, get the password and open the console. All that steps are already implemented in the "kyma install" but not applyable for an existing cluster

As with kyma2 there is no password present anymore, the goal can be summarized to:
Goal: Extract "adding certtificates to local storage" and "adding entries to host file" as dedicated commands to the CLI
Criterias:
- logic will not be called anymore as part of the kyma2 installation
- new commands requires to be executed with elevated permissions (sudo)
- kyma installation requires no sudo password anymore
- if logic is not implemented for windows at least a manual gets printed on what to do

**Reasons**
It is very useful in daily life

**Attachments**

Pranav-SA:
**Description**

- Addition of 'kyma add-certificates' command
- Requires elevated permissions (sudo) for adding certificates
- Deletion of certificate addition logic from deploy command

**Related issue(s)**
Resolves #335

### DIFF
--- a/docs/gen-docs/kyma.md
+++ b/docs/gen-docs/kyma.md
@@ -23,7 +23,7 @@ Kyma CLI allows you to install, test, and manage Kyma.
 
 ## See also
 
-* [kyma add-certificates](#kyma-add-certificates-kyma-add-certificates)	 - Adds certtificates to local storage.
+* [kyma add-certificates](#kyma-add-certificates-kyma-add-certificates)	 - Adds certificates to local storage.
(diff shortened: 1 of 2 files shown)


clebs:
LGTM, but I would suggest that someone else approves since I have been away for some time and might miss some context on the task at hand

PR #1040, opened 2021-10-15
reduce default count of k3d agents to 0

dennis-ge:
**Description**

- change the default count of k3d agents to 0. By doing this we prevent the `too many open files` bug.
- the `--kubeconfig-switch-context` flag is also added by default because it is a desired behavior

**Related issue(s)**
Fixes #1026


dennis-ge:
> Can't we find a way to increase the file handle limit (either in Docker or k3d)?

There is a docker run option called ulimit (see [here](https://docs.docker.com/engine/reference/commandline/run/)) but I haven't found a way of setting it in k3d

PR #1042, opened 2021-10-18
add alpha provision k3d command for k3d v5

dennis-ge:
**Description**

- introduce `kyma alpha provision k3d` to leverage k3d v5 
- refactoring of `internal/k3d` 
- introduce `internal/k3d.cmRunner` interface that allows the mocking of `os/exec` calls

**Related issue(s)**


NHingerl:
I know it's been discussed before, but I still wonder why we have both `--ci` and `--non-interactive`. What's the difference that makes worth keeping them both?

dennis-ge:
>  know it's been discussed before, but I still wonder why we have both --ci and --non-interactive. What's the difference that makes worth keeping them both?

@NHingerl, there is already a issue for the next sprint to remove --non-interactive :)

PR #1087, opened 2021-11-16
Add ability to override default k3d registry settings

dbadura:
**Description**

- add ability to override local k3d registry

**Related issue(s)**

### DIFF
--- a/internal/deploy/values/builder.go
+++ b/internal/deploy/values/builder.go
@@ -13,8 +13,9 @@ import (
 )
 
 type builder struct {
-	files  []string
-	values []map[string]interface{}
+	files                   []string
+	values                  []map[string]interface{}
+	dockerRegistryOverrides map[string]interface{}
 }
 
 func (b *builder) addValuesFile(filePath string) *builder {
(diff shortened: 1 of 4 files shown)


internal/deploy/values/builder.go
```diff
@@ -13,8 +13,9 @@ import (
 )
 
 type builder struct {
-	files  []string
-	values []map[string]interface{}
+	files                   []string
+	values                  []map[string]interface{}
+	dockerRegistryOverrides map[string]interface{}
```
skhalash, line 18: The bug that you discovered is the fact that value file overrides (--value-files) do not override default values, only the value overrides (-value) do. So it is applicable to `addGlobalDomainName` and `addGlobalTLSCrtAndKey`. Let's give this field a more generic name and also change above mentioned methods.
```suggestion
	defaultValues map[string]interface{}
```

skhalash: Let's also prefix all 3 methods with default to be consistent:
```go
addDefaultServerlessRegistryConfig
addDefaultGlobalDomainName
addDefaultGlobalTLSCrtAndKey
```

internal/deploy/values/builder.go
```diff
@@ -83,16 +85,21 @@ func (b *builder) mergeSources() (map[string]interface{}, error) {
...
-			return nil, err
+			return nil, errors.Wrap(err, "while merging overrides values")
 		}
 	}
 
+	err := mergo.Map(&result, b.dockerRegistryOverrides)
+	if err != nil {
+		return nil, errors.Wrap(err, "while merging docker registry overrides")
+	}
+
```
skhalash, line 102: Let's move it up (before values and files with `mergo.Overrides`) to keep the logical flow:
1. First we apply defaults
2. Then we add files
3. Then we add values

skhalash (changes requested):

skhalash (approved):

Issue #1088, opened 2021-11-16
Implement fish and powershell completion


pPrecel:
**Description**

Now kymaCLI allows generating completion for the `bash` and the `zsh` shells. To improve flexibility and level of being user-friendly it's a good move to add support for generating completion for the [fish](https://fishshell.com/) and the [powershell](https://docs.microsoft.com/en-us/powershell/) too.

**Attachments**

[Shell completion](https://github.com/spf13/cobra/blob/master/shell_completions.md) using cobra

PR #1089, opened 2021-11-16
Implement fish and powershell completion

pPrecel:
**Description**

- add the possibility to generate completion scripts for the fish and the PowerShell
- improve command description in general

**Related issue(s)**

https://github.com/kyma-project/cli/issues/1088

### DIFF
--- a/cmd/kyma/completion/cmd.go
+++ b/cmd/kyma/completion/cmd.go
@@ -50,10 +50,10 @@ Fish:
 
 Powershell:
 
-  PS> yourprogram completion powershell | Out-String | Invoke-Expression
+  PS> kyma completion powershell | Out-String | Invoke-Expression
 
   # To load completions for every new session, run:
-  PS> yourprogram completion powershell > yourprogram.ps1
+  PS> kyma completion powershell > kyma.ps1
   # and source this file from your PowerShell profile.
 `,
 		RunE:    completion,
--- a/cmd/kyma/completion/cmd.go
+++ b/cmd/kyma/completion/cmd.go
@@ -2,7 +2,6 @@ package completion
 
 import (
 	"fmt"
-	"io"
 	"os"
 
 	"github.com/pkg/errors"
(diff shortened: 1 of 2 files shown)


clebs (approved):

clebs (approved):

cmd/kyma/completion/cmd.go
```diff
@@ -15,7 +15,46 @@ func NewCmd() *cobra.Command {
...
+  $ kyma completion bash > /usr/local/etc/bash_completion.d/kyma
+
+Zsh:
+
+  $ source <(kyma completion zsh)
+
+  # To load completions for each session, execute once:
+  $ kyma completion zsh > "${fpath[1]}/_kyma"
+
+  # You will need to start a new shell for this setup to take effect.
```
NHingerl, line 37: If users need to start the new shell _before_ running the command above, we might want to change the order of instructions. First start a new shell, then run the command.

cmd/kyma/completion/cmd.go
```diff
@@ -15,7 +15,46 @@ func NewCmd() *cobra.Command {
...
+
+  # If shell completion is not already enabled in your environment,
+  # you will need to enable it.  You can execute the following once:
```
NHingerl, line 40:
```suggestion
  # you must enable it. Execute the following once:
```

"can" (it's not optional right?) and "will" are not needed. "must" is simpler than "need to" :D

cmd/kyma/completion/cmd.go
```diff
@@ -15,7 +15,46 @@ func NewCmd() *cobra.Command {
...
+Zsh:
+
+  $ source <(kyma completion zsh)
+
+  # To load completions for each session, execute once:
+  $ kyma completion zsh > "${fpath[1]}/_kyma"
+
+  # You will need to start a new shell for this setup to take effect.
+
+  # If shell completion is not already enabled in your environment,
```
NHingerl, line 39: I'm not a fan of splitting sentences across two lines, but you probably had reasons? 
Otherwise, I suggest to add the remaining words in this line, and drop them from the next line.

```suggestion
  # If shell completion is not already enabled in your environment, you must enable it.
```

cmd/kyma/completion/cmd.go
pPrecel: Users need to start a new shell after running these commands because it will cause refreshing shell configuration and cause it working

NHingerl (approved):

clebs (approved):

PR #1090, opened 2021-11-17
Resolve Checkmarx vulnerabilities - download istioctl

jeremyharisch:
**Description**

This PR should resolve Checkmarx vulnerabilities.

- Check if `..` exists in path
- Use buffered Reader

**Related issue(s)**

### DIFF
--- a/internal/deploy/istioctl/istio.go
+++ b/internal/deploy/istioctl/istio.go
@@ -385,18 +385,18 @@ func unZip(source, target string, deleteSource bool) error {
 		if err != nil {
 			return err
 		}
+		defer dstFile.Close()
 
 		fileInArchive, err := f.Open()
 		if err != nil {
 			return err
 		}
+		defer fileInArchive.Close()
 
 		if err := copyInChunks(dstFile, fileInArchive); err != nil {
 			return err
 		}
 
-		dstFile.Close()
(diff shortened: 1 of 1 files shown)


dennis-ge (dismissed):

internal/deploy/istioctl/istio.go
```diff
@@ -357,13 +358,13 @@ func unTar(source, target string, deleteSource bool) error {
 }
 
 func unZip(source, target string, deleteSource bool) error {
-	archive, err := zip.OpenReader(filepath.Clean(source))
+	source = filepath.Clean(source)
+	zipReader, err := initReader(source)
 	if err != nil {
 		return err
 	}
-	defer archive.Close()
```
chrkl, line 364: Why do we not have to close the zipReader?

jeremyharisch: Because we do not open it anymore, thus it cannot be closed

jeremyharisch: Without opening it anymore, we should not have the problem of resource exhaustion anymore. PLus, we do no need to open it, to extract the zip

chrkl (approved):

chrkl (approved):

PR #1105, opened 2021-11-30
Root access error message as per OS

Pranav-SA:
**Description**

- Error message change for windows. 
Elevated permissions requirement message - sudo for linux/darwin, administrator role for windows

### DIFF
--- a/internal/root/root_others.go
+++ b/internal/root/root_others.go
@@ -8,9 +8,9 @@ import (
 )
 
 // IsWithSudo tells if a command is runnig with root privileges on
-func IsWithSudo() (bool,error) {
+func IsWithSudo() (error) {
 	if os.Getenv("SUDO_UID") != "" {
-		return true, nil
+		return nil
 	}
-	return false, errors.New("Elevated permissions are required to make entries to host file. Make sure you are using sudo")
+	return errors.New("Elevated permissions are required to make entries to host file. Make sure you are using sudo")
 }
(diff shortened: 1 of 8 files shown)


internal/root/root_others.go
```diff
@@ -4,9 +4,13 @@ package root
 
 import (
 	"os"
+	"github.com/pkg/errors"
 )
 
 // IsWithSudo tells if a command is runnig with root privileges on
-func IsWithSudo() bool {
-	return os.Getenv("SUDO_UID") != ""
+func IsWithSudo() (bool,error) {
```
clebs, line 11: Since now this function returns error when not running on sudo and nil error when running is sudo, the boolean return value is innecessary and can be removed.

clebs (changes requested):

internal/root/root_others.go
Pranav-SA: Yes,   Sure.

lindnerby (approved):

internal/root/root_windows.go
```diff
@@ -8,12 +8,12 @@ import (
 )
 
 // IsWithSudo tells if a command is runnig with root privileges on
-func IsWithSudo() (bool,error){
+func IsWithSudo() (error){
```
clebs:
```suggestion
func IsWithSudo() error {
```

clebs (approved):

PR #1148, opened 2022-01-19
fix CLI version for non-release is always 'stable'

### LINKED ISSUE #1147: CLI version for non-release is always 'stable'
**Description**
When building a binary from a random CLI revision, the `kyma version` command returns a version with prefix `stable``

```
$ bin/kyma-darwin version
Kyma CLI version: stable-3-gc934862
```

The intention was to have the branch name as part of the version to quickly indicate from what branch it is taken. The stable prefix should be used only for binaries created out of the `stable` tag

**Expected result**
The version of non-releases should contain the branch name as prefix. Only the stable binaries should have the stable prefix.

**Actual result**
stable prefix gets used always

**Steps to reproduce**

**Troubleshooting**

dennis-ge:
**Description**

- change CLI version for non-releases to the pattern `<branch-name>-<commit-hash>`
- if the current commit has the stable tag, the CLI version should be the same as before: `stable-<commit-hash>`

**Related issue(s)**
Fixes #1147

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -97,3 +97,4 @@ ci-main: resolve validate build test integration-test upload-binaries
 
 .PHONY: ci-release
 ci-release: resolve validate build test integration-test archive release
+
--- a/Makefile
+++ b/Makefile
@@ -1,11 +1,11 @@
 .DEFAULT_GOAL := local
 
 ifndef VERSION
-	VERSION=${shell git rev-parse --abbrev-ref HEAD}-${shell git rev-parse --short HEAD} # branch-commit_hash
-endif
-
-ifeq (${shell git rev-list -n 1 tags/stable},${shell git rev-parse HEAD}) # if stable tag is the same as current commit
-	VERSION = stable-${shell git rev-parse --short HEAD}
+	ifeq (${shell git rev-list -n 1 tags/stable},${shell git rev-parse HEAD}) # if stable tag is the same as current commit
+		VERSION = stable-${shell git rev-parse --short HEAD}
+	else
+		VERSION = ${shell git rev-parse --abbrev-ref HEAD}-${shell git rev-parse --short HEAD} # branch-commit_hash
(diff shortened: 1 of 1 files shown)


Makefile
```diff
@@ -1,10 +1,10 @@
 .DEFAULT_GOAL := local
 
 ifndef VERSION
-	VERSION = ${shell git describe --tags --always}
+	VERSION=${shell git rev-parse --abbrev-ref HEAD}-${shell git rev-parse --short HEAD} # branch-commit_hash
 endif
 
-ifeq ($(VERSION),stable)
+ifeq (${shell git rev-list -n 1 tags/stable},${shell git rev-parse HEAD}) # if stable tag is the same as current commit
```
a-thaler, line 7: Thinking about the release poccess I'm wondering if that could overwrite a custom VERSION.
Maybe move the block also into the `ifndef VERSION` block to be on the safe side. A custom VERSION must win always

Makefile
```diff
@@ -1,11 +1,11 @@
 .DEFAULT_GOAL := local
 
 ifndef VERSION
-	VERSION = ${shell git describe --tags --always}
-endif
-
-ifeq ($(VERSION),stable)
-	VERSION = stable-${shell git rev-parse --short HEAD}
+	ifeq (${shell git rev-list -n 1 tags/stable},${shell git rev-parse HEAD}) # if stable tag is the same as current commit
```
chrkl, line 4: This will show an error to the user if the stable tag is not existing: 
```
fatal: ambiguous argument 'tags/stable': unknown revision or path not in the working tree.
Use '--' to separate paths from revisions, like this:
'git <command> [<revision>...] -- [<file>...]'
```

I expect this will happen to most of the users since tags have to be fetch explicitly. Maybe we can suppress the error if no tags are available.

chrkl (approved):

Issue #1180, opened 2022-02-08
undeploy: use deployed kyma version


clebs:
**Description**

The undeploy command should delete the exact same version of Kyma that can be found on the cluster because a different version could have changes on the charts and/or component list that would make the removal fail.

At the moment `main` is used as default undeploy source and optionally the user can provide a source via flag.

The flag should be removed and the undeploy source should be the one retrieved from the cluster.

**Expected result**

Undeploy uses the same kyma version as the one installed in the cluster.

**Actual result**

`main` is always used if flag is not provided

**Steps to reproduce**

Does not apply.

clebs:
As discussed with the team, we will follow up by better documenting the `--source` flag in the `undeploy` command since it should never be used in normal circumstances.
#1182

PR #1181, opened 2022-02-09
undeply: detect source kyma version from cluster

clebs:
**Description**

- Undeploy source version is now automatically detected from the cluster.
- Optionally source version can be overridden with the source flag.
- Removed unnecessary replace directives in go.mod.

**Related issue(s)**
#1180


cmd/kyma/undeploy/cmd.go
```diff
@@ -197,3 +201,22 @@ func (cmd *command) initialSetup(wsp string) error {
 	preReqStep.Success()
 	return nil
 }
+
+// resolveVersion determines which version of kyma has to be used for undeploy.
+// Version is automatically detected from the cluster and can be optionally overridden via flag.
+func (cmd *command) resolveVersion() error {
+	if cmd.opts.Source != "" {
+		return nil
+	}
```
ruanxin, line 210: If the version offered in cmd is not same as the version in the cluster, cluster version should win, or?
So this check should be here I think

clebs: It is the other way around, version in the flag should override what is on the cluster. It is a bypass in case the version on the cluster is wrong or something like that.

rakesh-garimella (dismissed):

cmd/kyma/undeploy/cmd.go
```diff
@@ -45,7 +45,7 @@ func NewCmd(o *Options) *cobra.Command {
 	cobraCmd.Flags().StringSliceVarP(&o.Components, "component", "", []string{}, "Provide one or more components to undeploy (e.g. --component componentName@namespace)")
 	cobraCmd.Flags().StringVarP(&o.ComponentsFile, "components-file", "c", "", `Path to the components file (default "$HOME/.kyma/sources/installation/resources/components.yaml" or ".kyma-sources/installation/resources/components.yaml")`)
 	cobraCmd.Flags().StringVarP(&o.WorkspacePath, "workspace", "w", "", `Path to download Kyma sources (default "$HOME/.kyma/sources" or ".kyma-sources")`)
-	cobraCmd.Flags().StringVarP(&o.Source, "source", "s", config.DefaultKyma2Version, `Source of installation to be undeployed:
+	cobraCmd.Flags().StringVarP(&o.Source, "source", "s", "", `Undeploy source usually automatically detected, use this flag to override the detection:
```
NHingerl, line 48:
```suggestion
	cobraCmd.Flags().StringVarP(&o.Source, "source", "s", "", `Undeploy source usually automatically detected. Use this flag to override the detection:
```

NHingerl (dismissed):

cmd/kyma/undeploy/cmd.go
ruanxin: I'm a bit confused, how is the version can be wrong on the cluster? If the kyam get deployed, then the version either fixed or after upgrade, it can be changed, then do the undeploy, we should always take the latest version in cluster as source, and download kyma repo based on that version, or I understood it in wrong way?

ruanxin (approved):

Issue #1122, opened 2021-12-13
Local function execution fails when the handler is in a subdirectory


marcelreppi:
**Description**

When I place the source code of my Kyma function into a subdirectory like `src`, run `kyma run function` and go to my browser and go to `http://localhost:8080/` it crashes with the error `Error: Cannot find module '/kubeless/handler.js'`.
Based on the [documentation](https://kyma-project.io/docs/kyma/latest/05-technical-reference/svls-06-function-configuration-file/) I would assume this should work.
Deploying this setup actually works fine in our Kyma cluster but locally the execution fails.

**Expected result**

The browser returns "Hello Serverless"

**Actual result**

There is an internal server error and the process crashes with the error `Error: Cannot find module '/kubeless/handler.js'`

**Steps to reproduce**

1. `kyma init function -d my-function`
2. `cd my-function`
3. `mkdir src`
4. `mv handler.js src`
5. Add `sourceHandlerName: src/handler.js` to `source` in the `config.yaml`
6. `kyma run function`
7. Open `http://localhost:8080/` in your browser

**Troubleshooting**

I tried different combinations of the `sourcePath` and `sourceHandlerName` fields but nothing worked.

kwiatekus:
thank you @Cortey 
This fixes the problem.

I can reference `sourcePath` as `src` subfolder and run it locally
```yaml
name: my-function
namespace: default
runtime: nodejs14
source:
    sourceType: inline
    sourcePath: src
```
or I can reference Handler from `src` folder
```yaml
name: my-function
namespace: default
runtime: nodejs14
source:
    sourceType: inline
    sourceHandlerName: src/handler.js
```

kwiatekus:
Fixed in Kyma CLI  -> version: `main-fd29944`

PR #1231, opened 2022-03-31
Fix for local functions running from subfolders

Cortey:
**Description**
Fix that allows user to move their function resources to subfolders and still run the function locally using `kyma run function`
**See also**
https://github.com/kyma-project/hydroform/pull/352
https://github.com/kyma-project/cli/issues/1122


moelsayed (approved):
/lgtm

moelsayed (approved):

cmd/kyma/run/function/function.go
```diff
@@ -130,11 +128,26 @@ func workspaceConfig(path string) (workspace.Cfg, error) {
 	if _, ok := supportedRuntimes[cfg.Runtime]; !ok {
 		return workspace.Cfg{}, fmt.Errorf("unsupported runtime: %s", cfg.Runtime)
 	}
-
+	sourceFile, depsfile, _ := workspace.InlineFileNames(cfg.Runtime)
```
clebs, line 131: Why is the last parameter being ignored? I am afraid I do not know what it represents from the implementation since it has no godocs.

Cortey: It returns a boolean value that represents if the runtime is supported or not which is not needed as its already validated in the 3 lines above.

clebs: I suspected that.
In that case, I would rethink how this supported runtimes topic is handled. This explicit check outside the function library looks like an implementation detail leak to me, checking the runtime should happen internally and be baked into all public API calls that need it (e.g. return a specific `RuntimeNotSupportedError` value instead of a boolean).

Cortey: I added the line to confirm runtime support. To change the function in hydroform we will need to look at all of its usages which will be best to be handled in a separate issue/pull request. I will definetly raise the topic to our PO during the next sync.

clebs: I agree, please create a follow up issue to refactor this on hydroform.

clebs (approved):

PR #1266, opened 2022-05-11
Bump hydroform function package

dbadura:
**Description**

- use newer hydroform

**Related issue(s)**
[](https://github.com/kyma-project/kyma/issues/11155)


pPrecel:
I have a small problem because when I'm trying to apply function workspace with a few labels then the output function is created without any additional labels :(

```yaml
name: function-gifted-michal5
namespace: default
runtime: nodejs14
labels:
  any-label: any-value
  something-else: else-value
source:
    sourceType: inline
    sourcePath: /tmp/funcs
```

pPrecel (approved):

clebs (approved):

Issue #1301, opened 2022-07-04
Command: create module


clebs:
**Description**

For the new Kyma operator the CLI needs to provide utilities to bundle Kyma modules as OCI container images following the [CNUDIE](https://gardener.github.io/component-spec/oci.html#cnudie-transport-format) format.

Those images will be used to release, distribute and deploy Kyma components.

The module creation has to support signing as well.

**Reasons**

Modular approach to the development and distribution of Kyma components as well as adhering to security standards with signing.

**Attachments**

Implementation should be based on:
https://github.com/gardener/component-cli

PR #1308, opened 2022-07-12
create module: mvp

clebs:
**Description**

- Added minimum functioning create module command with the following functionality:
    - Create a component-descriptor.
    - Bundle resources and add them to the descriptor.
    - Push generated bundle to a given repository.
- Added layout for other commands and module pkg.

Work left out of the create module command:
- Signing of images.
- Pulling of images.

**Related issue(s)**
#1301


docs/gen-docs/kyma_alpha_create_module.md
```diff
@@ -0,0 +1,53 @@
...
+```
+
+## Flags
+
+```bash
+      --clean                  Remove the mod-path folder and all its contents at the end.
+  -c, --credentials string     Basic authentication credentials for the given registry in the format user:password
+      --insecure               Use an insecure connection to access the registry.
+      --mod-path string        Specifies the path where the component descriptor and module packaging will be stored. If the path already has a descriptor use the overwrite flag to overwrite it (default "./mod")
+      --name-mapping string    Repository context name mapping. Possible values are: empty, urlPath or sha256-digest
```
jakobmoellersap, line 33: why do we need this customized? I think we should only allow for defualt urlPath mapping as sha256-digest would be created remotely on the oci registry only. What do you think?

pkg/module/oci/client.go
```diff
@@ -0,0 +1,216 @@
...
+	if err != nil {
+		return err
+	}
+
+	pusher, err := resolver.Pusher(ctx, ref)
+	if err != nil {
+		return err
+	}
+
+	if isSingleArchImage(desc.MediaType) {
```
jakobmoellersap, line 109: if its not single archive, wouldn't it error out after the rest?

pkg/module/resources.go
```diff
@@ -0,0 +1,157 @@
...
+}
+
+// ResourceDescriptorList contains a list of all informations to describe a resource.
+type ResourceDescriptorList struct {
+	Resources []ResourceDescriptor `json:"resources"`
+}
+
+// AddResources adds the resources in the given resource definitions into the archive and its FS.
+// A resource definition is a string with format: NAME:TYPE@PATH where NAME and TYPE can be omitted and will default to the last path element name and "helm-chart" respectively
+func AddResources(archive *ctf.ComponentArchive, c *ComponentConfig, log *zap.SugaredLogger, fs vfs.FileSystem, resourceDefs ...string) error {
```
jakobmoellersap, line 35: maybew we should extract the resourceDef type into a custom struct with some validation

jakobmoellersap (commented):
to the 🚀 🌔

docs/gen-docs/kyma_alpha_create_module.md
clebs: Good point, I added this because the reference implementation has it. But it can be left empty and will just do no mapping at all.

jakobmoellersap: I suggest we default to urlPath because of this logic in the current operators based on the ocm spec:
```go
func getOCIRef(repo *ocm.OCIRegistryRepository, descriptor *ocm.ComponentDescriptor, ref string) (*OCI, error) {
	layerRef := OCI{
		Repo: repo.BaseURL,
	}
	switch repo.ComponentNameMapping { //nolint:exhaustive
	case ocm.OCIRegistryURLPathMapping:
		repoSubpath := DefaultRepoSubdirectory
		if ext, found := descriptor.GetLabels().Get(
			fmt.Sprintf("%s%s", ocm.OCIRegistryURLPathMapping, "RepoSubpath")); found {
			repoSubpath = string(ext)
		}
		layerRef.Repo = fmt.Sprintf("%s/%s", repo.BaseURL, repoSubpath)
		layerRef.Name = descriptor.GetName()
		// if ref is not provided, we simply use the version of the descriptor, this will usually default
		// to a component version that is valid
		if ref == "" {
			layerRef.Ref = descriptor.GetVersion()
		} else {
			layerRef.Ref = ref
		}
	default:
		return nil, fmt.Errorf("error while parsing componentNameMapping %s: %w",
			repo.ComponentNameMapping, ErrComponentNameMappingNotSupported)
	}
	return &layerRef, nil
}
```

pkg/module/resources.go
clebs: That is a god idea yes! Will change it.

docs/gen-docs/kyma_alpha_create_module.md
clebs: Right, in that case I will just always use `urlPath` and drop the flag.

pkg/module/oci/client.go
clebs: I have not tested that case TBH, but if it is not single arch it should just push the manifest as is. I don't see a reason why that would error.

docs/gen-docs/kyma_alpha_create_module.md
clebs: All uses of mapping now use `urlPath` and the flag has been removed.

pkg/module/resources.go
clebs: Added a new type for resource definitions and moved all the parsing and setting defaults logic there.

cmd/kyma/alpha/alpha.go
```diff
@@ -0,0 +1,21 @@
...
+	"github.com/kyma-project/cli/internal/cli"
+	"github.com/spf13/cobra"
+)
+
+//NewCmd creates a new kyma CLI command
+func NewCmd(o *cli.Options) *cobra.Command {
+	cmd := &cobra.Command{
+		Use:   "alpha",
+		Short: "Experimental commands",
+		Long: `Alpha commands are experimental unreleased features that should only be used by the Kyma team. Use at your own risk.
```
NHingerl, line 14:
```suggestion
		Long: `Alpha commands are experimental, unreleased features that should only be used by the Kyma team. Use at your own risk.
```

cmd/kyma/alpha/create/module/module.go
```diff
@@ -0,0 +1,137 @@
...
+	}
+
+	cmd := &cobra.Command{
+		Use:   "module MODULE_NAME MODULE_VERSION <CONTENT_PATH> [flags]",
+		Short: "Creates a module bundled as an OCI image from the contents of the given path",
+		Long: `Use this command to create a Kyma module.
+
+### Detailed description
+
+Kyma modules are individual components that can be deplyed into a Kyma runtime. Modules are built and distributed as OCI continer images.
```
NHingerl, line 33:
```suggestion
Kyma modules are individual components that can be deployed into a Kyma runtime. Modules are built and distributed as OCI container images. 
```

cmd/kyma/alpha/create/module/module.go
```diff
@@ -0,0 +1,137 @@
...
+
+	cmd := &cobra.Command{
+		Use:   "module MODULE_NAME MODULE_VERSION <CONTENT_PATH> [flags]",
+		Short: "Creates a module bundled as an OCI image from the contents of the given path",
+		Long: `Use this command to create a Kyma module.
+
+### Detailed description
+
+Kyma modules are individual components that can be deplyed into a Kyma runtime. Modules are built and distributed as OCI continer images. 
+This command provides the means to create such images out of a folder's contents.
```
NHingerl, line 34: Trying to phrase it simpler and shorter.

```suggestion
With this command, you can create such images out of a folder's contents.
```

cmd/kyma/alpha/create/module/module.go
```diff
@@ -0,0 +1,137 @@
...
+
+This command will create a component descriptor in the descriptor path (./mod as a default) and package all the contents on the provided content path as a single layer.
+Optionally, additional layers can be added with contents in other paths.
```
NHingerl, line 37:
```suggestion
Optionally, you can create additional layers with contents in other paths.
```

cmd/kyma/alpha/unpack/module/module.go
```diff
@@ -0,0 +1,41 @@
...
+	"github.com/kyma-project/cli/internal/cli"
+	"github.com/spf13/cobra"
+)
+
+type command struct {
+	opts *Options
+	cli.Command
+}
+
+//NewCmd creates a new kyma CLI command
```
NHingerl, line 13:
```suggestion
//NewCmd creates a new Kyma CLI command
```

cmd/kyma/alpha/unpack/unpack.go
```diff
@@ -0,0 +1,21 @@
+package unpack
+
+import (
+	"github.com/kyma-project/cli/cmd/kyma/alpha/create/module"
+	"github.com/kyma-project/cli/internal/cli"
+	"github.com/spf13/cobra"
+)
+
+//NewCmd creates a new kyma CLI command
```
NHingerl, line 9:
```suggestion
//NewCmd creates a new Kyma CLI command
```

pkg/module/blob/blob.go
```diff
@@ -0,0 +1,315 @@
...
+// MediaTypeTar defines the media type for a tarred file
+const MediaTypeTar = "application/x-tar"
+
+// MediaTypeGZip defines the media type for a gzipped file
+const MediaTypeGZip = "application/gzip"
+
+// MediaTypeOctetStream is the media type for any binary data.
+const MediaTypeOctetStream = "application/octet-stream"
+
+// Output is the output if read BlobInput.
```
NHingerl, line 30: I don't understand this sentence, but that might be due to my lack of background knowledge?

pkg/module/blob/blob.go
```diff
@@ -0,0 +1,315 @@
...
+	DirInputType  = "dir"
+)
+
+// Input defines a local resource input that should be added to the component descriptor and
+// to the resource's access.
+type Input struct {
+	// Type defines the input type of the blob to be added.
+	// Note that a input blob of type "dir" is automatically tarred.
+	Type InputType `json:"type"`
+	// MediaType is the mediatype of the defined file that is also added to the oci layer.
```
NHingerl, line 50:
```suggestion
	// MediaType is the media type of the defined file that is also added to the OCI layer.
```
You used upper case for OCI everywhere else

pkg/module/blob/blob.go
```diff
@@ -0,0 +1,315 @@
...
+	MediaType string `json:"mediaType,omitempty"`
+	// Path is the path that points to the blob to be added.
+	Path string `json:"path"`
+	// CompressWithGzip defines that the blob should be automatically compressed using gzip.
+	CompressWithGzip *bool `json:"compress,omitempty"`
+	// PreserveDir defines that the directory specified in the Path field should be included in the blob.
+	// Only supported for Type dir.
+	PreserveDir bool `json:"preserveDir,omitempty"`
+	// IncludeFiles is a list of shell file name patterns that describe the files that should be included.
+	// If nothing is defined all files are included.
```
NHingerl, line 61:
```suggestion
	// If nothing is defined, all files are included.
```

pkg/module/blob/blob.go
```diff
@@ -0,0 +1,315 @@
...
+	Type InputType `json:"type"`
+	// MediaType is the mediatype of the defined file that is also added to the oci layer.
+	// Should be a custom media type in the form of "application/vnd.<mydomain>.<my description>"
+	MediaType string `json:"mediaType,omitempty"`
+	// Path is the path that points to the blob to be added.
+	Path string `json:"path"`
+	// CompressWithGzip defines that the blob should be automatically compressed using gzip.
+	CompressWithGzip *bool `json:"compress,omitempty"`
+	// PreserveDir defines that the directory specified in the Path field should be included in the blob.
+	// Only supported for Type dir.
```
NHingerl, line 58:
```suggestion
	// Only supported for Type directory.
```
Trying to make it more different from line 62.
`Type dir` vs `type "dir"` confused me a bit :D

pkg/module/blob/blob.go
```diff
@@ -0,0 +1,315 @@
...
+	// IncludeFiles is a list of shell file name patterns that describe the files that should be included.
+	// If nothing is defined all files are included.
+	// Only relevant for blobinput type "dir".
+	IncludeFiles []string `json:"includeFiles,omitempty"`
+	// ExcludeFiles is a list of shell file name patterns that describe the files that should be excluded from the resulting tar.
+	// Excluded files always overwrite included files.
+	// Only relevant for blobinput type "dir".
+	ExcludeFiles []string `json:"excludeFiles,omitempty"`
+	// FollowSymlinks configures to follow and resolve symlinks when a directory is tarred.
+	// This options will include the content of the symlink directly in the tar.
```
NHingerl, line 69:
```suggestion
	// This option will include the content of the symlink directly in the tar.
```

pkg/module/blob/blob.go
```diff
@@ -0,0 +1,315 @@
...
+	// This option should be used with care.
```
NHingerl, line 70:
```suggestion
	// Use this option with care!
```
Why?

pkg/module/blob/blob.go
```diff
@@ -0,0 +1,315 @@
...
+	FollowSymlinks bool `json:"followSymlinks,omitempty"`
+}
+
+// Compress returns if the blob should be compressed using gzip.
```
NHingerl, line 74:
```suggestion
// Compress returns whether the blob should be compressed using gzip.
```

pkg/module/blob/blob.go
```diff
@@ -0,0 +1,315 @@
...
+func (input Input) Compress() bool {
+	if input.CompressWithGzip == nil {
+		return false
+	}
+	return *input.CompressWithGzip
+}
+
+// SetMediaTypeIfNotDefined sets the media type of the input blob if its not defined
```
NHingerl, line 82:
```suggestion
// SetMediaTypeIfNotDefined sets the media type of the input blob if it's not defined.
```

pkg/module/blob/blob.go
```diff
@@ -0,0 +1,315 @@
...
+func (input *Input) Read(ctx context.Context, fs vfs.FileSystem) (*Output, error) {
+	inputInfo, err := fs.Stat(input.Path)
+	if err != nil {
+		return nil, fmt.Errorf("unable to get info for input blob from %q, %w", input.Path, err)
+	}
+
+	// automatically tar the input artifact if it is a directory
+	if input.Type == DirInputType {
+		if !inputInfo.IsDir() {
+			return nil, fmt.Errorf("resource type is dir but a file was provided")
```
NHingerl, line 100:
```suggestion
			return nil, fmt.Errorf("resource type is dir, but you provided a file")
```
Can we switch to active voice?

pkg/module/blob/blob.go
```diff
@@ -0,0 +1,315 @@
...
+		}
+
+		return &Output{
+			Digest: digest.FromBytes(data.Bytes()).String(),
+			Size:   int64(data.Len()),
+			Reader: ioutil.NopCloser(&data),
+		}, nil
+	} else if input.Type == FileInputType {
+		if inputInfo.IsDir() {
+			return nil, fmt.Errorf("resource type is file but a directory was provided")
```
NHingerl, line 138:
```suggestion
			return nil, fmt.Errorf("resource type is file, but you provided a directory")
```

pkg/module/blob/blob.go
```diff
@@ -0,0 +1,315 @@
...
+		return nil, fmt.Errorf("unknown input type %q", input.Path)
+	}
+}
+
+// TarFileSystemOptions describes additional options for tarring a filesystem.
+type TarFileSystemOptions struct {
+	IncludeFiles []string
+	ExcludeFiles []string
+	// PreserveDir defines that the directory specified in the Path field should be included in the blob.
+	// Only supported for Type dir.
```
NHingerl, line 185:
```suggestion
	// Only supported for Type directory.
```

pkg/module/blob/blob.go
```diff
@@ -0,0 +1,315 @@
...
+	PreserveDir    bool
+	FollowSymlinks bool
+
+	root string
+}
+
+// Included determines whether a file should be included.
+func (opts *TarFileSystemOptions) Included(path string) (bool, error) {
+	// if a root path is given remove it rom the path to be checked
```
NHingerl, line 194:
```suggestion
	// If a root path is given, remove it from the path to be checked.
```

pkg/module/blob/blob.go
```diff
@@ -0,0 +1,315 @@
...
+	if len(opts.root) != 0 {
+		path = strings.TrimPrefix(path, opts.root)
+	}
+	// first check if a exclude regex matches
```
NHingerl, line 198:
```suggestion
	// First, check if an exclude regex matches
```
Or is it the "first check"? Then drop my comma :D

pkg/module/build.go
```diff
@@ -0,0 +1,128 @@
...
+	"github.com/gardener/component-spec/bindings-go/codec"
+	"github.com/gardener/component-spec/bindings-go/ctf"
+	"github.com/mandelsoft/vfs/pkg/projectionfs"
+	"github.com/mandelsoft/vfs/pkg/vfs"
+	"sigs.k8s.io/yaml"
+)
+
+// ComponentConfig contains all configurable fields for a component descriptor
+type ComponentConfig struct {
+	ComponentArchivePath string // Location of the component descriptor. If it does not exist it will be created.
```
NHingerl, line 20:
```suggestion
	ComponentArchivePath string // Location of the component descriptor. If it does not exist, it is created.
```

pkg/module/build.go
```diff
@@ -0,0 +1,128 @@
...
+	Name                 string // Name of the module (mandatory)
+	Version              string // Version of the module (mandatory)
+	RegistryURL          string // Registry URL to push the image to (optional)
+	Overwrite            bool   // If true, existing module will be overwritten if the configuration differs.
```
NHingerl, line 24:
```suggestion
	Overwrite            bool   // If true, existing module is overwritten if the configuration differs.
```

pkg/module/build.go
```diff
@@ -0,0 +1,128 @@
...
+	if err := vfs.WriteFile(fs, compDescFilePath, data, os.ModePerm); err != nil {
+		return nil, fmt.Errorf("unable to write component descriptor to %s: %w", compDescFilePath, err)
+	}
+
+	return ctf.NewComponentArchive(cd, archiveFs), nil
+}
+
+func (cfg *ComponentConfig) validate() error {
+	if cfg.Name == "" {
+		return errors.New("The module name can not be empty")
```
NHingerl, line 119:
```suggestion
		return errors.New("The module name cannot be empty")
```

pkg/module/build.go
```diff
@@ -0,0 +1,128 @@
...
+	}
+	if cfg.Version == "" {
+		return errors.New("The module version can not be empty")
```
NHingerl, line 122:
```suggestion
		return errors.New("The module version cannot be empty")
```

pkg/module/def.go
```diff
@@ -0,0 +1,58 @@
...
+	resourceType string
+	path         string
+}
+
+// ResourceDefFromString creates a resource definition from a string with format NAME:TYPE@PATH
+// NAME and TYPE can be omitted and will default to the last path element name and "helm-chart" respectively.
+func ResourceDefFromString(def string) (ResourceDef, error) {
+	items := strings.FieldsFunc(def, func(r rune) bool { return r == ':' || r == '@' }) // split the elements of the format NAME:TYPE@PATH
+	if len(items) == 0 || len(items) == 2 {
+		return ResourceDef{}, fmt.Errorf("the given resource %q could not be parsed. Please make sure it at least contains a path and follows the format NAME:TYPE@PATH", def)
```
NHingerl, line 25:
```suggestion
		return ResourceDef{}, fmt.Errorf("the given resource %q could not be parsed. At least, it must contain a path and follow the format NAME:TYPE@PATH", def)
```

pkg/module/oci/cache.go
```diff
@@ -0,0 +1,59 @@
...
+// Store describes a read only descriptor store
+type Store interface {
+	Get(desc ocispecv1.Descriptor) (io.ReadCloser, error)
+}
+
+type inmemoryCache struct {
+	store map[string][]byte
+}
+
+// NewInMemoryCache creates a new in memory cache.
```
NHingerl, line 29:
```suggestion
// NewInMemoryCache creates a new-in memory cache.
```

pkg/module/oci/client.go
```diff
@@ -0,0 +1,213 @@
...
+	"github.com/opencontainers/go-digest"
+	ocispecv1 "github.com/opencontainers/image-spec/specs-go/v1"
+)
+
+const (
+	DefaultTimeout = 10 * time.Second
+)
+
+type Client interface {
+	// GetManifest returns the ocispec Manifest for a reference
```
NHingerl, line 27:
```suggestion
	// GetManifest returns the ocispec manifest for a reference
```
reference is in lower case in line 30 and 33, so why capitalized here?

pkg/module/remote.go
```diff
@@ -0,0 +1,75 @@
...
+
+// Remote represents remote OCI registry and means to access it
+type Remote struct {
+	Registry    string
+	Credentials string
+	Token       string
+	Insecure    bool
+}
+
+// Push picks up the archive described in the cnfig and pushes it to the provided registry.
```
NHingerl, line 23:
```suggestion
// Push picks up the archive described in the config and pushes it to the provided registry.
```

pkg/module/remote.go
```diff
@@ -0,0 +1,75 @@
...
+	// update repository context
+	if len(r.Registry) != 0 {
+		if err := cdv2.InjectRepositoryContext(archive.ComponentDescriptor, cdv2.NewOCIRegistryRepository(r.Registry, cdv2.OCIRegistryURLPathMapping)); err != nil {
+			return errors.Wrap(err, "unable to add repository context to component descriptor")
+		}
+	}
+
+	manifest, err := cdoci.NewManifestBuilder(ociClient.Cache(), archive).Build(ctx)
+	if err != nil {
+		return errors.Wrap(err, "unable to build oci artifact for component acrchive")
```
NHingerl, line 48:
```suggestion
		return errors.Wrap(err, "unable to build OCI artifact for component archive")
```

pkg/module/resources.go
```diff
@@ -0,0 +1,144 @@
...
+	"sigs.k8s.io/yaml"
+)
+
+// ResourceDescriptor contains all information to describe a resource
+type ResourceDescriptor struct {
+	cdv2.Resource
+	Input *blob.Input `json:"input,omitempty"`
+}
+
+// ResourceDescriptorList contains a list of all informations to describe a resource.
```
NHingerl, line 27:
```suggestion
// ResourceDescriptorList contains a list of all information to describe a resource.
```

pkg/module/resources.go
```diff
@@ -0,0 +1,144 @@
...
+type ResourceDescriptorList struct {
+	Resources []ResourceDescriptor `json:"resources"`
+}
+
+// AddResources adds the resources in the given resource definitions into the archive and its FS.
+// A resource definition is a string with format: NAME:TYPE@PATH where NAME and TYPE can be omitted and will default to the last path element name and "helm-chart" respectively
```
NHingerl, line 33:
```suggestion
// A resource definition is a string with format: NAME:TYPE@PATH, where NAME and TYPE can be omitted and will default to the last path element name and "helm-chart" respectively
```

pkg/module/resources.go
```diff
@@ -0,0 +1,144 @@
...
+		}
+		log.Debugf("Successfully added resource to component descriptor")
+	}
+	log.Debugf("Successfully added all resources to component descriptor")
+	return nil
+}
+
+// generateResources generates resources by parsing the given definitions.
+// Definitions have the following format: NAME:TYPE@PATH
+// If a definition does not have a name or type, the name of the last path element will be used and it will be assumed to be a helm-chart type.
```
NHingerl, line 85:
```suggestion
// If a definition does not have a name or type, the name of the last path element is used and it is assumed to be a helm-chart type.
```

pkg/module/blob/blob.go
clebs: It is true it is not very clear, I will change it to this:
`Output is the output generated when reading a blob.Input.`

NHingerl: Excellent, thank you!

pkg/module/oci/cache.go
NHingerl: Oh dear. Yes. LOL

pkg/module/oci/ref.go
```diff
@@ -0,0 +1,38 @@
...
+
+import (
+	"errors"
+	"fmt"
+
+	cdv2 "github.com/gardener/component-spec/bindings-go/apis/v2"
+	cdoci "github.com/gardener/component-spec/bindings-go/oci"
+)
+
+// Ref wraps the componentspec provided OciRef function by exposing the cdv2.Repository interface
```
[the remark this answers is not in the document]
NHingerl: It depends. I thought it means "something that has been provided by componentspec". 
But now that you ask - what DOES it mean? 😆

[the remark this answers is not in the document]
clebs: Hmm it is true that it is very confusing, this function takes a repository and extracts the registry information to create a complete reference (URL) to the image with `name` and `version`.
For example if the repo has a registry at `my-registry.com` and your image name and version are `compA` and `v0.0.999` you will get `http://my-registry.com/component-descriptors/compA:v0.0.999`.

Maybe a better description could be:
`Ref creates an absolute OCI URL to the component descriptor with name and version at the registry in the repository.`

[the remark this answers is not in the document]
NHingerl: Yes that would be better! :)

NHingerl (approved):

jakobmoellersap (approved):
💯

Issue #1323, opened 2022-07-26
Command: init module


clebs:
**Description**

To shorten the setup time for creating a new Kyma module, an `init module` command has to be implemented. It will initialise a new empty kyma module at the desired location with the given module name.

The command will generate the following contents inside a folder called <module_name>:
```
my-module/
    charts/          // folder containing a set of charts (each in a subfolder)
    crds/             // folder containing all CRDs required by the module
    operator/      // folder containing the operator needed to manage the module
    profiles/        // folder containing all profile settings
    channels/     // folder containing all channel settings
    config.yaml  // yaml file containing installation configuration for any chart in the module that requires custom helm settings
    default.yaml // yaml file containing the default CR needed to start the module installation.
    README.md // Documentation explaining the module format, how it translates to OCI images and how to develop one (can be mostly empty at the beginning)
```

AC:
* An `init` command scaffolds the default folder structure and generates mandatory files to start the development of a new module

**Reasons**

Lower the entry barrier to kyma module development for the kyma team and contributors.

**Attachments**
Follow up issue for #1317

tobiscr:
Would be awesome to extend this `init` command by these additional functionalities:

* A module-build configuration file is generated to allow customers to specify their build parameters in the file. The definition as CLI parameter should be still possible and has precedence over the values defined in the file.
* Customer can define in which programming language he would like to implement the module and the scaffolding also generates config-files for main-stream build-tool  (e.g. Makefile for Go, build.xml for Java+Maven or gradle.build for Java+Gradle etc.)

jakobmoellersap:
I agree, maybe to get everything in but still document and prioritise properly we should probably consider splitting this task into a high-prio and low-prio one depending on what we MUST implement vs what would be nice to have.

clebs:
Yes, this addition would in any case be a follow up issue and not part of the AC for this issue.

tobiscr:
Agree, having the option to define the module build-parameters in a file as a important thing as it simplifies the module build call heavily. Such file can also be committed and shared with other developers.

The option for extending the scaffolding to support programming languages (eg. integration for build-tools etc.) is a nice thing, but not mandatory for the beginning.

clebs:
Removed the crds, profiles and channels folders from the module layout in the description as per #1335

PR #1329, opened 2022-08-01
Add init module command

Tomasz-Smelcerz-SAP:
**Description**

The `init module` command initializes a new empty Kyma module at the desired location with the given module name.

- Add the `init module` command
- Add the resources to be generated according to the AC defined in the related issue

**Related issue(s)**
#1323


pkg/module/initcmd/builder.go
```diff
@@ -0,0 +1,139 @@
...
+	if err != nil {
+		rb.err = fmt.Errorf("An error while reading embedded file %q: %w", embeddedPathName, err)
+		return rb
+	}
+
+	targetFile, err := rb.targetFs.Create(pathName)
+	if err != nil {
+		rb.err = fmt.Errorf("An error while creating target file %q: %w", pathName, err)
+		return rb
+	}
```
clebs, line 85: This piece of code is duplicated in `createFileFromTemplate()`. Might be a god idea to encapsulate it.

clebs (changes requested):

cmd/kyma/alpha/alpha.go
```diff
@@ -2,6 +2,7 @@ package alpha
 
 import (
 	"github.com/kyma-project/cli/cmd/kyma/alpha/create"
+	initcmd "github.com/kyma-project/cli/cmd/kyma/alpha/initcmd"
```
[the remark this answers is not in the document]
Tomasz-Smelcerz-SAP: I will rename the package to "github.com/kyma-project/cli/cmd/kyma/alpha/init" and use alias as initcmd.

pkg/module/initcmd/builder.go
```diff
@@ -0,0 +1,139 @@
+package initcmd
```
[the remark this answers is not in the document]
Tomasz-Smelcerz-SAP: `scaffold` is OK for me. I'll rename it.

cmd/kyma/alpha/initcmd/module/module.go
```diff
@@ -0,0 +1,73 @@
...
+
+	cmd := &cobra.Command{
+		Use:   "module MODULE_NAME <PARENT_DIR>",
+		Short: "Initializes an empty module with the given name in the given parent directory.",
+		Long: `Use this command to create an empty Kyma module with the given name in a correspondingly named subdirectory of provided parent directory.
+
+### Detailed description
+
+Kyma modules are individual components that can be deployed into a Kyma runtime. 
+With this command, you can initialize an empty module folder's for the purpose of further development.
```
NHingerl, line 33: Not sure about the "folder's" but it seems a bit weird. LMK if my suggestion is misunderstanding the meaning:

```suggestion
With this command, you can initialize an empty module folder for the purpose of further development.
```

cmd/kyma/alpha/initcmd/module/module.go
```diff
@@ -0,0 +1,73 @@
...
+
+This command creates a directory with a given name in the target directory.
+In this directory you'll find the template files and subdirectories corresponding to the required module structure:
```
NHingerl, line 36:
```suggestion
In this directory, you'll find the template files and subdirectories corresponding to the required module structure:
```

cmd/kyma/alpha/initcmd/module/module.go
```diff
@@ -0,0 +1,73 @@
...
+    charts/       // folder containing a set of charts (each in a subfolder)
+    crds/         // folder containing all CRDs required by the module
+    operator/     // folder containing the operator needed to manage the module
+    profiles/     // folder containing all profile settings
+    channels/     // folder containing all channel settings
+    config.yaml   // yaml file containing installation configuration for any chart in the module that requires custom helm settings
```
NHingerl, line 42:
```suggestion
    config.yaml   // YAML file containing installation configuration for any chart in the module that requires custom Helm settings
```

cmd/kyma/alpha/initcmd/module/module.go
```diff
@@ -0,0 +1,73 @@
...
+    README.md     // Documentation explaining the module format, how it translates to OCI images and how to develop one (can be mostly empty at the beginning)
```
NHingerl, line 43:
```suggestion
    README.md     // document explaining the module format, how it translates to OCI images, and how to develop one (can be mostly empty at the beginning)
```

pkg/module/initcmd/builder.go
```diff
@@ -0,0 +1,139 @@
...
+
+	_, err = targetFile.Write(data)
+	if err != nil {
+		rb.err = fmt.Errorf("An error while writing data to a target file %q: %w", pathName, err)
+	}
+
+	return rb
+}
+
+// createFileFromTemplate creates a file in the target filesystem using a golang template data from the file with the same name in the embedded filesystem
```
NHingerl, line 96: Not sure what this means. Either drop the "data", or add a word (just guessing: "with", maybe?):

```suggestion
// createFileFromTemplate creates a file in the target filesystem using a Golang template from the file with the same name in the embedded filesystem
```

```suggestion
// createFileFromTemplate creates a file in the target filesystem using a Golang template with data from the file with the same name in the embedded filesystem
```

pkg/module/initcmd/resources/config.yaml
```diff
@@ -0,0 +1,2 @@
+# This is the Helm configuration file for the module "{{ .ModuleName}}"
+# Add here the installation configuration for any chart in the module that requires custom helm settings
```
NHingerl, line 2:
```suggestion
# Add the installation configuration for any chart in the module that requires custom Helm settings.
```

pkg/module/initcmd/resources/default.yaml
```diff
@@ -0,0 +1,2 @@
+# This is the default K8s CustomResource definition file for the module "{{ .ModuleName}}"
```
NHingerl, line 1:
```suggestion
# This is the default Kubernetes CustomResourceDefinition file for the module "{{ .ModuleName}}"
```

pkg/module/initcmd/resources/default.yaml
```diff
@@ -0,0 +1,2 @@
...
+# Add the default CR definition below
```
NHingerl, line 2:
```suggestion
# Add the default CustomResourceDefinition below.
```

pkg/module/initcmd/builder.go
```diff
@@ -0,0 +1,139 @@
...
+	if rb.err != nil {
+		return rb
+	}
+
+	pathName = strings.TrimPrefix(pathName, "/")
+	embeddedPathName := path.Join(embeddedRoot, pathName)
+
+	data, err := embeddedRes.ReadFile(embeddedPathName)
+	if err != nil {
+		rb.err = fmt.Errorf("An error while reading embedded file %q: %w", embeddedPathName, err)
```
NHingerl, line 108: I don't think we need "An" here. Disregard my suggestions in case we use "An" in other documents of this kind; in that case, let's stay consistent :D 

```suggestion
		rb.err = fmt.Errorf("Error while reading embedded file %q: %w", embeddedPathName, err)
```

pkg/module/initcmd/builder.go
```diff
@@ -0,0 +1,139 @@
...
+		return rb
+	}
+
+	t, err := template.New("t").Parse(string(data))
+	if err != nil {
+		rb.err = fmt.Errorf("An error while parsing template from embedded file %q: %w", embeddedPathName, err)
```
NHingerl, line 114:
```suggestion
		rb.err = fmt.Errorf("Error while parsing template from embedded file %q: %w", embeddedPathName, err)
```

pkg/module/initcmd/builder.go
```diff
@@ -0,0 +1,139 @@
...
+		return rb
+	}
+
+	targetFile, err := rb.targetFs.Create(pathName)
+	if err != nil {
+		rb.err = fmt.Errorf("An error while creating target file %q: %w", pathName, err)
```
NHingerl, line 120:
```suggestion
		rb.err = fmt.Errorf("Error while creating target file %q: %w", pathName, err)
```

pkg/module/initcmd/builder.go
```diff
@@ -0,0 +1,139 @@
...
+	}
+	defer targetFile.Close()
+
+	templateOpts := dataTemplateOptions{
+		builderOptions: rb.opts,
+		Path:           pathName,
+	}
+	err = t.Execute(targetFile, templateOpts)
+	if err != nil {
+		rb.err = fmt.Errorf("An error while writing data to a target file %q: %w", pathName, err)
```
NHingerl, line 131:
```suggestion
		rb.err = fmt.Errorf("Error while writing data to a target file %q: %w", pathName, err)
```

pkg/module/initcmd/builder.go
Tomasz-Smelcerz-SAP: Removed as it was causing linter warning (unused function)

pkg/module/scaffold/init.go
```diff
@@ -0,0 +1,84 @@
...
+	"os"
+	"path"
+
+	"github.com/mandelsoft/vfs/pkg/cwdfs"
+	"github.com/mandelsoft/vfs/pkg/vfs"
+)
+
+const directoryMode os.FileMode = 0755
+
+type resourcesToCreate struct{}
```
clebs, line 15: This type is redundant. Just having the unexported function `start(rb resourceBuilder) error` without a receiver would achieve the same result in `createResources()`.

Probably you can then just merge `start()` and `createResources()` into a single function.

cmd/kyma/alpha/init/module/module.go
```diff
@@ -0,0 +1,72 @@
...
+//NewCmd creates a new Kyma CLI command
+func NewCmd(o *Options) *cobra.Command {
+
+	c := command{
+		Command: cli.Command{Options: o.Options},
+		opts:    o,
+	}
+
+	cmd := &cobra.Command{
+		Use:   "module MODULE_NAME <PARENT_DIR>",
```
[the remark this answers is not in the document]
Tomasz-Smelcerz-SAP: Do you think it's more usable both for the user and on the implementation side?
I'd say an explicit module name is better. Let's consider three examples:
- `kyma alpha init module foo` => module foo, assume current dir
- `kyma alpha init module /a/b/foo` => module foo, but it's implicit - user must parse the path with their eyes to see the module name - maybe the path is very long and copied from the clipboard?
- `kyma alpha init module a/b/foo` (no leading slash) => report error?

I really prefer current syntax more, if anyone wants to use current dir, there's a tool for that already:

`kyma alpha init module foo $PWD` 

Maybe we make `<PARENT_DIR>` optional and when it's not provided - we assume current directory? I was thinking about it, but I thought it will be a source of errors for unexperienced users.

pkg/module/scaffold/init.go
Tomasz-Smelcerz-SAP: 👍

cmd/kyma/alpha/init/module/module.go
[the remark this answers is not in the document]
clebs: I don't see the advantage of the 2 args option. For me it is easier to read `/a/b/foo` than `foo /a/b`. Also having 2 args is much more error prone (e.g. users can mix the order and end up having a module called `/a/b` inside a `foo` folder) and needs more validation.

Regarding the `a/b/foo` case assume current dir too and create nested structure (`os.MkdirAll()` handles all 3 cases for you without problem).

And for using pwd, you can do `kyma alpha init module $PWD/foo` as far as I know.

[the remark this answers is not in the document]
Tomasz-Smelcerz-SAP: Well, in my opinion the explicit params are always better than combined or "implicit" ones, because the latter are usually confusing for new users. Experienced users will be fine with both syntaxes.
 
If the module name is explicit, then user have no doubts what is the module name - because it's given explicitly.
If the module name is embedded in the path `/a/b/foo`, then users may think that the entire thing is the module name: `/a/b/foo`. Hierarchical names are very common after all.

Maybe we should approach this issue by looking at the following problem: "How to explain to the user what is the tool  doing?"

In the "two params syntax", users must know the syntax, of course, but it's also the case with every CLI tool that exists.
And we only have to explain that: `kyma alpha init module foo /a/b` creates a directory for the module "foo" as a subdirectory of "/a/b". No other cases.

In the "combined syntax": `kyma alpha init module /a/b/foo`, we have to first tell the users that last path element is the module name - because it's important.
Then we have to document all the cases we discussed above.

That's why I think that it's actually easier to have two parameters, because it's easier to explain how it works. No different cases, no implicit behavior.

Having said that, I don't think it's really a big issue. As I said, it may only be a problem for new users. Experienced ones will learn the convention.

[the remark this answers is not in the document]
Tomasz-Smelcerz-SAP: After thinking about it for a while, I think the best solution would be to remove the path at all from the positional parameters.
- `kyma alpha init module foo` creates the foo directory in the current working dir.
- `kyma alpha init module foo --path=/a/b` creates the foo directory in the given location, that must exist.

No ambiguity, no way to mix the order of parameters, simple to explain/document.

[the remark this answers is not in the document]
clebs: IMHO developers will not be confused by passing a module path and in general I try to keep things simple; the less number of arguments and the more the tool can automatically detect the better.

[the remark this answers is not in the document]
tobiscr: Have not clear favourite for this approach. I would also go to use the plain module-name and use the current directory. If needed, the user can set the `--path` parameter if we want to use a different directory.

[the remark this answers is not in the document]
clebs: I just had a look at the `init function` command and for consistency we could follow the same approach:
https://github.com/kyma-project/cli/blob/0972b8f3088b84bb083fa4a6112eb1b1107334ee/cmd/kyma/init/function/function.go#L50

Only difference is that the `--name` flag would be mandatory and the command would fail if not provided.

Additionally, we could validate that the module name has more than 3 chars and no special characters (except `-`, `_` or `.` ).

@Tomasz-Smelcerz-SAP What do you think about this approach? It would actually be very clse to your initial one but more explicit.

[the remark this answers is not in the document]
Tomasz-Smelcerz-SAP: OK, so the command would look like:
`kyma alpha init module --name=foo --path=/a/b` ? It's fine for me.

cmd/kyma/alpha/init/module/module.go
```diff
@@ -0,0 +1,94 @@
...
+The name of the  module directory is the same as the name of the module.
+Module name must start with a letter and may only consist of alphanumeric and '.', '_', or '-' characters.
+You may optionally provide a "--path" flag to create module directory at some different location.
+In the module directory you'll find the template files and subdirectories corresponding to the required module structure:
+    charts/       // folder containing a set of charts (each in a subfolder)
+    crds/         // folder containing all CRDs required by the module
+    operator/     // folder containing the operator needed to manage the module
+    profiles/     // folder containing all profile settings
+    channels/     // folder containing all channel settings
+    config.yaml   // YAML file containing installation configuration for any chart in the module that requires custom Helm settings
```
NHingerl, line 46:
```suggestion
    config.yaml   // YAML file containing the installation configuration for any chart in the module that requires custom Helm settings
```

pkg/module/scaffold/resources/default.yaml
```diff
@@ -0,0 +1,2 @@
+# This is the default Kubernetes CustomResourceDefinition file for the module "{{ .ModuleName}}"
```
clebs, line 1: This file has to contain a default instance of the module Custom Resource, not the Custom Resource Definition.

cmd/kyma/alpha/create/module/module.go
```diff
@@ -61,6 +61,11 @@ func (c *command) Run(args []string) error {
 		cli.AlphaWarn()
 	}
 
+	_, err := module.ValidateModuleName(args[0])
```
clebs, line 64: This kind of stutters, I would suggest calling it `module.ValidateName()`. Package name always gives context and has to be taken into consideration when naming types and functions.

Tomasz-Smelcerz-SAP: 👍

clebs (approved):

NHingerl (approved):

PR #1346, opened 2022-08-18
Fix sec vuln rel 2.6

khlifi411:
**Description**

1. Update `go.mod` version from `1.17`to `1.18`
2. Update [golang.org/x/crypto](https://pkg.go.dev/golang.org/x/crypto) to latest
3. Updated `k8s.io/client-go` to latest
4. Update `golang.org/x/net` to latest
5. Update `github.com/prometheus/client_golang` to latest
6. Update `github.com/containerd/containerd` to latest
7. Remove unnecessary `replace` directives from `go.mod` file.


Makefile
```diff
@@ -14,7 +14,7 @@ FLAGS = -ldflags '-s -w -X github.com/kyma-project/cli/cmd/kyma/version.Version=
 
 .PHONY: resolve
 resolve:
-	go mod tidy -compat=1.17
+	go mod tidy -compat=1.18
```
jakobmoellersap, line 17: this will break with importing the reconciler, you will need to use compat 1.17 I think

```
The -compat flag preserves any additional checksums needed for the
'go' command from the indicated major Go release to successfully load
the module graph, and causes tidy to error out if that version of the
'go' command would load any imported package from a different module
version. By default, tidy acts as if the -compat flag were set to the
version prior to the one indicated by the 'go' directive in the go.mod
file.
```

Considering this, we can only increment if reconciler is also pushed to 1.18 in compat (I think this will break though)

khlifi411: I have just tried it with latest reconciler, and it worked 😅

adityabhatia (approved):

Issue #1365, opened 2022-08-31
create module: CRD and default CR validation


clebs:
**Description**

Kyma modules are represented at runtime with a specific `CRD` which needs to conform to the `lifecycle-manager`'s specification. 
To ensure a valid `CRD` is defined for a module it needs to be validated.

Additionally, to instantiate a module, developers have to provide a default CR instance for the module they want to create a module template. Currently the CR is not validated before building the module and could later fail when deploying it to a cluster.

To make module development safer we have to validate the CRD and the default CR against the CRD usually found in the module's operator.

**AC**
- Validate that the CRD defined in the module is correct as per `lifecycle-manager`.
- Validate that the default CR provided for the module template is correct.

**Reasons**

Avoid building defective modules and templates.

Tomasz-Smelcerz-SAP:
The `kubectl` offers client-side validation of the CR (vs CRD schema), but it reads the CRD OpenAPI v3 schema from the cluster. Perhaps we could reuse that code, pointing to a definition from a local file instead. We can't query the cluster for the CRD, because, at the development time, there may be no cluster at all. In addition `kubectl` codebase contains all the k8s ecosystem patches already. We could use a generic OpenAPI v3 schema validator, but then we risk that some custom validations are not implemented there. Besides that, there isn't a lot tools for OpenAPI v3 schema validations. I was able to find [this](https://openapi.tools/#data-validators) list, just one generic project written in golang.

jakobmoellersap:
After looking at this again:

1. The CR cannot be safely verified against an OpenAPI scheme unless we do client side validation against the CRD bundled in the module
2. To find the correct CRD to verify against we have to iterate over all CRDs bundled in the module and compare it through the GVK against one specified in default.yaml
3. In addition to normal YAML validation against the CR I think the more important part of validation would be to validate against `status.state`, maybe we should prioritize this first because the webhook in `lifecycle-manager` already contains such a validation

Tomasz-Smelcerz-SAP:
@jakobmoellersap status object shouldn't be set by end users, but by the controller only, isn't it? The scope of the validation, as I understand it, is to detect errors in user's input, i.e. CR `spec` object. I'd say we should maybe validate that there's no status at all in the object's definition YAML. The proper status will be set/created once the CR is reconciled by the controller in the target cluster.

clebs:
Hi @Tomasz-Smelcerz-SAP, @jakobmoellersap,

[...]

PR #1379, opened 2022-09-19
Feature/cr validation

Tomasz-Smelcerz-SAP:
**Description**
Implements: https://github.com/kyma-project/cli/issues/1365

Extended the `kyma alpha create module` command with:
- Skipping the validation if there's no `default.yaml` file
- Installation of envtest binaries in the `~/.kyma` directory
- Finding the file containing the CRD matching the CR defined in the `default.yaml` file (by "group" and "kind")
- Bootstraping the envtest with the proper CRD file
- Creating the object defined in the `default.yaml` (CR)  in the envtest environment
- Reporting the errors, if any
- Added a CLI flag to explicitly skip the validation

**Related issue(s)**
https://github.com/kyma-project/cli/issues/1365


Tomasz-Smelcerz-SAP:
Testing:
In the [template-operator](https://github.com/kyma-project/lifecycle-manager)/samples/template-operator directory: 
```
kyma alpha create module kyma.project.io/module/foobar 0.0.4 $PWD --overwrite
```

pkg/module/validation.go
```diff
@@ -3,11 +3,100 @@ package module
...
+	if err != nil {
+		return false, err
+	}
+	if !exists {
+		//If there's no "default.yaml" file, skip the validation.
+		return true, nil
+	}
+
+	//TODO: How to get this directory path programatically? What's the contract?
+	crdDirectoryPath := os.Getenv("CRD_DIRECTORY")
```
jakobmoellersap, line 57: IMO we should make this check smarter and discover the CRD based on the packaged layers instead of injecting it as an environment variable. Otherwise one has to always define the directory, otherwise validation won't work.

Tomasz-Smelcerz-SAP: @jakobmoellersap Thanks for the idea. I thought about source files scanning. But this could also work. There's one problem though. The validation is meant to "fail fast", without even building the module. But maybe we can "delay" it, actually building the module is not that much time/resource consuming.

internal/kube/client.go
```diff
@@ -51,6 +51,35 @@ func NewFromConfig(url, file string) (KymaKube, error) {
...
+	istioClient, err := istio.NewForConfig(config)
+	if err != nil {
+		return nil, err
+	}
+
+	return &client{
+			static:  sClient,
+			dynamic: dClient,
+			istio:   istioClient,
+			restCfg: config,
```
Tomasz-Smelcerz-SAP, line 78: For the kubeconfig field try to return not a zero value here (nil), but rather try to construct an empty/dummy kubeconfig and return that.

pkg/module/validation.go
```diff
@@ -3,11 +3,100 @@ package module
...
+	exists, crData, err := readDefaultCR(modulePath)
+	if err != nil {
+		return false, err
+	}
+	if !exists {
+		//If there's no "default.yaml" file, skip the validation.
+		return true, nil
+	}
+
+	//TODO: How to get this directory path programatically? What's the contract?
```
Tomasz-Smelcerz-SAP, line 56: The way to go is to traverse all the files under the `charts` directory of the module path, and find the one that matches GVK of the given default CR.

pkg/module/validation.go
```diff
@@ -3,11 +3,100 @@ package module
...
+		if stopErr != nil {
+			//TODO: This doesn't seem to print anything...
+			log.Error(fmt.Errorf("Error during stopping envTest: %w", stopErr))
+			//THIS does: fmt.Println(fmt.Errorf("Error during stopping envTest: %w", stopErr))
+		}
+	}()
+
+	kc, err := kube.NewFromRestConfigWithTimeout(restCfg, 30*time.Second)
+
+	err = kc.Apply(crData)
```
Tomasz-Smelcerz-SAP, line 75: We need to find the way to set a `default` namespace if the namespace is not provided or it's set to a  different value, because the envtest doesn't provide any "defaulting" for namespace.

Tomasz-Smelcerz-SAP: This is done, unit test provided.

pkg/module/validation.go
```diff
@@ -3,11 +3,150 @@ package module
...
+
+	crData, err := os.ReadFile(crPath)
+	if err != nil {
+		return false, nil, fmt.Errorf("Error reading the default CR file `%q`: %w", crPath, err)
+	}
+
+	return true, crData, nil
+}
+
+//ensureDefaultNamespace parses crData and ensures that metadata.namespace is set to the value "default". This is because of how we use the envtest to validate the CR.
```
Tomasz-Smelcerz-SAP, line 46:
```suggestion
//ensureDefaultNamespace parses crData and ensures that the metadata.namespace attibute exists and it's value is "default". This is because of how we use the envtest to validate the CR.
```

pkg/module/validation.go
Tomasz-Smelcerz-SAP: Turned out to be too fragile, YAML parsing was necessary.

Tomasz-Smelcerz-SAP:
Testing:
In the [lifecycle-manager](https://github.com/kyma-project/lifecycle-manager)/samples/template-operator directory run: 
```
kyma alpha create module kyma.project.io/module/foobar 0.0.4 $PWD --overwrite
```

It now automatically finds the correct CRD (if it's there). Namespace issues (no namespace attribute in the "default.yaml") are also handled.

Tomasz-Smelcerz-SAP:
Note for reviewer: There's a problem with step presentation. On a Linux machine, I get this:
```
  Installed envtest binaries in version 1.24.2, path: "/home/myuser/.kyma/k8s/1.24.2-linux-amd64"
- using envtest from "/home/myuser/.kyma/k8s/1.24.2-linux-amd64"
- Default CR validation succeeded
- Creating module archive at "./mod"
- Resources added
```
Note the "- Default CR validation succeeded" line.

On OSX I get:
```

```

cmd/kyma/alpha/create/module/opts.go
```diff
@@ -8,16 +8,17 @@ import (
...
+	RegistryURL       string
+	Credentials       string
+	TemplateOutput    string
+	Channel           string
+	Token             string
+	Insecure          bool
+	ResourcePaths     []string
+	Overwrite         bool
+	Clean             bool
+	ValidateDefaultCR bool
```
clebs, line 21: Do we really need this field? I can't think of a use case where we want to allow unvalidated CRs into the image.

pkg/module/validation.go
```diff
@@ -3,11 +3,348 @@ package module
...
+	//TODO: Do we need to override the name or it's always "default.yaml"?
+	crPath := filepath.Join(modulePath, defaultCRName)
+
+	fileInfo, err := os.Stat(crPath)
+	if errors.Is(err, os.ErrNotExist) {
+		return false, nil, nil
+	}
+	if err != nil {
+		return false, nil, fmt.Errorf("Error accessing the default CR file `%q`: %w", crPath, err)
+	}
```
clebs, line 96: You only need to check the kind of error if there is an error. This way is more readable as well:
```suggestion
	if err != nil {
		if errors.Is(err, os.ErrNotExist) {
			return false, nil, nil
		}
		return false, nil, fmt.Errorf("Error accessing the default CR file `%q`: %w", crPath, err)
	}
```

pkg/module/validation.go
```diff
@@ -3,11 +3,348 @@ package module
...
+
+	if !fileInfo.Mode().IsRegular() {
+		return false, nil, fmt.Errorf("Error reading the default CR file `%q`: Not a regular file", crPath)
+	}
+
+	crData, err := os.ReadFile(crPath)
```
clebs, line 102: Doesn't `os.ReadFile()` already do all the checks done above? I am quite certain that you can omit all lines prior to this in the function and check if file exists or not after calling `os.ReadFile()`.

pkg/module/validation.go
```diff
@@ -3,11 +3,348 @@ package module
...
+	envTest, restCfg, err := startValidationEnv(crdFilePath, envtestBinariesPath)
+	if err != nil {
+		return false, err
+	}
+
+	defer func() {
+		stopErr := envTest.Stop()
+		if stopErr != nil {
+			//TODO: This doesn't seem to print anything...
+			log.Error(fmt.Errorf("Error during stopping envTest: %w", stopErr))
```
clebs, line 71:
```suggestion
			log.Error(fmt.Errorf("Error stopping envTest: %w", stopErr))
```

pkg/module/validation.go
```diff
@@ -3,11 +3,348 @@ package module
...
+			//THIS does: fmt.Println(fmt.Errorf("Error during stopping envTest: %w", stopErr))
+		}
+	}()
+
+	kc, err := kube.NewFromRestConfigWithTimeout(restCfg, 30*time.Second)
+
+	err = kc.Apply(properCR)
+	if err != nil {
+		return false, fmt.Errorf("Error during applying the default CR: %w", err)
```
clebs, line 80:
```suggestion
		return false, fmt.Errorf("Error applying the default CR: %w", err)
```

pkg/module/validation.go
```diff
@@ -3,11 +3,348 @@ package module
...
+	return true, crData, nil
+}
+
+// ensureDefaultNamespace ensures that the metadata.namespace attribute exists and it's value is "default". This is because of how we use the envtest to validate the CR.
+func ensureDefaultNamespace(modelMap map[string]interface{}) error {
+
+	//Traverse the Map to look for "metadata.namespace"
+	metadataMap, err := mustReadMap(modelMap, "metadata")
+	if err != nil {
+		return fmt.Errorf("Error during parsing default CR: %w", err)
```
clebs, line 116:
```suggestion
		return fmt.Errorf("Error parsing default CR: %w", err)
```

pkg/module/validation.go
```diff
@@ -3,11 +3,348 @@ package module
...
+	}
+
+	return asString, nil
+}
+
+func parseYamlToMap(crData []byte) (map[string]interface{}, error) {
+	modelMap := make(map[string]interface{})
+	err := yaml.Unmarshal([]byte(crData), &modelMap)
+	if err != nil {
+		return nil, fmt.Errorf("Error during parsing default CR: %w", err)
```
clebs, line 188:
```suggestion
		return nil, fmt.Errorf("Error parsing default CR: %w", err)
```

pkg/module/validation.go
```diff
@@ -3,11 +3,348 @@ package module
...
+	}
+
+	return modelMap, nil
+}
+
+func renderYamlFromMap(modelMap map[string]interface{}) ([]byte, error) {
+
+	output, err := yaml.Marshal(modelMap)
+	if err != nil {
+		return nil, fmt.Errorf("Error during processing default CR data: %w", err)
```
clebs, line 198:
```suggestion
		return nil, fmt.Errorf("Error processing default CR data: %w", err)
```

pkg/module/validation.go
```diff
@@ -3,11 +3,348 @@ package module
...
+		return false, err
+	}
+	if !exists {
+		//If there's no "default.yaml" file, skip the validation.
+		return true, nil
+	}
+
+	crMap, err := parseYamlToMap(crData)
+	if err != nil {
+		return false, fmt.Errorf("Error during parsing default CR: %w", err)
```
clebs, line 35:
```suggestion
		return false, fmt.Errorf("Error parsing default CR: %w", err)
```

pkg/module/validation.go
```diff
@@ -3,11 +3,348 @@ package module
...
+	err = kc.Apply(properCR)
+	if err != nil {
+		return false, fmt.Errorf("Error during applying the default CR: %w", err)
+	}
+	return false, nil
+}
+
+// readDefaultCR reads the default CR file's contents. The returned bool is true if the data is successfully read, it's false otherwise.
+func readDefaultCR(modulePath string) (bool, []byte, error) {
+	//TODO: Do we need to override the name or it's always "default.yaml"?
```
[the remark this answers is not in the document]
Tomasz-Smelcerz-SAP: :+1:

cmd/kyma/alpha/create/module/opts.go
Tomasz-Smelcerz-SAP: This validation takes around 5 seconds on average (depends on the machine). I think people will want to skip it in most cases during development, because waiting 5 seconds each time even if I don't change anything in CR/CRD is painful.
Maybe there's another way? We don't mind if the local image is not validated - after all it's on the local machine only. We don't want it to be published. Maybe the validation should be only required when publishing the module? The next [issue](https://github.com/kyma-project/lifecycle-manager/issues/226) will integrate pushing the module with the `kyma alpha create module` command. Perhaps we can solve this problem in that other issue.

pkg/module/validation.go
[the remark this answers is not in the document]
Tomasz-Smelcerz-SAP: Because it's something we "may need" in the future, for now I'll just add an environment variable for that.
It will be not visible as a CLI option and it will not be documented. Since it's an alpha command, I think it's OK. We'll be able to test the scenario you described without introducing any additional flags (contract) to the command.
Later on, if required, we'll add an official flag for it.

pkg/module/validation.go
Tomasz-Smelcerz-SAP: Good catch! Fixed.

cmd/kyma/alpha/create/module/opts.go
clebs: Good point yes! In that case I would say it makes sense to validate CR always when pushing the image.

Pushing is already implemented and working, see:
https://github.com/kyma-project/cli/blob/main/cmd/kyma/alpha/create/module/module.go#L132

Tomasz-Smelcerz-SAP: Done.

cmd/kyma/alpha/create/module/module.go
```diff
@@ -81,34 +83,14 @@ func (cmd *command) Run(args []string) error {
 
 	l := cli.NewLogger(cmd.opts.Verbose).Sugar()
 
+	defaultCRValidated := false
```
clebs, line 86: This boolean is redundant and calling `validateDefaultCR()` in 2 places as well.
 You can just do it once as an individual step in line 141:
 ```
 if cmd.opts.ValidateDefaultCR || cmd.opts.RegistryURL != "" {
    validateStep := cmd.NewStep("Validating resources...")

    // do validation checks here

    validateStep.Success()
} 
```

Comment (Tomasz-Smelcerz-SAP) [reply cmd/kyma/alpha/create/module/module.go:86]: @clebs OK, let me think about it.

File: cmd/kyma/alpha/create/module/module.go [hunk 17]
```diff
@@ -56,12 +62,13 @@ Finally, if a registry is provided, the created module is pushed.
 	cmd.Flags().BoolVarP(&o.Overwrite, "overwrite", "w", false, "overwrites the existing mod-path directory if it exists")
 	cmd.Flags().BoolVar(&o.Insecure, "insecure", false, "Use an insecure connection to access the registry.")
 	cmd.Flags().BoolVar(&o.Clean, "clean", false, "Remove the mod-path folder and all its contents at the end.")
+	cmd.Flags().BoolVar(&o.ValidateDefaultCR, "validateCR", false, "Validate the custom resource defined in the \"default.yaml\" file on demand. This validation always runs when pushing the module.")
```
NHingerl, line 65: mhm, wait. Line 47 says that validation happens _before_ pushing; but this line says it is always triggered by the pushing, so it happens after pushing starts (and probably aborts the pushing if validation is unsuccessful?).

cmd/kyma/alpha/create/module/module.go
```diff
@@ -39,6 +44,7 @@ This command creates a component descriptor in the descriptor path (./mod as a d
 Optionally, you can create additional layers with contents in other paths.
 
 Finally, if a registry is provided, the created module is pushed.
```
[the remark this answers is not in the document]
Tomasz-Smelcerz-SAP: @NHingerl the point is, we have a validation that is always performed when the module is pushed - it's mandatory. For now, it's the "default CR" validation. Maybe it will be extended in the future, but for now it's that one.

clebs (approved):

cmd/kyma/alpha/create/module/module.go
Tomasz-Smelcerz-SAP: Validation - if triggered - always executes early in the command, to "fail fast".
Validation is triggered in two cases:
 - user asked for it by setting the `validateCR` flag (on-demand validation)
 - user wants to push the artifact to the registry by configuring `--registry=https://some.registry.com` flag
 
 So if the command is configured to push the artifact - which is at the end of the command, the validation will be always triggered at the  beginning of the command (to fail fast in case of errors).
 Now we have to find a way to describe this in a nice and short way :)

cmd/kyma/alpha/create/module/module.go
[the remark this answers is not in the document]
NHingerl:
```suggestion
Finally, if you provided a registry to which to push the artifact, the created module is pushed and validated. For example, the default CR defined in the \"default.yaml\" file is validated against CustomResourceDefinition.
```

cmd/kyma/alpha/create/module/module.go
NHingerl: Thank you for the great explanation, this helps! 

I'll try to document that information in the `Detailed description`, see my proposal here: https://github.com/kyma-project/cli/pull/1379/files#r992524816

clebs (dismissed):

cmd/kyma/alpha/create/module/module.go
```diff
@@ -38,7 +43,9 @@ With this command, you can create such images out of a folder's contents.
 This command creates a component descriptor in the descriptor path (./mod as a default) and packages all the contents on the provided content path as a single layer.
 Optionally, you can create additional layers with contents in other paths.
 
-Finally, if a registry is provided, the created module is pushed.
+Finally, if you provided a registry to which to push the artifact, the created module is pushed and validated. For example, the default CR defined in the \"default.yaml\" file is validated against CustomResourceDefinition.
```
Tomasz-Smelcerz-SAP, line 46:
```suggestion
Finally, if you provided a registry to which to push the artifact, the created module is validated and pushed. For example, the default CR defined in the \"default.yaml\" file is validated against CustomResourceDefinition.
```

NHingerl (approved):

clebs (approved):

Issue #1373, opened 2022-09-12
Provision k3d with cni enabled by default


strekm:
**Description**

For Kyma 2.8 Istio will be bumped to Istio 1.15 and Istio cni plugin will be used instead of Istio init-containers. Because of that for k3d provisioning additional setup needs to be done: enabling cni plugin via kubelet params. 

ACs:
- [x] cni is enabled for k3d cluster
- [x] kyma 2.6 installs / upgrades (no Istio cni is used)
- [x] kyma with Istio using cni plugin installs / upgrades

**Reasons**

Security improvements

**Attachments**

part of: https://github.com/kyma-project/kyma/issues/15347

**PR with the change**
- https://github.com/kyma-project/cli/pull/1383

barchw:
Cli PR: https://github.com/kyma-project/cli/pull/1383

PR #1383, opened 2022-09-21
Add value for istio specifying k3d as deployment

barchw:
**Description**

- Add k3d parameter used when deploying Kyma on k3d

**Related issues**

- https://github.com/kyma-project/cli/issues/1373

### DIFF
--- a/internal/deploy/values/values.go
+++ b/internal/deploy/values/values.go
@@ -60,7 +60,7 @@ func addClusterSpecificDefaults(builder *builder, clusterInfo clusterinfo.Info) 
 			addDefaultGlobalDomainName(defaultLocalKymaDomain).
 			addDefaultGlobalTLSCrtAndKey(defaultLocalTLSCrtEnc, defaultLocalTLSKeyEnc).
 			addDefaultServerlessKanikoForce().
-			addDefaultk3dFlagForIstio()
+			addk3dFlagForIstio()
 	} else if gardener, isGardener := clusterInfo.(clusterinfo.Gardener); isGardener {
 		builder.addDefaultGlobalDomainName(gardener.Domain)
 	}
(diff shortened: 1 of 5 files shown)


clebs (dismissed):
This does not need a flag. The CLI is capable of detecting if a cluster is k3d.
Please see:
https://github.com/kyma-project/cli/blob/7a600e8151b5fd23ff1be0f79c31d8259999714d/internal/clusterinfo/k3d.go#L17

internal/deploy/values/values.go
```diff
@@ -59,7 +59,8 @@ func addClusterSpecificDefaults(builder *builder, clusterInfo clusterinfo.Info)
 			addDefaultServerlessRegistryConfig(defaultRegistryConfig).
 			addDefaultGlobalDomainName(defaultLocalKymaDomain).
 			addDefaultGlobalTLSCrtAndKey(defaultLocalTLSCrtEnc, defaultLocalTLSKeyEnc).
-			addDefaultServerlessKanikoForce()
+			addDefaultServerlessKanikoForce().
+			addDefaultk3dValuesForIstio()
```
clebs, line 63: This should be added only if the target cluster is K3d.

clebs (changes requested):

internal/deploy/values/values.go
barchw:
```go
if k3d, isK3d := clusterInfo.(clusterinfo.K3d); isK3d {
```
This is not the correct place to have this param added?

clebs: Ah yes, sorry... I missed the fact that this call is inside a block that only runs if the cluster is k3d.

clebs (approved):

PR #1392, opened 2022-09-28
Add GCloud authentication

Halamix2:
**Description**

- Add GCloud authentication to fix pushing artifacts

**Related issue(s)**

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -15,7 +15,7 @@ FLAGS = -ldflags '-s -w -X github.com/kyma-project/cli/cmd/kyma/version.Version=
 .PHONY: gcp-authenticate
 gcp-authenticate:
 	@echo "Authenticating to gcloud"
-    @gcloud auth activate-service-account --key-file "$GOOGLE_APPLICATION_CREDENTIALS" || exit 1
+	@gcloud auth activate-service-account --key-file "$GOOGLE_APPLICATION_CREDENTIALS" || exit 1
 
 .PHONY: resolve
 resolve:
--- a/Makefile
+++ b/Makefile
@@ -12,7 +12,7 @@ endif
 
 FLAGS = -ldflags '-s -w -X github.com/kyma-project/cli/cmd/kyma/version.Version=$(VERSION)'
 
-.PHONE: gcp-authenticate
+.PHONY: gcp-authenticate
 gcp-authenticate:
 	@echo "Authenticating to gcloud"
     @gcloud auth activate-service-account --key-file "$GOOGLE_APPLICATION_CREDENTIALS" || exit 1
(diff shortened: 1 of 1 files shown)


Makefile
```diff
@@ -12,6 +12,11 @@ endif
 
 FLAGS = -ldflags '-s -w -X github.com/kyma-project/cli/cmd/kyma/version.Version=$(VERSION)'
 
+.PHONY: gcp-authenticate
```
clebs, line 15: I am not sure the CLI should have any attachment or notion of `gcloud`. Shouldn't this be part of the CLI jobs and scripts in Prow?

Ressetkk: @clebs stable release of CLI does not use any of the scripts on the test-infra. We call `make ci-main` directly in the ProwJob. If this job needs to do some stuff on GCP, then it won't be authenticated without invoking the command.

clebs: Yes I know, but that does not mean that this belongs in the `Makefile` just because the job calls `make` directly.
It looks to me that the job needs to leverage the scripts and utilities we already have in Prow.

Ressetkk: Yes I agree, but we also aim that build jobs **do not** have hard requirement to the test-infra repository. That means you have to include some of the required commands somewhere and the best candidate is separate make step that authenticates to GCP with already included credentials.

It's a short-term solution until we manage to migrate to GKE workload identity, so we can map service accounts with GCP SAs. Then we won't need any of those things.

clebs: As far as I know almost every job currently is using a script in test-infra, is there a full migration of logic out f the repository planned?
Also, what is the benefit on not requiring the test-infra repo (other than not having to clone it)?

Maybe the better approach would be to include all there CI tools in the images prow uses?

Ressetkk: cloningxevery job with test-infra makes problem with maintainability. A long-term plan is to eventually move away from that, so all required scripts are included either in image, or target repository, because you could re-use jobs across different branches on the same repo. This also doesn't require making cut-offs on test-infra side and creating even more copy-pasted jobs.

The best approach for current situation is to include scripts required in CLI inside this repo, so they can be used without relying on external repos. This PR solves just this corner-case where job `stable-kyma-cli` requires GCP auth to use `gsutil`. That's it.

clebs: Then we first need to decide which one we are going for, either image contains CI/CD logic or target repos do.

Ressetkk: I think it's something we can work on in future improvements in test infra. Right now I think this is the best and least invasive solution, since removing build-generic.sh usage also overlaps with SLC-29 certification.

clebs: As a quick and temporary hack this is acceptable, but it should be kept at a minimum and removed as soon as possible.

clebs (approved):

Issue #1360, opened 2022-08-29
Show warning before CLI is changing a resource on SKR cluster


tobiscr:
**Description**

Customers are often going to use the Kyma CLI to modify an SKR cluster. This is a very dangerous step as the CLI is not aware about the used overrides in SKR. The risk is very high that such a change will break the Kyma runtime.

The CLI has to detec whether it's going to change an SKR Kyma runtime, show a warning message and ask the customer whether he is aware about the risk and whether he wants to proceed.

AC:
* Check if the Kyma CLI is connected to an SKR cluster
* If the user calls the `deploy`, `undeploy` command, show a warning and ask wether he wants to proceed:
     ```
     CAUTION: You are trying to use Kyma CLI to change a managed Kyma runtime (SAP Kyma Runtime). This action may corrupt the Kyma runtime. Proceed at your own risk.

     Do you really want to proceed? [N/y]
     ```
* The customer has to decide wether he wants to proceed. Per default (e.g.  customer just presses `enter`), the Kyma CLI will not modify the SKR installation.

**Reasons**

**Attachments**

NHingerl:
Instead of this:

> You are going to change a managed Kyma runtime (SAP SKR offering). Applying changes on a managed installation by using the Kyma CLI can lead to a corrupted Kyma runtime and is on your own risk.
> 
> Do you really want to proceed? [N/y]
> 

How about this:

> CAUTION: You are trying to use Kyma CLI to change a managed Kyma runtime (SAP Kyma Runtime). This action may corrupt your Kyma runtime. Proceed at your own risk.
> 
> Do you really want to proceed? [N/y]

Tomasz-Smelcerz-SAP:
A question about detecting, if "we are connected to the SKR cluster".
How shoul we check that?
We could - for example - test if the `kyma-system` namespace exists.
Or perhaps w can re-use the `kyma version` detection logic?

clebs:
Hi @Tomasz-Smelcerz-SAP,

that is exactly the challenge in this task. We need to find a way to reliably know if the cluster we are deploying to is on SKR.

The approaches you mention are unfortunately not going to work because a local dev cluster (any cluster with kyma on it actually) will have a `kyma-system` namespace and also a version that is indistinguishable from an SKR.

Tomasz-Smelcerz-SAP:
@NHingerl @tobiscr FYI: I am taking Nina's version of warning message :)

PR #1425, opened 2022-10-25
Feature/managed cluster deploy warning

Tomasz-Smelcerz-SAP:
**Description**

Show warning before CLI is executing `kyma deploy` or `kyma undeploy` on an SKR cluster

- Add prompt to `kyma deploy` command if on SKR
- Add prompt to `kyma undeploy` command if on SKR
- Add SKR detection method based on K8s API Server URL

**Related issue(s)**
#1360

### DIFF
--- a/internal/cli/managed.go
+++ b/internal/cli/managed.go
@@ -15,7 +15,7 @@ func DetectManagedEnvironment(cmd *Command) error {
 	if clusterinfo.IsManagedKyma(cmd.K8s.RestConfig()) {
 		detectStep.LogWarn(managedKymaWarning)
 		if !detectStep.PromptYesNo("Do you really want to proceed? ") {
-			return errors.New("Deploy command stopped by user")
+			return errors.New("Command stopped by user")
 		}
 	}
 	detectStep.Success()
--- a/internal/cli/managed.go
+++ b/internal/cli/managed.go
@@ -1,0 +1,23 @@
+package cli
+
+import (
+	"errors"
+
+	"github.com/kyma-project/cli/internal/clusterinfo"
+)
+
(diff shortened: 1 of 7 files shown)


internal/cli/managed.go
```diff
@@ -0,0 +1,23 @@
...
+	"errors"
+
+	"github.com/kyma-project/cli/internal/clusterinfo"
+)
+
+const managedKymaWarning = "CAUTION: You are trying to use Kyma CLI to change a managed Kyma runtime (SAP Kyma Runtime). This action may corrupt the Kyma runtime. Proceed at your own risk."
+
+// DetectManagedEnvironment introduces a step that checks if the target runtime is a managed Kyma runtime. CLI should be used with caution in such environment and user is prompted for confirmation.
+func DetectManagedEnvironment(cmd *Command) error {
+	detectStep := cmd.NewStep("Detecting managed Kyma runtime")
```
clebs, line 13: IMHO not everything that happens on the process needs to be added as a step. Steps should be well thought out and provide useful feedback to the user. This will in most cases pollute the output and in the cases it is relevant, the prompt already provides all the information.

With that change, it is enough to provide the current step to the function and not the whole command.

Tomasz-Smelcerz-SAP: The problem is, we don't have any "current step" at the time this function is invoked... I'll take a look at what we can do.

clebs: I see, in that case we can pass an empty step and only use it to prompt.

Tomasz-Smelcerz-SAP: 👍

clebs (approved):

Issue #1454, opened 2022-11-22
k3d coredns config disallows hostname lookup via http in local setup


kwiatekus:
**Description**

Provide a necessary coredns setup that allows to lookup local k3d registry (that is created by kyma provision command ) under `k3d-kyma-registry.localhost` hostname

The `.localhost` suffix is needed as module manager is able fetch from local k3d registry via http only if it is suffixed with `.localhost`. This comes directly from the inherent design of [k3d registries](https://k3d.io/v4.4.8/usage/guides/registries/ )

Kyma CLI provision command does not create `k3d-kyma-registry.localhost` but  `k3d-kyma-registry` and such registry is expected when running `kyma deploy` command.

**Expected result**

Coredns config should allow module manager to resolve `k3d-kyma-registry.localhost` and  fetch module images from local registry when a local setup is provisioned using kyma cli commands:

```
kyma provision k3d
kyma alpha deploy
```

**Actual result**

Currently when using kyma cli to test modular kyma locally we get host lookup error for `k3d-kyma-registry.localhost`

```bash
kyma provision k3d
kyma alpha deploy
```

Logs from module manager:
```
[...]

PR #1455, opened 2022-11-22
Fix coredns patching for modular kyma on k3d

kwiatekus:

- patch coredns not only for `deploy` but also for `alpha deploy` command
- update the `NodeHosts` template to include `k3d-{clustername}-registry.localhost` variant 

**Related issue(s)**
https://github.com/kyma-project/cli/issues/1454
Found by https://github.com/kyma-project/keda-manager/issues/57

### DIFF
--- a/cmd/kyma/alpha/deploy/cmd.go
+++ b/cmd/kyma/alpha/deploy/cmd.go
@@ -31,6 +31,13 @@ type command struct {
 	opts *Options
 }
 
+const (
+	hostsTemplate = `
+    {{ .K3dRegistryIP}} {{ .K3dRegistryHost}}
+    {{ .K3dRegistryIP}} {{ .K3dRegistryHost}}.localhost
+`
+)
+
 // NewCmd creates a new deploy command
 func NewCmd(o *Options) *cobra.Command {
 
@@ -168,7 +175,7 @@ func (cmd *command) deploy(start time.Time) error {
 		return err
 	}
 
-	if _, err := coredns.Patch(l.Desugar(), cmd.K8s.Static(), false, clusterInfo); err != nil {
+	if _, err := coredns.Patch(l.Desugar(), cmd.K8s.Static(), false, clusterInfo, hostsTemplate); err != nil {
 		return err
 	}
 
--- a/cmd/kyma/alpha/deploy/cmd.go
+++ b/cmd/kyma/alpha/deploy/cmd.go
@@ -6,6 +6,8 @@ import (
 	"os"
 	"time"
 
+	"github.com/kyma-project/cli/internal/clusterinfo"
+	"github.com/kyma-project/cli/internal/coredns"
 	"github.com/kyma-project/cli/internal/deploy"
 	"github.com/kyma-project/cli/internal/kustomize"
 	"github.com/kyma-project/cli/pkg/dashboard"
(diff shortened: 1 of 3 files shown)


internal/coredns/coredns.go
```diff
@@ -49,6 +49,7 @@ const (
 `
 	hostsTemplate = `
     {{ .K3dRegistryIP}} {{ .K3dRegistryHost}}
+	{{ .K3dRegistryIP}} {{ .K3dRegistryHost}}.localhost
```
jakobmoellersap, line 52: would it be possible to do the localhost patching only on the alpha command. there are concerns in the team that we do not want localhost exposure for the current deploy.

cmd/kyma/alpha/deploy/cmd.go
```diff
@@ -131,6 +134,11 @@ func (cmd *command) deploy(start time.Time) error {
 		defer func() { os.Stderr = stderr }()
 	}
 
+	clusterInfo, err := clusterinfo.Discover(context.Background(), cmd.K8s.Static())
```
jakobmoellersap, line 137: Would you kindly define a timeout here and also try to think if you can pass a different ctx here than main?

cmd/kyma/alpha/deploy/cmd.go
```diff
@@ -145,6 +153,10 @@ func (cmd *command) deploy(start time.Time) error {
 		return err
 	}
 
+	if _, err := coredns.Patch(l.Desugar(), cmd.K8s.Static(), false, clusterInfo); err != nil {
```
jakobmoellersap, line 156: could you write a test for this?

jakobmoellersap (changes requested):
Just 3 nits, other than that LGTM

cmd/kyma/alpha/deploy/cmd.go
kwiatekus: the same discover function is used exactly the same way in the `deploy` and `undeploy` command.
If you think its necessary to change it, than it should be changed all over the place.
I would prefer not to change it as part of this PR

cmd/kyma/alpha/deploy/cmd.go
kwiatekus: the same `coredns.Patch` is executed in `deploy command`.
The patching logic lives in its own package and it was introduced w/o tests.
I agree this needs tests, but I wouldn't add those as part of this PR.

cmd/kyma/alpha/deploy/cmd.go
jakobmoellersap: Im not worried about the Discover function. I am saying you should probably think of passing `context.WithTimeout(cmd.Context(), INSERT_TIMEOUT_HERE)`.

cmd/kyma/alpha/deploy/cmd.go
jakobmoellersap: Just because a command is not tested is not exactly a reason for me to not test it in the future. Feel free to contribute tests as needed. If you feel like this should not be done here, then I would block this PR until the original Patching was tested. WDYT?

kwiatekus: how come the original contribution with patch command passed the review than ? 
Thats a mystery

jakobmoellersap: While I am not agreeing with this sentiment, I feel like I'm in the minority here so I will close the comment. I will say however that this way the CLI will never increase coverage. I am also gonna talk internally about more stricter standards in general for CLI so there won't be this argument anymore.

jakobmoellersap (dismissed):

kwiatekus:
Adjustment for error handling in alpha deploy command was necessary after rebase.
@jakobmoellersap your approval was automatically dismissed.
Please revisit

jakobmoellersap (approved):

Issue #1422, opened 2022-10-24
deploy: start busola at the end


clebs:
**Description**
The deploy command installs Kyma on a cluster but it does not enable any modules per default. To lower the barrier of entry, start busola after deploying to guide the user's next steps and configuration.

**Reasons**

Lower barrier of entry into Kyma for new users.

PR #1468, opened 2022-12-01
Open the dashboard at the end of deployment

clebs:
**Description**

- Refactored the dashboard command.
- Encapsulated dashboard logic.
- Added dashboard to the deploy command.

**Related issue(s)**
#1422

### DIFF
--- a/cmd/kyma/alpha/deploy/cmd.go
+++ b/cmd/kyma/alpha/deploy/cmd.go
@@ -277,8 +277,9 @@ func (cmd *command) wizard() error {
 		cmd.CurrentStep.Failure()
 		return err
 	}
-	// make sure the dahboard container always stops at the end
+	// make sure the dahboard container always stops at the end and the cursor restored
 	cmd.Finalizers.Add(dash.StopFunc(context.Background(), func(i ...interface{}) { fmt.Print(i...) }))
+	cmd.Finalizers.Add(func() { cmd.CurrentStep.Stop(true) })
 
 	if err := dash.Open(fmt.Sprintf("/cluster/%s/namespaces/%s/kymas/details/%s", cluster, ns, name)); err != nil {
 		cmd.CurrentStep.Failure()
--- a/cmd/kyma/alpha/deploy/cmd.go
+++ b/cmd/kyma/alpha/deploy/cmd.go
@@ -1,6 +1,7 @@
 package deploy
 
 import (
+	"context"
 	"fmt"
 	"log"
 	"os"
@@ -8,13 +9,16 @@ import (
 
 	"github.com/kyma-project/cli/internal/deploy"
 	"github.com/kyma-project/cli/internal/kustomize"
+	"github.com/kyma-project/cli/pkg/dashboard"
 	"github.com/kyma-project/cli/pkg/errs"
 
-	"github.com/pkg/errors"
+	"errors"
+
 	"github.com/spf13/cobra"
 	"go.uber.org/zap"
 	appsv1 "k8s.io/api/apps/v1"
 	corev1 "k8s.io/api/core/v1"
+	v1 "k8s.io/apimachinery/pkg/apis/meta/v1"
 
 	"github.com/kyma-project/cli/internal/cli"
 	"github.com/kyma-project/cli/internal/kube"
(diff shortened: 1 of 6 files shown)


cmd/kyma/alpha/deploy/cmd.go
```diff
@@ -100,10 +104,18 @@ func (cmd *command) run() error {
...
+	if err := cmd.deploy(start); err != nil {
+		return err
+	}
+
+	if err := cmd.wizard(); err != nil {
+		return err
 	}
 
-	return cmd.deploy(start)
+	return cmd.wizard()
```
jeremyharisch, line 118: Any specific reason why you call the wizard twice? Once should be enough.

I think you can remove the 
```
if err := cmd.wizard(); err != nil {
		return err
}
```

internal/deploy/kyma.go
```diff
@@ -25,6 +26,12 @@ spec:
     enabled: {{ .Sync }}
 `
 
+var KymaGVR = schema.GroupVersionResource{
```
jeremyharisch, line 29: If you want to avoid having dependencies to the lifecycle-manager, it is fine - If not I would suggest importing the GVK from the lifecycle-manager directly instead of hardcoding.

cmd/kyma/alpha/deploy/cmd.go
```diff
@@ -100,10 +104,18 @@ func (cmd *command) run() error {
...
 	if cmd.K8s, err = kube.NewFromConfigWithTimeout("", cmd.KubeconfigPath, cmd.opts.Timeout); err != nil {
-		return errors.Wrap(err, "failed to initialize the Kubernetes client from given kubeconfig")
+		return fmt.Errorf("failed to initialize the Kubernetes client from given kubeconfig: %w", err)
+	}
+
+	if err := cmd.deploy(start); err != nil {
+		return err
+	}
+
+	if err := cmd.wizard(); err != nil {
```
jeremyharisch, line 114: Wizard should not be enabled if `ci`-flag is set. At least we need a mechanism to turn off the default behaviour of opening the dashboard, i.e. by using the CI flag

jeremyharisch (commented):
When running `kyma alpha deploy` the Busola dashboard gets started in a docker container and the browser automatically opens for the user, which is great I think. But in the console the User will be stuck with 
```
/ Start module dashboard 
```
with a spinning dash at the front. Thus, the user needs to quite it by using CTRL + C. At least give a hint at the beginning to the User that the dashboard will open at the end and the user needs to quit it using CTRL + C. Or figure a more user friendly way out.

Plus, if the user quits it, it shows the following:
```
- Removing container 123999b3961bda2505241e08b3f6e16f011feccef25cdbf8996ad5dc4ced5c58...
\ Start module dashboard %     
```
When the container is going to be removed, please make sure the `Start module dashboard` is not shown anymore, since this is not correct.
And maybe also while Busola is running inside of Docker container, already change the msg from `\ Start module dashboard` to something like `Kyma dashboard is running in local docker container - use CTRL + C to quit`

Furthermore, I think we should stick with `Kyma dashboard` or `Busola` instead of `module dashboard`.

jeremyharisch:
Furthermore, I also think the log output of Busola should be hidden bedefault

Screenshot 2022-12-06 at 16 06 30

I do not think this is very appealing to first time user. Such specific log output should be hidden by default and could be enabled by a flag. What do you think?

cmd/kyma/alpha/deploy/cmd.go
clebs: This was a huge mistake! Thanks for catching it.

cmd/kyma/alpha/deploy/cmd.go
clebs: Absolutely, although I thought about it, I then did not add it here.

internal/deploy/kyma.go
clebs: Yes, I did it this way on purpose. Did not see the value of pulling in and maintaining a dependency for just a couple of strings.

But of course if in the future the CLI needs something else from the lifecycle manager it would be good to use that GVK.

clebs:
Hi @jeremyharisch,
thanks for the in depth review and test. Regarding the output of the command:
- Now the following message appears after starting the dashboard: `Dashboard started. To stop press Ctrl+C`.
- Logs from the container only show in verbose mode.
- Dashboard does not start in non interactive/CI mode.

jeremyharisch (approved):
Looks great now 🔥

Issue #1470, opened 2022-12-05
Kyma CLI complains on missing `kustomize` binary


pbochynski:
**Description**

Kyma CLI complains on missing `kustomize` binary. I had to add kustomize manually to ~/.kyma/bin folder to continue.

**Expected result**

Kustomize is downloaded automatically or user is instructed what to do.

**Actual result**

```
kyma alpha create module --version 0.0.1 -w  --registry myregistry
WARNING: This command is experimental and might change in its final version. Use at your own risk.
X Parse and build module...
Error: could not edit kustomization image: error getting kustomize binary: exec: "/Users/myuser/.kyma/kustomize": stat /Users/myuser/.kyma/kustomize: no such file or directory
```

Tomasz-Smelcerz-SAP:
Confirmed. The `kustomize` binary is not installed in the setup of the `alpha create module` command.

PR #1473, opened 2022-12-07
Setup kustomize on kyma alpha create module

Tomasz-Smelcerz-SAP:
**Description**

kustomize installation is not ensured in `kyma alpha create module` command, although kustomize binary may be used during the execution of this command

- Add kustomize setup step to the `kyma alpha create module` command
- Extract kustomize and envTest setup code to a separate package: `internal/cli/setup`

**Related issue(s)**
Fixes: #1470

### DIFF
--- a/cmd/kyma/alpha/create/module/module.go
+++ b/cmd/kyma/alpha/create/module/module.go
@@ -12,6 +12,7 @@ import (
 	"go.uber.org/zap"
 
 	"github.com/kyma-project/cli/internal/cli"
+	"github.com/kyma-project/cli/internal/cli/setup"
 	"github.com/kyma-project/cli/pkg/module"
 )
 
@@ -93,6 +94,10 @@ func (cmd *command) Run(args []string) error {
 		Overwrite:     cmd.opts.Overwrite,
 		RegistryURL:   cmd.opts.RegistryURL,
 		DefaultCRPath: cmd.opts.DefaultCRPath,
+	}
+
+	if err := setup.Kustomize(&cmd.Command); err != nil {
+		return err
 	}
 
 	/* -- Inspect and build Module -- */
--- a/cmd/kyma/alpha/deploy/cmd.go
+++ b/cmd/kyma/alpha/deploy/cmd.go
@@ -2,12 +2,10 @@ package deploy
 
 import (
 	"fmt"
-	"log"
 	"os"
 	"time"
 
 	"github.com/kyma-project/cli/internal/deploy"
-	"github.com/kyma-project/cli/internal/kustomize"
 	"github.com/kyma-project/cli/pkg/errs"
 
 	"github.com/pkg/errors"
@@ -17,6 +15,7 @@ import (
 	corev1 "k8s.io/api/core/v1"
 
 	"github.com/kyma-project/cli/internal/cli"
+	"github.com/kyma-project/cli/internal/cli/setup"
 	"github.com/kyma-project/cli/internal/kube"
 	"github.com/kyma-project/cli/internal/nice"
 )
@@ -107,7 +106,7 @@ func (cmd *command) run() error {
 }
 
 func (cmd *command) deploy(start time.Time) error {
-	if err := cmd.initialSetup(); err != nil {
+	if err := setup.Kustomize(&cmd.Command); err != nil {
 		return err
 	}
 
@@ -177,15 +176,6 @@ func (cmd *command) deploy(start time.Time) error {
 
 	deployTime := time.Since(start)
 	return summary.Print(deployTime)
-}
-
-func (cmd *command) initialSetup() error {
-	s := cmd.NewStep("Setting up kustomize...")
-	if err := kustomize.Setup(s, true); err != nil {
-		log.Fatal(err)
-	}
-	s.Successf("Kustomize ready")
-	return nil
 }
 
 func (cmd *command) dryRun() error {
--- a/internal/cli/setup/kustomize.go
+++ b/internal/cli/setup/kustomize.go
@@ -1,0 +1,129 @@
+package setup
+
+import (
+	"errors"
+	"fmt"
+	"log"
+	"os"
+	"os/exec"
+	"path/filepath"
+	"regexp"
+	"strings"
+
+	"github.com/kyma-project/cli/internal/cli"
+	"github.com/kyma-project/cli/internal/envtest"
+	"github.com/kyma-project/cli/internal/files"
+	"github.com/kyma-project/cli/internal/kustomize"
+	"github.com/kyma-project/cli/pkg/step"
+)
+
+const (
+	DefaultVersion  = "1.24.x!" //1.24.x! means the latest available patch version for 1.24 branch
+	versionEnv      = "ENVTEST_K8S_VERSION"
+	envtestSetupBin = "setup-envtest"
+)
+
+// based on "kubernetes-sigs/controller-runtime/tools/setup-envtest/versions/parse.go", but more strict
+var envtestVersionRegexp = regexp.MustCompile(`^(0|[1-9]\d{0,2})\.(0|[1-9]\d{0,2})\.(0|[1-9]\d{0,3})$`)
+
+func Kustomize(cmd *cli.Command) error {
+	s := cmd.NewStep("Setting up kustomize...")
+	if err := kustomize.Setup(s, true); err != nil {
+		log.Fatal(err)
+	}
+	s.Successf("Kustomize ready")
+	return nil
+}
+
+func EnvTest(step step.Step, verbose bool) (*envtest.Runner, error) {
+	p, err := files.KymaHome()
+	if err != nil {
+		return nil, err
+	}
+
+	//Install setup-envtest
+	if _, err := os.Stat(filepath.Join(p, envtestSetupBin)); os.IsNotExist(err) {
+		kymaGobinEnv := "GOBIN=" + p
+		envtestSetupCmd := exec.Command("go", "install", "sigs.k8s.io/controller-runtime/tools/setup-envtest@latest")
+		envtestSetupCmd.Env = os.Environ()
+		envtestSetupCmd.Env = append(envtestSetupCmd.Env, kymaGobinEnv)
+
+		//go install is silent when executed successfully
+		out, err := envtestSetupCmd.CombinedOutput()
+		if err != nil {
+			return nil, fmt.Errorf("error installing setup-envtest: %w. Details: %s", err, string(out))
+		} else if verbose {
+			step.LogInfof("Installed setup-envtest in: %q", p)
+		}
+
+	}
+
+	//Install envtest binaries using setup-envtest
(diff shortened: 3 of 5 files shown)


internal/cli/setup/setup.go
```diff
@@ -0,0 +1,129 @@
...
+const (
+	envtestDefaultVersion = "1.24.x!" //1.24.x! means the latest available patch version for 1.24 branch
+	envtestVersionEnv     = "ENVTEST_K8S_VERSION"
+	envtestSetupBin       = "setup-envtest"
+)
+
+// based on "kubernetes-sigs/controller-runtime/tools/setup-envtest/versions/parse.go", but more strict
+var envtestVersionRegexp = regexp.MustCompile(`^(0|[1-9]\d{0,2})\.(0|[1-9]\d{0,2})\.(0|[1-9]\d{0,3})$`)
+
+func Kustomize(cmd *cli.Command) error {
```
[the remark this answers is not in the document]
Tomasz-Smelcerz-SAP: I just want to remove duplication here... Let's see what I can accomplish here :)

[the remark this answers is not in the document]
clebs: Yes, I see where this is coming from, and would be great to remove as much duplication as possible without making it too stiff.

cmd/kyma/alpha/create/module/module.go
```diff
@@ -12,12 +12,14 @@ import (
...
 	"github.com/kyma-project/cli/internal/cli"
+	kustomizeSetup "github.com/kyma-project/cli/internal/cli/setup/kustomize"
 	"github.com/kyma-project/cli/pkg/module"
 )
 
 type command struct {
-	opts *Options
 	cli.Command
+	opts *Options
+	kustomizeSetup.ExtendCmd
```
ruanxin, line 22: why not directly use the function `SetupKustomize` in `kustomize` package, instead you define an `ExtendCmd`?

cmd/kyma/alpha/create/module/module.go
```diff
@@ -95,6 +97,13 @@ func (cmd *command) Run(args []string) error {
 		DefaultCRPath: cmd.opts.DefaultCRPath,
 	}
 
+	setupStep := cmd.NewStep(kustomizeSetup.DefaultNewStepMsg)
+	if err := cmd.SetupKustomize(setupStep); err != nil {
```
ruanxin, line 101: 1. Create the const value of `DefaultNewStepMsg` I don't thing it brings too much value, and the dependency is a bit messy here, here only need to write some output message, why I have to introduce dependency from internal`kustomize` package?
2. Why need to generate this `setupSetup` and pass in to SetupKustomize, instead of directly use `cmd.CurrentStep`? although they are same. I think you can simply use `cmd.NewStep`, `cmd.CurrentStep.Successf`

internal/cli/setup/kustomize/setup.go
```diff
@@ -0,0 +1,22 @@
...
+	"github.com/kyma-project/cli/internal/kustomize"
+	"github.com/kyma-project/cli/pkg/step"
+)
+
+const (
+	DefaultNewStepMsg     = "Setting up kustomize..."
+	DefaultStepSuccessMsg = "Kustomize ready"
+)
+
+// ExtendCmd is used to extend the CLI command with the SetupKustomize() function
```
ruanxin, line 13: IMO this `ExtendCmd` struct is over engineered, you can simply define a `func SetupKustomize(s step.Step) error`

ruanxin (changes requested):
in general lgtm, please check my comments.

cmd/kyma/alpha/create/module/module.go
Tomasz-Smelcerz-SAP: 1. This constant is used twice in different commands: `alpha create module` and `alpha deploy`. I personally hate code/constant repetition, so this is my take on removing the duplication. Do you have any other idea how could we share 

2. It was like this before, kustomize setup was a separate step with a distinct name. I didn't want to change that.

internal/cli/setup/kustomize/setup.go
Tomasz-Smelcerz-SAP: As described above, it's my take on introducing some shared logic into existing command structure.
Essentially a way to easily add it to the commands just by extending their types.
A single "mixin" doesn't bring that much value, but if we have more of them, then you can tell what the command does, at least partially, just by looking at it's type.
I'd prefer to do it with a proper mixin mechanism known from other languages, but golang doesn't have that.
Let me know what you think :)

cmd/kyma/alpha/create/module/module.go
ruanxin: 1. I still think it's not a good approach to introduce this dependencies just because somehow the message is coincidentally same. I mean the message in different process can be different in order to provide more specific information. e.g `Setting up kustomize...` can be `Setting up kustomize during create module` in create module command and `Setting up kustomize during deploy` for deploy command. What I mean is we don't need to worry about sharing such values.

2. You can reference the style just a few line after, all these step messages using `cmd.NewStep`, `cmd.CurrentStep`
directly.
```
cmd.NewStep("Parse and build module...")
// Create base resource defs with module root and its sub-layers
if err := module.Inspect(modDef, cmd.opts.ResourcePaths, cmd.CurrentStep, l); err != nil {
cmd.CurrentStep.Failure()
return err
}
cmd.CurrentStep.Successf("Module built")
```

cmd/kyma/alpha/create/module/module.go
ruanxin: OK, thanks for explain, the purpose to composite command is understandable. But I actually thinking from different point of view:
1. If we are talking about the mixin, the original idea for mixin is want to let a class (e.g: class A) which contains methods defined by other classes (e.g: class B) without inheritance (so called composition). But why we need to declare class B inside class A first? Because in other common OO language like JAVA, there is no way you can simply define a method without belonging to a class. But golang already provide us this ability, just simply import package, then you can immediately use public method inside those package, which means the composite is naturally achievable in package level. So we don't need to declare an empty `ExtendCmd` struct in order to use method. To me, declare a class (structs) to achieve mixin is just because some language lack of the ability to directly use method, so it's a compromised indirect way.

2. Regarding readability, maybe by read the structs, I mean `command` struct, we can briefly know this command contains kustomize setup command, or later other commands. But if you want to find in the code where is it actually used, simple have no idea, because those `ExtendCmd` is implied by the `command`, every place you use the actually method declared in `ExtendCmd`, you basically use it with `command`, like current case, instead of calling `ExtendCmd.SetupKustomize`, we directly call `cmd.SetupKustomize`, to me, it's against improving readability. Instead, by import package, I can use different package name to call this method.

jeremyharisch: @Tomasz-Smelcerz-SAP I think @ruanxin got a good point here. IMO this seems a bit like over engineering. We already have the public available kustomize Setup function in [here](https://github.com/kyma-project/cli/blob/f94071d921d50b73316483c317ce7edbeb20858d/internal/kustomize/kustomize.go#L70). This could be directly called instead of having to extend it with a new struct.

I totally get your point, if having a short declaration including all the necessary steps to have kind of like a `shorter` summary of what the command is doing, just by reading the struct. But for this I think we need to come up with a proper interface solution, which could take more effort, since we then also need to adapt all the other commands.

I would suggest to go now for the simplest solution, meaning calling the kustomize setup step directly from the package.

cmd/kyma/alpha/create/module/module.go
jeremyharisch: Or you  could introduce a new public function in the ksutomize package which already includes the step, something like
```
func SetupWithDefaultStep(cmd *command, verbose bool) error {
	setupStep := cmd.NewStep("Setting up kustomize...")
	return Setup(setupStep, verbose)
}
```

And then call this method instead of creating two steps with the same message

internal/cli/setup/kustomize/setup.go
jeremyharisch: https://github.com/kyma-project/cli/pull/1473/files#r1044166902

jeremyharisch (commented):
Works, but could be improved from code perspective - Please have a look at my comment

cmd/kyma/alpha/create/module/module.go
Tomasz-Smelcerz-SAP: @jeremyharisch This was my initial solution. And it was strictly objected by Borja, who said that new steps should be only created by commands. And indeed, now the "NewStep" is only called in the "cmd" package tree

Tomasz-Smelcerz-SAP: @ruanxin ```[...] The message in different process can be different[...]. What I mean is we don't need to worry about sharing such values.```
In my case I WANT to have the same message. I know it could be different, but I WANT the same.
This means I WANT to use the language features to actually give other developers a strict, clear message about my intentions: I WANT it to be the same.
That's why I require this message not only to be the same value (copied in two places) - I require it to be the same instance of string, so if anyone changes the value, it's changed in two places.
If any other dev in the future decides that the message should be different for some reasons, then the dev will introduce necessary changes. But for now I don't see a reason we should give up and accept constant duplication.

ruanxin: ok, but tbh, 
first, I still not get the point what's the reason need these messages to be same so strictly? Maybe you can explain more about it, if this is truly the case, in order to prevent difference, then we need have a test to make sure the message generated from different command should be strictly same.

second, with my own experience and common refactoring understanding, define a const value normally will give developer the impression this value should be always same, in very rear case, people will have the thoughts to change it in later refactoring, that's why I insist for this flexible hint message, there is no needs to define them as const, because they can be changed in any time. I think with this point, @clebs shared same thought, by referencing his comment: `not all commands will want the same message when calling NewStep() in the kustomize setup step.`, by moving this message out of `SetupKustomize` funtion but always use those two default const message surrounded is another way to indicate this command should use same message.

In current code base, we can also find many "duplicate" messages existing, like `"Kyma deployed successfully!"`, `"Deploying Kyma"`, developers should not care about this duplication during writing such hint messages, unless we have solid reason.

ruanxin: ok, thanks for the feedback @Tomasz-Smelcerz-SAP , since the above discussion between us is related to how to deal with hint message and it's really not something related to critical process, I suggest we end up here. At least we are not prevent further refactoring to improve the code.

Last but not least, regarding point 2. I think you misunderstood me, the thing I want to test is testing **why** these two message needs to be same, if they are different, what kind of business logic will be break, not testing these two message expect to be same, that's of cause not necessary.

ruanxin: Regarding point 5, I think this is a major difference between us (to me, it's definitely not treat as code smell, and there is not necessary to get rid of these so called duplication) and will affect later the direction how we maintain the code, I suggest we create a short discussion meeting next week, including few more opinions. What do you think, can you manage to take care of this discussion? @Tomasz-Smelcerz-SAP

ruanxin (approved):

PR #1489, opened 2022-12-27
Read default CR

pbochynski:
**Description**

- handle properly default CR

Solves: https://github.com/kyma-project/cli/issues/1491

### DIFF
--- a/cmd/kyma/alpha/create/module/module.go
+++ b/cmd/kyma/alpha/create/module/module.go
@@ -195,6 +195,15 @@ func (cmd *command) Run(args []string) error {
 
 func (cmd *command) validateDefaultCR(modDef *module.Definition, l *zap.SugaredLogger) error {
 	cmd.NewStep("Validating Default CR")
+
+	if modDef.DefaultCRPath != "" {
+		cr, err := os.ReadFile(modDef.DefaultCRPath)
+		if err != nil {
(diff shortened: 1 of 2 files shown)


pbochynski:
Can someone review this PR?

cmd/kyma/alpha/create/module/module.go
```diff
@@ -195,6 +195,15 @@ func (cmd *command) Run(args []string) error {
 
 func (cmd *command) validateDefaultCR(modDef *module.Definition, l *zap.SugaredLogger) error {
 	cmd.NewStep("Validating Default CR")
+
+	if modDef.DefaultCRPath != "" {
```
ruanxin, line 199: This prefill DefaultCR is also done in `inspectProject` project step https://github.com/kyma-project/cli/blob/main/pkg/module/resources.go#L213-L226 , but have the precondition `if def.RegistryURL != "" `, I don't have too much detail context for this decision. IMO, if this `DefaultCR` in anyway needs to be filled when `DefaultCRPath` exits, then it's better remove that precondition in `module.Inspect`, this step is happen before `validateDefaultCR`

ruanxin (changes requested):
only have some concern for the place to do this `DefaultCR` initialization.

pbochynski:
@ruanxin I moved DefaultCR initialization before validation and removed duplicated code from another place.

ruanxin (approved):

PR #1500, opened 2023-01-18
add enable-registry-cred flag to cli

ruanxin:
**Description**
For private oci registry, this feature request introduce a flag to automatically appended label into component descriptor resources, which can be used for the secret label which provide registry credentials.

**Related issue(s)**
https://github.com/kyma-project/module-manager/issues/201

### DIFF
--- a/cmd/kyma/alpha/create/module/module.go
+++ b/cmd/kyma/alpha/create/module/module.go
@@ -66,6 +66,7 @@ Build module my-domain/modB in version 3.2.1 and push it to a local registry "un
 	cmd.Flags().StringVar(&o.ModCache, "mod-cache", "./mod", "Specifies the path where the module artifacts are locally cached to generate the image. If the path already has a module, use the overwrite flag to overwrite it.")
 	cmd.Flags().StringArrayVarP(&o.ResourcePaths, "resource", "r", []string{}, "Add an extra resource in a new layer with format <NAME:TYPE@PATH>. It is also possible to provide only a path; name will default to the last path element and type to 'helm-chart'")
 	cmd.Flags().StringVar(&o.RegistryURL, "registry", "", "Repository context url for module to upload. The repository url will be automatically added to the repository contexts in the module")
+	cmd.Flags().BoolVar(&o.EnableRegistryCred, "enable-registry-cred", false, "Enable this flag to append a label into each descriptor resource, it indicates the dockerconfigjson secret (config separately) which provide credential for private repository.")
 	cmd.Flags().StringVarP(&o.Credentials, "credentials", "c", "", "Basic authentication credentials for the given registry in the format user:password")
 	cmd.Flags().StringVar(&o.DefaultCRPath, "default-cr", "", "File containing the default custom resource of the module. If the module is a kubebuilder project, the default CR will be automatically detected.")
 	cmd.Flags().StringVarP(&o.TemplateOutput, "output", "o", "template.yaml", "File to which to output the module template if the module is uploaded to a registry")
@@ -173,7 +174,7 @@ func (cmd *command) Run(args []string) error {
 		cmd.CurrentStep.Success()
 
 		cmd.NewStep("Generating module template")
-		t, err := module.Template(archive, cmd.opts.Channel, modDef.DefaultCR)
+		t, err := module.Template(archive, cmd.opts.Channel, modDef.DefaultCR, cmd.opts.EnableRegistryCred)
 		if err != nil {
 			cmd.CurrentStep.Failure()
 			return err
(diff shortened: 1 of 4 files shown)


pkg/module/template.go
```diff
@@ -38,13 +39,25 @@ spec:
...
 		return nil, err
 	}
-
-	ref, err := oci.ParseRef(d.Name)
+	if enableRegistryCred {
+		for i := range descriptor.Resources {
+			resource := &descriptor.Resources[i]
+			resource.SetLabels([]v2.Label{{
+				Name:  OCIRegistryCredLabel,
+				Value: json.RawMessage(fmt.Sprintf(`{"%s": "%s"}`, credSecretLabel, descriptor.Name)),
```
ruanxin, line 56: To Reviewer:
The reason why I choose the way not let users to define the label by themselves is a) simplify the cli command, so that users don't have to write labels in the command, b) avoid label validation. c) provide a standard label which can be used by module teams to provide it in the credential secret.

Of cause if user still have the use case to use customized label, they can simply manually overwrite the label in the generated module template, but not recommend.

jakobmoellersap: I believe your main intention is good and I support having a fixed label for the registry credential. However, I believe that also fixing the registry value is not ideal. Especially because that means that every module template will need its own secret which is pretty counter-intuitive and hard to maintain. I suggest to leave the value open for users to fill and only have the key defined. That way multiple modules will be able to use the same registry credential if desired. WDYT?

jakobmoellersap (changes requested):
I believe the flag should not be a boolean since restricting the value of the label (not the key) is imo too static for a real life scenario. Could we change this to something more flexible?

pkg/module/template.go
ruanxin: Yes, you are right, to use a shared secret is not supported by this way. But just to clarify, I think for security concern we should encourage teams use their own repository credential if necessary compare to share the credential, or? 

But to leave the value customizable is fine to me, I'll change the flag to something like `--registry-cred-selector-value`, is it fine?

docs/gen-docs/kyma_alpha_create_module.md
```diff
@@ -47,6 +47,7 @@ Build module my-domain/modB in version 3.2.1 and push it to a local registry "un
       --clean                  Remove the mod-path folder and all its contents at the end.
   -c, --credentials string     Basic authentication credentials for the given registry in the format user:password
       --default-cr string      File containing the default custom resource of the module. If the module is a kubebuilder project, the default CR will be automatically detected.
+      --enable-registry-cred   Enable this flag to append a label into each descriptor resource, it indicates the dockerconfigjson secret (config separately) which provide credential for private repository.
```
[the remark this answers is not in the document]
ruanxin: Yes, agree, tbh, it's hard to me explain this clearly enough with a short description, do you have some good suggestions? And btw, the detail description will be presented in the template operator document.

pkg/module/template.go
```diff
@@ -38,13 +39,27 @@ spec:
 {{yaml .Descriptor | printf "%s" | indent 4}}
 `
 
-func Template(archive *ctf.ComponentArchive, channel string, data []byte) ([]byte, error) {
-	d, err := remoteDescriptor(archive)
+	//nolint:gosec
+	OCIRegistryCredLabel = "oci-registry-cred"
```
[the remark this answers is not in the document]
ruanxin: no idea 😂 , I think maybe lint just found it contains some sensitive words like `cred` something.

pkg/module/template.go
ruanxin: As discussed, I made the change to accept the label selectors as flag input.

docs/gen-docs/kyma_alpha_create_module.md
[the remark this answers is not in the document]
ruanxin: I changed the flag, and modified the description a bit, please check if it's fine, and the guide also provided in this pr https://github.com/kyma-project/lifecycle-manager/pull/352

cmd/kyma/alpha/create/module/module.go
```diff
@@ -66,6 +66,8 @@ Build module my-domain/modB in version 3.2.1 and push it to a local registry "un
 	cmd.Flags().StringVar(&o.ModCache, "mod-cache", "./mod", "Specifies the path where the module artifacts are locally cached to generate the image. If the path already has a module, use the overwrite flag to overwrite it.")
 	cmd.Flags().StringArrayVarP(&o.ResourcePaths, "resource", "r", []string{}, "Add an extra resource in a new layer with format <NAME:TYPE@PATH>. It is also possible to provide only a path; name will default to the last path element and type to 'helm-chart'")
 	cmd.Flags().StringVar(&o.RegistryURL, "registry", "", "Repository context url for module to upload. The repository url will be automatically added to the repository contexts in the module")
+	cmd.Flags().StringVar(&o.RegistryCredSelector, "registry-cred-selector", "",
+		"Use this flag to config label selectors to identify the dockerconfigjson secret (config separately) which provide credential for private repository. Example: \"label1=value1,label2=value2\"")
```
jakobmoellersap, line 70:
```suggestion
		"label selector to identify a secret of type kubernetes.io/dockerconfigjson (that needs to be created externally) which allows the image to be accessed in private image registries. This can be used if you push your module to a registry with authenticated access. Example: \"label1=value1,label2=value2\"")
```

jakobmoellersap:
Needs a doc adjustment, after that it should run fine

jakobmoellersap (approved):

jakobmoellersap:
Needs TW review

IwonaLanger (approved):

Issue #1494, opened 2023-01-02
CLI can't push image to docker hub oci registry


ruanxin:
[docker support oci already](https://www.docker.com/blog/announcing-docker-hub-oci-artifacts-support/), however, kyma cli can't push image to oci registry by offering `-c` flag with `username:accesstoken`

```
Error: unable to push config: server message: insufficient_scope: authorization failed
```

_Update: We now know this is caused by the generated oci artifact reference that has more than two path segments, while the dockerhub registry only allows for two. I am extending the AC's for this task_

AC:
- [ ] cli should support docker hub oci registry (edit: this is the original AC)
- [ ] `kyma alpha create module` should allow users to fully control the oci artifact reference (URL) via flags.

### Proposed syntax change:

1. Two parameters for the entire artifact reference: version + registry
syntax: `kyma alpha create module --version 0.0.5 --registry 'registry-1.docker.io/someuser/module' -c 'credentials'`
The resulting ref would be: `registry-1.docker.io/someuser/module:0.0.5`
The `--name` flag would be dropped from the CLI.

2. Two parameters that define the artifact reference: version + registry + name
syntax: `kyma alpha create module --version 0.0.5 --registry 'registry-1.docker.io' --name 'someuser/module' -c 'credentials'`
The resulting ref would be: `registry-1.docker.io/someuser/module:0.0.5`

janmedrek:
We would need to investigate whether the OCI images without the docker layer can be pushed to the docker registry at all.

Besides that - we need to clarify what are the requirements regarding docker hub support.

Tomasz-Smelcerz-SAP:
@ruanxin I confirm that docker hub should support OCI images. I have made a few tests using [ORAS](https://oras.land/cli/) tool. But I can't reproduce your error. Can you paste your command (without credentials)? Thanks!

Tomasz-Smelcerz-SAP:
Thanks, I was able to reproduce it using template operator.
From what I can see, our current code cannot push, but before pushing it tries to `fetch` - which makes sense, because the artifact/layer may already be pushed. And that `fetch` also fails, with a similar error
Here's the relevant entry from the logs (debug mode):
`DEBU[0007] Unable to check existence, continuing with push  error="server message: insufficient_scope: authorization failed"`
The `check existence` refers to the previous `HEAD` calls which are related to the mentioned `fetch` action.

ruanxin:
[...]

PR #1502, opened 2023-01-23
Add support for sha256-digest Component Name Mapping

Tomasz-Smelcerz-SAP:
**Description**

This PR adds support for `sha256-digest` Component Name Mapping to the `kyma alpha create module` command.
This is necessarty to allow the CLI to push the module artifact to the Dockerhub (registry-1.docker.io)

**Details**

The existing `kyma alpha create module` command is using OCM libraries with hardcoded constant for [Component Name Mapping](https://gardener.github.io/component-spec/component_descriptor_registries.html#component-name-mapping).
The resulting artifact names have always at least three path segments. According to the [OCM spec](https://gardener.github.io/component-spec/component_descriptor_registries.html#component-name-mapping), it is: `<baseUrl>/componentDescriptors/<componentName>:<versionTag>`
For `baseUrl` you have to choose something like: `registry-1.docker.io/username`.
The `componentName` [must include](https://gardener.github.io/component-spec/format.html#component-names) a domain name and an optional path suffix. In practice, users would always want to have a path suffix, for example: `kyma.io/istio-operator`.
In summary, that gives the following registry URL: `https://registry-1.docker.io/username/componentDescriptors/kyma.io/istio-operator:someTag`
Such URL does not work with dockerhub, because, apparently, the dockerhub only supports names like this: `registry-1.docker.io/username/repositoryName`

Now, the OCM spec provides an alternative name mapping, called [sha256-digest](https://gardener.github.io/component-spec/component_descriptor_registries.html#component-name-mapping-sha256-digest).
It allows for component references like: `<baseUrl>/<componentNameDigest>:<versionTag>`
This fits the Dockerhub restrictions, if one chooses `baseUrl` to be a standard `registry-1.docker.io/username`.

The price for that feature is that a "nice", human-readable artifact name is gone, it is replaced by `componentNameDigest` which is nothing else but the sha256 of the "real" component name. For example:  `kyma.io/istio-operator` becomes `0e2fe793c621f74359484c53fd4f15ea6a736dae26f88d4d2e1a92a98ca1cdee`
For now though, until the [OCM spec](https://github.com/gardener/component-spec) changes or the Dockerhub starts supporting longer paths, we don't have any other option.

- add `--nameMapping` flag to the `kyma alpha create module` command
- update OCM-related code paths to respect the flag value

**How to test**
`kyma alpha create module -v --version 0.0.1 -w --registry registry-1.docker.io/username --nameMapping=sha256-digest -c 'yourcredentials'`

**Related issue(s)**
#1494


cmd/kyma/alpha/create/module/module.go
```diff
@@ -123,14 +126,6 @@ func (cmd *command) Run(args []string) error {
 	}
 	cmd.CurrentStep.Successf("Module built")
 
-	if modDef.DefaultCRPath != "" {
```
jakobmoellersap, line 126: glad that you refactored this into the Inspect!

jakobmoellersap (commented):
My biggest question here is if this can be read and used as a module by the operators as Is. I am afraid that using the SHA digest will not actually help us resolve the module in the operators right? Did you verify that a module can be bundled and run like this by our operator stack? Other than that explanation makes sense to me and the Flag LGTM.

go.mod
```diff
@@ -36,9 +36,9 @@ require (
 	gotest.tools v2.2.0+incompatible
 	helm.sh/helm/v3 v3.10.3
 	istio.io/client-go v1.14.0
-	k8s.io/api v0.26.0
-	k8s.io/apiextensions-apiserver v0.26.0
-	k8s.io/apimachinery v0.26.0
+	k8s.io/api v0.25.4
```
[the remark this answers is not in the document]
Tomasz-Smelcerz-SAP: yes.

cmd/kyma/alpha/create/module/opts.go
```diff
@@ -5,16 +5,18 @@ import (
...
 	"github.com/pkg/errors"
-	"regexp"
 )
 
 // Options defines available options for the create module command
 type Options struct {
 	*cli.Options
 
 	Name                 string
+	NameMappingMode      string
```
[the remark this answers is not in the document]
Tomasz-Smelcerz-SAP: I've initially have this, but (most) other validations happen in the `validate()` function inside `pkg/module/def.go`. So I've moved it there. The validation inside module is safer because it's closer to the business logic, and we have more natural access to some constants - like the ones imported already from gardener's libs. Having double-validation here wasn't worth it in my opinion. But if you think we should have it, I'll add it again.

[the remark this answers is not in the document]
jakobmoellersap: sounds reasonable to me, lets leave it

pkg/module/remote.go
```diff
@@ -38,7 +39,18 @@ func Push(archive *ctf.ComponentArchive, r *Remote, log *zap.SugaredLogger) erro
 
 	// update repository context
 	if len(r.Registry) != 0 {
-		if err := cdv2.InjectRepositoryContext(archive.ComponentDescriptor, cdv2.NewOCIRegistryRepository(r.Registry, cdv2.OCIRegistryURLPathMapping)); err != nil {
+		nameMapping := cdv2.OCIRegistryURLPathMapping
+		if rc := archive.ComponentDescriptor.GetEffectiveRepositoryContext(); rc != nil {
```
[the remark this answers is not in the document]
Tomasz-Smelcerz-SAP: Thanks for catching that. Mechanical refactoring resulted in the code, that now looks a bit strange :)
I am getting the a value from the repositoryContext just to build a new object using this value and inject it back into the same repositoryContext... It may be redundant now, I will double-check.

Tomasz-Smelcerz-SAP:
@jakobmoellersap `My biggest question here is if this can be read and used as a module by the operators as Is.` You are rigth, we can't make use of it immediately. The lifecycle-manager, in `pkg/img/parse.go: 132`, returns an error: `error while parsing componentNameMapping sha256-digest: componentNameMapping not supported`.
I think technically we are able to add support for that.

jakobmoellersap:
How would that support look like? If you have an idea, feel free to submit a PR to Lifecycle Manager

Tomasz-Smelcerz-SAP:
@jakobmoellersap OK, I will see what I can do.

pkg/module/remote.go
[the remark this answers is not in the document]
Tomasz-Smelcerz-SAP: Fixed

Tomasz-Smelcerz-SAP:
This is blocked until https://github.com/kyma-project/lifecycle-manager/issues/364 is implemented. Though the code works, there's no point in merging if we're not sure that the other components can support the new mode.

Tomasz-Smelcerz-SAP:
@jakobmoellersap Because https://github.com/kyma-project/lifecycle-manager/issues/362 is resolved and tested with **this** code, we can merge this one. Please take a look again, I've updated the `sign module` cmd and introduced typed constants for the new flag, to avoid passing strings around to minimum.

cmd/kyma/alpha/sign/module/module.go
```diff
@@ -40,6 +40,7 @@ This command signs all module resources recursively based on an unsigned compone
 	cmd.Flags().StringVar(&o.ModPath, "mod-path", "./mod", "Specifies the path where the signed component descriptor will be stored")
 	cmd.Flags().StringVar(&o.SignatureName, "signature-name", "", "name of the signature for signing")
 	cmd.Flags().StringVar(&o.RegistryURL, "registry", "", "Repository context url where unsigned component descriptor located")
+	cmd.Flags().StringVar(&o.NameMappingMode, "nameMapping", "urlPath", "Overrides the OCM Component Name Mapping. Allowed values: \"urlPath\" or \"sha256-digest\". This flag is optional.")
```
jakobmoellersap, line 43:
```suggestion
	cmd.Flags().StringVar(&o.NameMappingMode, "nameMapping", "urlPath", "Overrides the OCM Component Name Mapping. Allowed values: \"urlPath\" or \"sha256-digest\".")
```
The flag is automatically optional since there is a default value, so no need for that in the docs

jakobmoellersap (dismissed):
Great work, thanks for the fast turnaround. I believe the docs need a bit more refinement as it's hard for me to understand but I also see that it's hard to fully see through the use case here. LGTM for me.

docs/gen-docs/kyma_alpha_create_module.md
```diff
@@ -20,9 +20,11 @@ Alternatively, a custom (non kubebuilder) module can be created by providing a p
 
 Optionally, you can manually add additional layers with contents in other paths (see [resource flag](#flags) for more information).
 
-Finally, if you provided a registry to which to push the artifact, the created module is validated and pushed. For example, the default CR defined in the \"default.yaml\" file is validated against CustomResourceDefinition.
+Finally, if you provided a registry to which to push the artifact, the created module is validated and pushed. During the validation the default CR defined in the optional \"default.yaml\" file is validated against CustomResourceDefinition.
+Alternatively, you can trigger an on-demand default CR validation with "--validateCR=true", in case you don't push to registry.
```
grego952, line 24:
```suggestion
Alternatively, you can trigger an on-demand default CR validation with `--validateCR=true`, in case you don't push to the registry.
```

docs/gen-docs/kyma_alpha_create_module.md
```diff
@@ -20,9 +20,11 @@ Alternatively, a custom (non kubebuilder) module can be created by providing a p
...
+
+To push the artifact into some registries, for example the central docker.io registry, you have to change the OCM Component Name Mapping with the following flag: --nameMapping=sha256-digest. This is necessary because the registry does not accept artifact URLs with more than two path segments, and such URLs are generated with the default name mapping: \"urlPath\". In the case of the \"sha256-digest\" mapping the artifact URL contains just a sha256 digest of the full Component Name and fits the path length restrictions.
```
grego952, line 26:
```suggestion
To push the artifact into some registries, for example, the central `docker.io` registry, you have to change the OCM Component Name Mapping with the following flag: `--nameMapping=sha256-digest`. This is necessary because the registry does not accept artifact URLs with more than two path segments, and such URLs are generated with the default name mapping: `\urlPath\`. In the case of the `\sha256-digest\` mapping, the artifact URL contains just a sha256 digest of the full Component Name and fits the path length restrictions.
```

docs/gen-docs/kyma_alpha_create_module.md
```diff
@@ -50,6 +52,7 @@ Build module my-domain/modB in version 3.2.1 and push it to a local registry "un
       --insecure                        Use an insecure connection to access the registry.
       --mod-cache string                Specifies the path where the module artifacts are locally cached to generate the image. If the path already has a module, use the overwrite flag to overwrite it. (default "./mod")
   -n, --name string                     Override the module name of the kubebuilder project. If the module is not a kubebuilder project, this flag is mandatory.
+      --nameMapping string              Overrides the OCM Component Name Mapping. Allowed values: "urlPath" or "sha256-digest". This flag is optional. (default "urlPath")
```
grego952, line 55:
```suggestion
      --nameMapping string              Overrides the OCM Component Name Mapping. Allowed values: `urlPath` or `sha256-digest`. This flag is optional. (default `urlPath`)
```

docs/gen-docs/kyma_alpha_sign_module.md
```diff
@@ -23,6 +23,7 @@ kyma alpha sign module MODULE_NAME MODULE_VERSION [flags]
   -c, --credentials string       Basic authentication credentials for the given registry in the format user:password
       --insecure                 Use an insecure connection to access the registry.
       --mod-path string          Specifies the path where the signed component descriptor will be stored (default "./mod")
+      --nameMapping string       Overrides the OCM Component Name Mapping. Allowed values: "urlPath" or "sha256-digest". (default "urlPath")
```
grego952, line 26:
```suggestion
      --nameMapping string       Overrides the OCM Component Name Mapping. Allowed values: `urlPath` or `sha256-digest`. (default `urlPath`)
```

Tomasz-Smelcerz-SAP:
@grego952 Hi!
Unfortunately, I can't accept your suggestions about replacing double-quotes with backticks. This is because of the specifics of the Kyma-CLI. Reasons below:
- Kyma-CLI is a command-line tool
- As such, it provides users with a help/usage message that is printed in the text terminal. This is the primary source of the documentation for the commands - users read that while interacting with the tool.
- To handle the flags/options and displaying the mentioned messages we're using [this library](https://github.com/spf13/cobra)
- The same library is actually used to generate the `*.md` files - we just take the text output for "short help" and "long descritpion", and concatenate it together to produce a valid `*.md` file.

Now the library "doesn't like" backticks, especially in the flags description. It strips them off.
I think I know what the reason is: In the terminal/shell, a backtick is NOT a quotation character! It has a very different meaning (execute a command a subshell).
This means that if we provide the help message with backticks embedded, and the user just copy-paste this into the terminal window (with backticks), a lot of bad things may happen.
For that reason we shouldn't embed backticks in CLI help/usage messages, and because these messages are sources from which the `*.md` files are generated, we're not able to provide backticks there. We can of course do some manual-patching, but it will be gone next time the `*.md` are generated. That would be a maintenance hell.

That's why I propose we stay with double quotes. It is a well established convention for delivering help/usage messages in countless cli tools.
Yes, the docs in `*.md` format will not be "perfect" (though perfectly readable), but as I said, this is a secondary source of the CLI documentation, the primary one is always the terminal output from the tool itself.

grego952 (approved):

jakobmoellersap (approved):

PR #1531, opened 2023-02-14
Security Scanners information in Module Template

khlifi411:
**Description**

- Introduce new step in create module command to add security scanning data from `sec-scanners-config.yaml` file provided by module developer teams


khlifi411 (commented):
**To reviewer**
The current implementation will fail when a `sec-scanners-config.yaml` is not provided. As this requirement wasn't communicated to team developers yet, should we fail silently when the file is not found?

cmd/kyma/alpha/create/module/module.go
```diff
@@ -158,6 +158,13 @@ func (cmd *command) Run(args []string) error {
 
 	cmd.CurrentStep.Successf("Image created")
 
+	/* -- ADD SECURITY SCANNING METADATA -- */
+	cmd.NewStep("Configuring security scanning...")
```
jakobmoellersap, line 162: can we change this so that the step is only activated if the file is found? that way you can silently opt out of this process if the file is not present. Alternative would be to define a bool flag which is disabled by default and can take in a file

pkg/module/security_scan.go
```diff
@@ -0,0 +1,152 @@
...
+
+var ErrFailedToParseImageURL = errors.New("error parsing protecode image URL")
+
+const (
+	secConfigFileName = "sec-scanners-config.yaml"
+	secScanLabelKey   = "scan.security.kyma-project.io"
+	secScanEnabled    = "enabled"
+)
+
+func AddSecurityScanningMetadata(descriptor *v2.ComponentDescriptor, modDef *Definition, fs vfs.FileSystem) error {
```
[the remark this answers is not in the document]
khlifi411: The lifecycle manager complains that the `ociRegistry` access type is not supported. I will raise a PR to introduce support for it 👍

jakobmoellersap (commented):
LGTM, needs a second review and then we are good to go I believe 👍

pkg/module/security_scan.go
```diff
@@ -0,0 +1,162 @@
...
+		if err != nil {
+			return err
+		}
+		imageLayerMetadata := v2.IdentityObjectMeta{
+			Name:    imageName,
+			Type:    v2.OCIImageType,
+			Version: descriptor.Version,
+			Labels:  []v2.Label{imageTypeLabel},
+		}
+		imageLayerAccess, err := v2.NewUnstructured(v2.NewOCIRegistryAccess(imageURL))
```
ruanxin, line 88: better mentioned in the [original issue](https://github.tools.sap/kyma/backlog/issues/3338) that the access layer format is different than proposed (there is no ref field), and also highlight it's according to gardener component-spec api definition https://github.com/gardener/component-spec/blob/master/bindings-go/apis/v2/accesstypes.go

```
  - access:
      imageReference: artifact.registry.hostname/path/to/the/registry/k8s-tools:v20221213-1840f45a
      type: ociRegistry
    labels:
    - name: type.scan.security.kyma-project.io
      value:
        type: third-party-image
    name: k8s-tools
    relation: external
    type: ociImage
    version: 0.3.4
```

ruanxin (approved):

grego952 (approved):

PR #1550, opened 2023-02-22
alpha create module: Enhance Default CR Selection

### LINKED ISSUE #1533: alpha create module: Enhance default CR selection
**Description**

When creating a module with multiple CRs in the samples folder, the CLI asks the user to specify which file is used as default CR. The user has to enter the exact file name to proceed. This is inconvenient and prone to errors.

A better way would be to enter a number corresponding to the CR to select it.

**Reasons**

Better UX and convenience.

LeelaChacha:
**Description**

Replaces the manual entry of default CR name when creating a module from template-operator with selection from a list.

**Related issue(s)**

Closes #1533


LeelaChacha:
Sorry guys there has been been a mistake. I only changed one file not 42. I will fix this

PR #1551, opened 2023-02-22
alpha create module: Improve Default CR Selection

### LINKED ISSUE #1533: alpha create module: Enhance default CR selection

LeelaChacha:
**Description**

Replaces the manual entry of default CR name when creating a module from template-operator with selection from a list.

**Related issue(s)**

Closes #1533

### DIFF
--- a/pkg/module/kubebuilder/project.go
+++ b/pkg/module/kubebuilder/project.go
@@ -5,6 +5,7 @@ import (
 	"io/fs"
 	"os"
 	"path/filepath"
+	"strconv"
 	"strings"
 
 	"github.com/kyma-project/cli/internal/kustomize"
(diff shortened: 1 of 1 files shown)


pkg/module/kubebuilder/project.go
```diff
@@ -155,13 +156,23 @@ func (p *Project) DefaultCR(s step.Step) ([]byte, error) {
 	defaultCR := ""
 	if len(d) > 1 {
 		// ask for specific file
-		names := []string{}
-		for _, f := range d {
-			names = append(names, f.Name())
+		var promptString strings.Builder
+		promptString.WriteString(fmt.Sprintf("Please specify the file to use as default CR in %s:\n", samplesDir))
+
+		files := map[int32]string{}
```
ruanxin, line 162: please not use int32, just int, the golang compiler will decide which during compile time based on your OS, if 64 bit then int64 or int32 in 32 bit system.

jeremyharisch (commented):
Besides two minor comments from me and Xin, looks good and works

pkg/module/kubebuilder/project.go
```diff
@@ -155,13 +156,23 @@ func (p *Project) DefaultCR(s step.Step) ([]byte, error) {
...
+		for i, f := range d {
+			files[int32(i+1)] = f.Name()
+			promptString.WriteString(fmt.Sprintf("[%d] %s\n", i+1, f.Name()))
```
ruanxin, line 165: consider give this `i+1` a meaning name

pkg/module/kubebuilder/project.go
```diff
@@ -155,13 +156,23 @@ func (p *Project) DefaultCR(s step.Step) ([]byte, error) {
...
+		}
+
+		answer, err := s.Prompt(promptString.String())
+		answerInt, _ := strconv.ParseInt(answer, 10, 32)
```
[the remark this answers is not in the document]
ruanxin: also consider the default input, if press enter, by default choosing the first option so that don't need to type 1, this will improve user experience.

ruanxin (changes requested):
overall looks quite good! just some small comments, please have a check.

pkg/module/kubebuilder/project.go
LeelaChacha: That's an helpful insight :)

pkg/module/kubebuilder/project.go
```diff
@@ -155,13 +156,23 @@ func (p *Project) DefaultCR(s step.Step) ([]byte, error) {
...
 	if len(d) > 1 {
 		// ask for specific file
-		names := []string{}
-		for _, f := range d {
-			names = append(names, f.Name())
+		var promptString strings.Builder
+		promptString.WriteString(fmt.Sprintf("Please specify the file to use as default CR in %s:\n", samplesDir))
+
+		files := map[int32]string{}
+		for i, f := range d {
```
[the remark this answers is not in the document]
LeelaChacha: Ah okay. Good tip. I just used the original variable name. But I guess it's wise to leave code better than you found it.

pkg/module/kubebuilder/project.go
```diff
@@ -144,30 +145,53 @@ func (p *Project) DefaultCR(s step.Step) ([]byte, error) {
 	}
 
 	samplesDir := filepath.Join(p.path, samplesPath)
-	d, err := os.ReadDir(samplesDir)
+	filez, err := os.ReadDir(samplesDir)
```
ruanxin, line 148: Another thing we can improve, at the moment, this readDir will basically read all files in the directory. But if it contains file other than yaml, it will also list them as options. This actually not a edge case, for example, if user edits CR files with vim, but quit unexpected, then it will have a temporary .swp file generated, like the example below:
```
2023-02-23T10:26:35.969+0100	DEBUG	module/resources.go:179	Kubebuilder project detected.
? Please specify the file to use as default CR in  template-operator/config/samples:
[1] .operator.kyma-project.io_v1alpha1_samplehelm.yaml.swp
[2] operator.kyma-project.io_v1alpha1_samplehelm.yaml
```

So I suggest implement some logic to filter out those non `yaml` or `yml` file.

pkg/module/kubebuilder/project.go
```diff
@@ -144,30 +145,53 @@ func (p *Project) DefaultCR(s step.Step) ([]byte, error) {
...
+		if answer == "" {
+			parsedAnswer = 1 // Default to the first choice
+		} else {
+			parsedAnswer, err = strconv.Atoi(answer)
+			if err != nil {
+				return nil, fmt.Errorf("could not obtain default CR from user prompt: %w", err)
+			}
+		}
+		fileName := files[parsedAnswer]
+		if len(fileName) == 0 {
```
ruanxin, line 184: `files` is a map, when you get value by key, you can return two variable, the second indicate if the key exists.

```
fileName, exists := files[parsedAnswer]
if exists {
}

pkg/module/kubebuilder/project.go
LeelaChacha: Is that filter really necessary? It seems that the function has already bin expanded a lot. Extra logic might affect the readability of the code.

LeelaChacha: This has been now implemented. User Experience >>

ruanxin (approved):

ruanxin (approved):

PR #1566, opened 2023-03-02
internal/kube/client: Modify Watch Timeout Strategy

### LINKED ISSUE #1492: Fix timeout contract in the "watch" function
**Description**

The `watch()` function in `./internal/kube/client_test.go` accepts a `timeout` parameter.
There's a non-standard contract on the function behavior when this parameter is zero.
Typically, a "zero timeout" means "no timeout", i.e.: "wait forever".
The `watch()` function, if given a zero value, time outs "as soon as possible".
This doesn't actually make sense - what is the use-case for such behavior?
Even worse, the actual delay here is not really zero - it is low, but unpredictable due to the nature of the golang/OS scheduler.

**Reasons**
A zero-timeout in this very function already caused problems with randomly failing pipelines: https://github.com/kyma-project/cli/pull/1490
A standard behavior is less surprising for developers, who are used to the already established and well-known contract.

LeelaChacha:
**Description**
Change the meaning of timeout of 0 to no timeout,
i.e. waiting forever.
Closes #1492

### DIFF
--- a/internal/kube/client.go
+++ b/internal/kube/client.go
@@ -406,13 +406,15 @@ func (c *client) DefaultNamespace() string {
 }
 
 // watch provides a unified implementation to watch resources.
-// timeout of zero does NOT mean "no timeout", i.e. "wait forever" - it means the function will time-out almost immediately - at the discretion of the golang scheduler.
+// timeout of zero means "no timeout", i.e. "wait forever"
 func watch(
 	ctx context.Context, res string, watchFn func(ctx context.Context) (bool, error), timeout time.Duration,
 ) error {
-
-	ctx, cancel := context.WithTimeout(ctx, timeout)
-	defer cancel()
+	if timeout.Milliseconds() != 0 {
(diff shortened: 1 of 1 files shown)


LeelaChacha:
I am a little bit unsure about the change as the kube client is used in a lot commands. Please let tme know if this can cause unexpected errors down the line.

lindnerby:
> I am a little bit unsure about the change as the kube client is used in a lot commands. Please let me know if this can cause unexpected errors down the line.

Should be fine, can/did you check if all commands use reasonable default values for timeouts? If yes, lgtm

LeelaChacha:
> > I am a little bit unsure about the change as the kube client is used in a lot commands. Please let me know if this can cause unexpected errors down the line.
> 
> Should be fine, can/did you check if all commands use reasonable default values for timeouts? If yes, lgtm

Yes all alpha commands have a non-zero timeout. Looks good

lindnerby (approved):

PR #1575, opened 2023-03-03
chore: bump golang version

lindnerby:
**Description**

- golang bump should fix https://github.tools.sap/kyma/security-scans/issues/24278

**Related issue(s)**

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -14,7 +14,7 @@ FLAGS = -ldflags '-s -w -X github.com/kyma-project/cli/cmd/kyma/version.Version=
 
 .PHONY: resolve
 resolve:
-	go mod tidy -compat=1.18
+	go mod tidy -compat=1.19
 
 .PHONY: validate
 validate:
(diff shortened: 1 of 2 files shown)


jakobmoellersap (changes requested):
Also the combat make `resolve` target should be updated, and also we need to verify compatibility with legacy reconciler first. In addition, the linked issues will only be solved if the correct patch is used, not when go.mod is bumped. I think it is based on the go version that is used for building.

lindnerby:
> Also the combat make `resolve` target should be updated, and also we need to verify compatibility with legacy reconciler first. In addition, the linked issues will only be solved if the correct patch is used, not when go.mod is bumped. I think it is based on the go version that is used for building.

So we want to do [this](https://github.com/kyma-incubator/reconciler/pull/1308) for reconciler first? All succeeding jobs are enough to verify that reconciler works?

lindnerby:
> Basically I am saying the `go 1.19` in the go.mod decides on version compatibility with golang, but it does not decide which one is actually used for building (as that would be the go installation used in the build process). WDYT?

You are right and it seem CLI should already be build with 1.20. Gonna close this.

PR #1580, opened 2023-03-09
Default template deployment

lindnerby:
**Description**

- ...
- ...
- ...

**Related issue(s)**

### DIFF
--- a/cmd/kyma/alpha/deploy/opts.go
+++ b/cmd/kyma/alpha/deploy/opts.go
@@ -18,6 +18,7 @@ type Options struct {
 	DryRun              bool
 	OpenDashboard       bool
 	Force               bool
+	Bare                bool
 	CertManagerVersion  string
 	Namespace           string
 	Channel             string
--- a/cmd/kyma/alpha/deploy/opts.go
+++ b/cmd/kyma/alpha/deploy/opts.go
@@ -23,9 +23,10 @@ type Options struct {
 	Namespace           string
 	Channel             string
 	KymaCR              string
+	Target              string
 	Modules             []string
 	Kustomizations      []string
-	Templates           []string
+	AdditionalTemplates []string
 	Timeout             time.Duration
 	Filters             []kio.Filter
 }
--- a/cmd/kyma/alpha/create/module/module.go
+++ b/cmd/kyma/alpha/create/module/module.go
@@ -61,7 +61,7 @@ Build module my-domain/modA in version 1.2.3 and push it to a remote registry
 Build module my-domain/modB in version 3.2.1 and push it to a local registry "unsigned" subfolder without tls
 		kyma alpha create module -n my-domain/modB --version 3.2.1 -p /path/to/module --registry http://localhost:5001/unsigned --insecure
 `,
-		RunE:    func(cobraCmd *cobra.Command, args []string) error { return c.Run(cobraCmd.Context(), args) },
+		RunE:    func(cobraCmd *cobra.Command, args []string) error { return c.Run(cobraCmd.Context()) },
 		Aliases: []string{"mod"},
 	}
 
@@ -77,7 +77,7 @@ Build module my-domain/modB in version 3.2.1 and push it to a local registry "un
 	)
 	cmd.Flags().BoolVar(
 		&o.PersistentArchive, "module-archive-persistence", false,
-		"Uses the host filesystem instead of inmemory archiving to build the module.",
+		"Uses the host filesystem instead of in-memory archiving to build the module.",
 	)
 	cmd.Flags().BoolVar(&o.ArchiveVersionOverwrite, "module-archive-version-overwrite", false, "Overwrites existing component's versions of the module. If set to false, the push is a No-Op.")
 
@@ -134,7 +134,7 @@ Build module my-domain/modB in version 3.2.1 and push it to a local registry "un
 	return cmd
 }
 
-func (cmd *command) Run(ctx context.Context, args []string) error {
+func (cmd *command) Run(ctx context.Context) error {
 	osFS := osfs.New()
 
 	if cmd.opts.CI {
--- a/cmd/kyma/alpha/deploy/cmd.go
+++ b/cmd/kyma/alpha/deploy/cmd.go
@@ -79,7 +79,7 @@ By default, Lifecycle Manager is deployed from the GitHub main branch.`,
 	// TODO remove this flag when module templates can be fetched from release.
 	// Might be worth keeping this flag with another name to install extra templates??
 	cobraCmd.Flags().StringArrayVar(
-		&o.Templates, "templates", []string{modulesKustomization}, `Provide one or more module templates to deploy.
+		&o.Templates, "templates", []string{modulesKustomization}, `Provide one or more extra module templates to deploy.
 WARNING: This is a temporary flag for development and will be removed soon.`,
 	)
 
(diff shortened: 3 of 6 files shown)


cmd/kyma/alpha/deploy/opts.go
```diff
@@ -70,3 +76,11 @@ func (o *Options) validateFilters() error {
 	o.Filters = filters
 	return nil
 }
+
+func (o *Options) validateTarget() error {
+	if o.Target == "control-plane" || o.Target == "remote" {
```
ruanxin, line 81: I suggest we define these two values as const, so that the following returned errors can also reuse, keeping the value consistent.

docs/gen-docs/kyma_alpha_deploy.md
```diff
@@ -24,27 +24,28 @@ kyma alpha deploy [flags]
...
-  -n, --namespace string            The Namespace to deploy the the Kyma custom resource in. (default "kyma-system")
-      --open-dashboard              Opens the Busola Dashboard at startup. Only works when a graphical interface is available and when running in interactive mode
-      --templates stringArray       Provide one or more module templates to deploy.
-                                    WARNING: This is a temporary flag for development and will be removed soon. (default [https://github.com/kyma-project/kyma/modules@main])
-  -t, --timeout duration            Maximum time for the deployment. (default 20m0s)
-      --wildcard-permissions        Creates a wildcard cluster-role to allow for easy local installation permissions of Lifecycle Manager.
-                                    Allows for Lifecycle Manager usage without worrying about modules requiring specific RBAC permissions.
-                                    WARNING: DO NOT USE ON PRODUCTIVE CLUSTERS! (default true)
+  -b, --bare                          Skip applying default module templates after the deployment.
```
ruanxin, line 27: I have a bit of doubt about this `bare` option, it's just for me if not reading the description, don't know what's the meaning, can we change it to something more straightforward? like `skip-templates`. Also, I think this option is not a flag that will often be used, no worth to give a short name.

lindnerby: Don't really mind :) Just took that from the ACs, guess we are free in deciding that? Would rename it to `skip-default-templates` then.

docs/gen-docs/kyma_alpha_deploy.md
```diff
@@ -24,27 +24,28 @@ kyma alpha deploy [flags]
...
-      --open-dashboard              Opens the Busola Dashboard at startup. Only works when a graphical interface is available and when running in interactive mode
-      --templates stringArray       Provide one or more module templates to deploy.
-                                    WARNING: This is a temporary flag for development and will be removed soon. (default [https://github.com/kyma-project/kyma/modules@main])
-  -t, --timeout duration            Maximum time for the deployment. (default 20m0s)
-      --wildcard-permissions        Creates a wildcard cluster-role to allow for easy local installation permissions of Lifecycle Manager.
-                                    Allows for Lifecycle Manager usage without worrying about modules requiring specific RBAC permissions.
-                                    WARNING: DO NOT USE ON PRODUCTIVE CLUSTERS! (default true)
+      --cert-manager string           Installs cert-manager from the specified static version. An empty string skips the installation. (default "v1.11.0")
+  -c, --channel string                Select which channel to deploy from. (default "regular")
```
IwonaLanger, line 28:
```suggestion
  -c, --channel string                Selects which channel to deploy from. (default "regular")
```

docs/gen-docs/kyma_alpha_deploy.md
```diff
@@ -24,27 +24,28 @@ kyma alpha deploy [flags]
...
+      --dry-run                       Renders the Kubernetes manifests without actually applying them.
+      --extra-templates stringArray   Provide one or more additional module templates via URL or local path to apply after deployment.
```
IwonaLanger, line 30:
```suggestion
      --extra-templates stringArray   Provides one or more additional module templates via URL or local path to apply after deployment.
```

docs/gen-docs/kyma_alpha_deploy.md
```diff
@@ -24,27 +24,28 @@ kyma alpha deploy [flags]
...
+  -f, --force-conflicts               Forces the patching of Kyma spec modules in case their managed field was edited by a source other than Kyma CLI.
+  -k, --kustomization stringArray     Provides one or more kustomizations to deploy. 
+                                      Each flag occurrence accepts a URL with an optional reference (commit, branch, or release) in URL@ref format or a local path to the directory of the kustomization file.
+                                      By default, Lifecycle Manager is deployed from the GitHub main branch. (default [https://github.com/kyma-project/lifecycle-manager/config/default])
+      --kyma-cr string                Provide a custom Kyma CR file for the deployment.
```
IwonaLanger, line 35:
```suggestion
      --kyma-cr string                Provides a custom Kyma CR file for the deployment.
```

docs/gen-docs/kyma_alpha_deploy.md
```diff
@@ -24,27 +24,28 @@ kyma alpha deploy [flags]
...
+      --lifecycle-manager string      Installs Lifecycle Manager with the specified image:
+                                      - Use "my-registry.org/lifecycle-manager:my-tag"" to use a custom version of Lifecycle Manager.
+                                      - Use "europe-docker.pkg.dev/kyma-project/prod/lifecycle-manager@sha256:cb74b29cfe80c639c9ee9..." to use a custom version of Lifecycle Manager with a digest.
+                                      - Specify a tag to override the default one. For example, when specifying "v20230220-7b8e9515",  the "eu.gcr.io/kyma-project/lifecycle-manager:v20230220-7b8e9515" tag is used. (default "eu.gcr.io/kyma-project/lifecycle-manager:latest")
+  -m, --module stringArray            Provide one or more modules to activate after the deployment is finished. Example: "--module name@namespace" (namespace is optional).
```
IwonaLanger, line 40:
```suggestion
  -m, --module stringArray            Provides one or more modules to activate after the deployment is finished. Example: "--module name@namespace" (namespace is optional).
```

docs/gen-docs/kyma_alpha_deploy.md
```diff
@@ -24,27 +24,28 @@ kyma alpha deploy [flags]
...
+  -n, --namespace string              The Namespace to deploy the the Kyma custom resource in. (default "kyma-system")
```
IwonaLanger, line 41:
```suggestion
  -n, --namespace string              The Namespace to deploy the Kyma custom resource in. (default "kyma-system")
```

docs/gen-docs/kyma_alpha_deploy.md
```diff
@@ -24,27 +24,28 @@ kyma alpha deploy [flags]
...
+      --open-dashboard                Opens the Busola Dashboard at startup. Only works when a graphical interface is available and when running in interactive mode
+  -s, --skip-default-templates        Skip applying default module templates after the deployment.
```
IwonaLanger, line 43:
```suggestion
  -s, --skip-default-templates        Skips applying default module templates after the deployment.
```

docs/gen-docs/kyma_alpha_deploy.md
lindnerby: @IwonaLanger Are you sure about the tense here? The description is formulated as that the user has to provide one ore more module templates to use this flag. The flag itself does not actually provide module templates. How can we formulate it better?

lindnerby: Same for the other flag descriptions that start with _Provide_

IwonaLanger: Ah, I see. In that case my suggestions are irrelevant and everything is perfectly fine here. Sorry!

IwonaLanger: If there's a possibility that the user might get it wrong (as I did), you could rephrase it like this: `You must provide...`. This will make a clear difference between a description and an instruction.

IwonaLanger (approved):

jakobmoellersap (approved):

PR #1583, opened 2023-03-10
cmd alpha enable module: Add Channel Input Validation

LeelaChacha:
**Description**

Changes proposed in this pull request:
- Channel is validated by checking against all available ModuleTemplates for the module to be enabled.

**Related issue(s)**

Closes #1577

PR #1587, opened 2023-03-13
Use the Kyma global channel for disable module command

### LINKED ISSUE #1576: kyma disable module command can't handle module use global channel
**Description**

in disable module command, if module defined in kyma use global level channel,
```
spec:
    channel: alpha
    modules:
    - customResourcePolicy: CreateAndDelete
      name: keda
```

**Expected result**

 the disable command use channel flag should possible disable modules which defined in global level channel

**Actual result**

the disable command use channel flag will report not found error:
```
kyma alpha disable module keda -c alpha -n kcp-system -k default-kyma
```
```
Error: could not disable module as it was not found: keda in channel alpha
```
**Steps to reproduce**

**Troubleshooting**

nesmabadr:
**Description**

- Use the Kyma global channel if the module doesn't overwrite it for the `disable module` command

**Related issue(s)**
Fixes #1576


cmd/kyma/alpha/disable/module/module.go
```diff
@@ -157,7 +157,7 @@ func disableModule(modules []v1beta1.Module, name, channel string) ([]v1beta1.Mo
 			continue
 		}
 		if channel != "" {
-			if mod.Channel != channel {
+			if mod.Channel != "" && mod.Channel != channel {
```
jakobmoellersap, line 160: if the module was found but not in the matching channel, maybe we should warn the user then or log this scenario. WDYT?

nesmabadr: @jakobmoellersap You mean warn the user that we are disabling the module from the global Kyma channel? because here the global channel is only considered IF the module doesn't have a channel that overwrites it.

ruanxin, line 160: 1. It's better to have a unit test for this disableModule function.
2.Just want to point out, the validation of the given `channel` is a valid `kyma.spec.channel` is missing, however, it's being implemented in this [PR](https://github.com/kyma-project/cli/pull/1583/files#diff-8c9ab491077df3621b75d4945099d239b8623f5d657388b8b698781ac05cc5d5R138-R143)

cmd/kyma/alpha/disable/module/module_test.go
```diff
@@ -0,0 +1,118 @@
+package module
```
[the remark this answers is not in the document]
nesmabadr: Didn't know about that, thanks!

ruanxin (approved):

Issue #1472, opened 2022-12-06
Detect a managed Kyma instance 


Tomasz-Smelcerz-SAP:
**Description**

the https://github.com/kyma-project/cli/issues/1360 introduced a way to detect a managed cluster. The idea was to display a warning if the user tries to deploy/undeploy kyma in a managed environment using the CLI.

The proposed solution failed: https://github.com/kyma-project/cli/issues/1460

It looks like instead of detecting a _managed cluster_, the CLI should try to detect a _managed Kyma instance._
The reasons for that are:
- detecting a "cluster type" (managed/not managed) is quite hard.
- It may change with time as the underlying infrastructure changes - our solution will break silently.
- it may be even impossible in the future, with the support for "Bring Your Own Cluster" model.

What if we don't need to detect "cluster type" at all?
All we need is to check if there is a _managed kyma instance_ installed (or to-be-installed) on the cluster.
- If yes, we display a warning.
- If no, i.e: the cluster is empty or there is a _manually installed Kyma instance_ running, we can proceed.

How to detect manually installed Kyma instance?
I have two ideas here:

The first approach is to mark some global objects in the cluster with the help of the central components that mange Kyma instances running in the control-plane.
An example of that would be adding annotations to the *kyma-system* namespace by the reconciler and/or annotating *Kyma* objects managed by lifecycle manager.
The model requires adding the same logic to several distinct components: reconciler, lifecycle-manger, CLI
This increases the maintenance effort

The second approach is to assume that, by default, **all** instances are managed unless explictily marked as "manually manged" by the CLI. In that case it is the CLI that is directly on indirectly adding some "special" annotations during Kyma installation.
This model requires adding the logic only to one component: CLI.

tobiscr:
The KCP reconciler should inject an annotation to the resources indicating that this resource is centrally managed by KCP and should not be touched by by a different tool.

When using server-side-apply, we could set a unique `manager` for the resources which indicates the it's a managed resources from KCP (https://kubernetes.io/docs/reference/using-api/server-side-apply/#managers).

Tomasz-Smelcerz-SAP:
@clebs Could you give an example when this CLI-only approach may break? The only scenario that comes to my mind is a corner case, where user connects to a manged cluster just-in-time before KCP started Kyma installation. The CLI would then see an empty cluster, and would not show the warning. The user starts the Kyma installation using the CLI and 30 seconds later KCP starts managed Kyma installation on the same cluster.
[...]

PR #1602, opened 2023-03-24
Detect managed kyma

nesmabadr:
**Description**

- For the deploy command, check if there is a managed Kyma deployed
- Based on this check, display a prompt for the user whether to continue with the deployment or not

**Related issue(s)**
#1472

### DIFF
--- a/cmd/kyma/alpha/deploy/cmd.go
+++ b/cmd/kyma/alpha/deploy/cmd.go
@@ -209,12 +209,6 @@ func (cmd *command) deploy(ctx context.Context) error {
 	}
 	clusterAccess.Successf("Successfully connected to cluster")
 
-	if !cmd.opts.CI && !cmd.opts.NonInteractive {
-		if err := cmd.detectManagedKyma(ctx); err != nil {
-			return err
-		}
-	}
-
 	if !cmd.opts.Verbose {
 		stderr := os.Stderr
 		os.Stderr = nil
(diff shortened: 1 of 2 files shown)


cmd/kyma/alpha/deploy/cmd.go
Tomasz-Smelcerz-SAP, line 1: The `deploy` command is not the only one affected, the `undeploy` is another one.

cmd/kyma/alpha/deploy/cmd.go
```diff
@@ -208,6 +209,12 @@ func (cmd *command) deploy(ctx context.Context) error {
 	}
 	clusterAccess.Successf("Successfully connected to cluster")
 
+	if !cmd.opts.CI && !cmd.opts.NonInteractive {
```
Tomasz-Smelcerz-SAP, line 212: The legacy logic for detecting managed runtime is already configured in line 129. At the time the CLI should use both approaches. In the future we'll remove the legacy `reconciler-based` approach.

nesmabadr: @Tomasz-Smelcerz-SAP this is the `Kyma alpha deploy` not the `Kyma deploy` command

Tomasz-Smelcerz-SAP: You are right. I've looked up in the `deploy` command file.
I think it's correct.

Tomasz-Smelcerz-SAP (approved):

PR #1614, opened 2023-03-30
Remove homemade lint script

ruanxin:
**Description**
* Since `pull-cli-lint` has been introduced in the pipeline, we can remove our homemade lint script
* Add explicit lint target for local development to lint with latest golangci-lint version
* Fix one lint issue

**Related issue(s)**


hack/verify-lint.sh
```diff
@@ -1,65 +0,0 @@
...
-    typecheck unused \
-    # additional lints
-    revive gofmt misspell gochecknoinits unparam exportloopref gosec
-  )
-
-  ENABLE=$(sed 's/ /,/g' <<< "${LINTS[@]}")
-
-  echo "Checks: ${LINTS[*]}"
-  cd ${ROOT_PATH}
-  golangci-lint --disable-all --enable="${ENABLE}" --timeout=10m run --config $CURRENT_DIR/.golangci.yml
```
lindnerby, line 44: We also need to make sure that we are using the golangci config in the job args

ruanxin: but I moved that `.golangci.yml` to root folder already.

lindnerby (approved):

Issue #1632, opened 2023-04-17
Explicit registry credentials are ignored


Tomasz-Smelcerz-SAP:
**Description**

`kyma alpha create module --registry="..." --credentials="..."` fails when a docker config file exists without an entry for the configured registry.

If you do the `docker logout`, the docker tooling removes a registry entry from the docker config file (`$HOME/.docker/config.json`)
When the ocm libraries used in the CLI are parsing the configuration for the registry, instead of returning `nil` or `error`, an empty `Auth` object is returned instead.
The logic in the CLI incorrectly detects this as a valid Auth configuration and ignores credentials provided explicitly with `--credentials=` flag.

**Expected result**

If the local docker config does not contain an entry for the configured registry, the explicit credentials should be used.

**Actual result**

The push to the registry fails with the following error: `Error: could not finish component transfer: component references for kyma-project.io/template-operator:v0.0.16: failed to update state: unable to add blob: failed to push: server message: insufficient_scope: authorization failed`

**Steps to reproduce**

`kyma alpha create module -v --version v0.0.16 --registry registry-1.docker.io/someuser --name-mapping=sha256-digest -p /path/to/the/module/directory --credentials="user:password"`

**Troubleshooting**

Delete the local docker config file: `$HOME/.docker/config.json`

PR #1634, opened 2023-04-17
Configure creds for empty auth

Tomasz-Smelcerz-SAP:
**Description**

Fixes: #1632 

- Detect when an empty Auth object is created from local docker configuration
- Use explicit credentials for the empty Auth.

**Related issue(s)**
#1632

### DIFF
--- a/pkg/module/remote.go
+++ b/pkg/module/remote.go
@@ -7,7 +7,6 @@ import (
 	"regexp"
 	"strings"
 
-	"github.com/containerd/containerd/log"
 	"github.com/open-component-model/ocm/pkg/common"
 	"github.com/open-component-model/ocm/pkg/contexts/credentials"
 	"github.com/open-component-model/ocm/pkg/contexts/credentials/repositories/dockerconfig"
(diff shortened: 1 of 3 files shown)


pkg/module/remote.go
```diff
@@ -103,6 +104,19 @@ func (r *Remote) getCredentials(ctx cpi.Context) credentials.Credentials {
 	return creds
 }
 
+// See: github.com/open-component-model/ocm/pkg/contexts/credentials/repositories/dockerconfig/repository.go#IsEmptyAuthConfig()
+func isEmptyAuth(creds credentials.Credentials) bool {
+
+	if len(creds.GetProperty("auth")) != 0 {
+		return false
+	}
+	if len(creds.GetProperty("username")) != 0 {
```
lindnerby, line 113: Just wondering if it could happen that `username` property is empty while `auth` is not? Do you think it makes sense to rework this check into returning true if `auth` or `username` is empty rather than returning early if only `auth` is not? WDYT?

lindnerby (approved):

PR #1663, opened 2023-05-12
Control in-kcp-mode flag using target flag

nesmabadr:
**Description**

- Patch `--in-kcp-mode` during LM deployment if `target` passed to `Kyma alpha deploy` is set to `control-plane`
- Added unit test

**Related issue(s)**
Resolves https://github.com/kyma-project/lifecycle-manager/issues/575

### DIFF
--- a/cmd/kyma/alpha/deploy/cmd.go
+++ b/cmd/kyma/alpha/deploy/cmd.go
@@ -185,8 +185,12 @@ func (cmd *command) runWithTimeout(ctx context.Context) error {
 }
 
 func (cmd *command) deploy(ctx context.Context) error {
+	isInKcpMode := false
+	if cmd.opts.Target == targetControlPlane {
+		isInKcpMode = true
+	}
 	if cmd.opts.DryRun {
-		return cmd.dryRun(ctx)
+		return cmd.dryRun(ctx, isInKcpMode)
 	}
 
 	start := time.Now()
@@ -221,7 +225,7 @@ func (cmd *command) deploy(ctx context.Context) error {
 	deployStep := cmd.NewStep("Deploying Kustomizations")
 	deployStep.Start()
 	hasKyma, err := deploy.Bootstrap(
-		ctx, cmd.opts.Kustomizations, cmd.K8s, cmd.opts.Filters, cmd.opts.WildcardPermissions, cmd.opts.Force, false,
+		ctx, cmd.opts.Kustomizations, cmd.K8s, cmd.opts.Filters, cmd.opts.WildcardPermissions, cmd.opts.Force, false, isInKcpMode,
 	)
 	if err != nil {
 		deployStep.Failuref("Failed to deploy Kustomizations %s: %s", cmd.opts.Kustomizations, err.Error())
@@ -292,7 +296,7 @@ func (cmd *command) deployCertManager(ctx context.Context) {
 	)
 }
 
-func (cmd *command) dryRun(ctx context.Context) error {
+func (cmd *command) dryRun(ctx context.Context, isInKcpMode bool) error {
 	if cmd.opts.CertManagerVersion != "" {
 		if err := deploy.CertManager(ctx, cmd.K8s, cmd.opts.CertManagerVersion, cmd.opts.Force, true); err != nil {
 			return err
@@ -300,7 +304,7 @@ func (cmd *command) dryRun(ctx context.Context) error {
 	}
 
 	hasKyma, err := deploy.Bootstrap(
-		ctx, cmd.opts.Kustomizations, cmd.K8s, cmd.opts.Filters, cmd.opts.WildcardPermissions, cmd.opts.Force, true,
+		ctx, cmd.opts.Kustomizations, cmd.K8s, cmd.opts.Filters, cmd.opts.WildcardPermissions, cmd.opts.Force, true, isInKcpMode,
 	)
 	if err != nil {
 		return err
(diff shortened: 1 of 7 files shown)


cmd/kyma/alpha/deploy/cmd.go
```diff
@@ -185,8 +185,12 @@ func (cmd *command) runWithTimeout(ctx context.Context) error {
 }
 
 func (cmd *command) deploy(ctx context.Context) error {
+	isInKcpMode := false
```
lindnerby, line 188:
```suggestion
	isInKcpMode := cmd.opts.Target == targetRemote
```

internal/deploy/bootstrap.go
```diff
@@ -63,16 +73,59 @@ func Bootstrap(
...
+func PatchDeploymentWithInKcpModeFlag(ctx context.Context, k8s kube.KymaKube, manifestObjs []ctrlClient.Object, isInKcpMode bool) (*appsv1.Deployment, error) {
+	var patchedDeployment *appsv1.Deployment
+	for _, manifest := range manifestObjs {
+		if manifest.GetObjectKind().GroupVersionKind().Kind == "Deployment" {
+			hasKcpFlag := false
+			if manifestObj, success := manifest.(*unstructured.Unstructured); success {
+				manifestJSON, err := json.Marshal(manifestObj.Object)
+				if err == nil {
+					deploymentSpec := appsv1.Deployment{}
+					if err = json.Unmarshal(manifestJSON, &deploymentSpec); err == nil {
```
lindnerby, line 100: This level of nesting is pretty crazy, please just extract some function for the deployment lookup where you used early returns

internal/deploy/apply.go
```diff
@@ -21,28 +22,30 @@ type applyOpts struct {
 	initialBackoff time.Duration
 }
 
-func applyManifests(ctx context.Context, k8s kube.KymaKube, manifests []byte, opts applyOpts) error {
+func applyManifests(ctx context.Context, k8s kube.KymaKube, manifests []byte, opts applyOpts) ([]client.Object, error) {
```
lindnerby, line 25: Can you please split the actual "incremental retry apply" and the parsing of of bytes to manifestObjs? So you can patch the deployment obj in memory already?

lindnerby (changes requested):

internal/deploy/bootstrap.go
```diff
@@ -63,16 +65,92 @@ func Bootstrap(
...
 	}
 
 	return hasKyma(string(manifests))
 }
 
+func PatchDeploymentWithInKcpModeFlag(manifestObjs []ctrlClient.Object, isInKcpMode bool) error {
+	var deployment *appsv1.Deployment
+	for _, manifest := range manifestObjs {
+		manifestJSON, err := getManifestJSONForDeployment(manifest)
+		if err == nil && manifestJSON != nil {
```
lindnerby, line 90: Still a bit too nested within this function body. Can you work with early returns or continues here e.g. if `error != nil` or `manifestJSON == nil`  etc. Same for the rest of the func, and  maybe other funcs in this file

lindnerby (changes requested):

internal/deploy/bootstrap.go
```diff
@@ -63,16 +65,97 @@ func Bootstrap(
...
 
 	return hasKyma(string(manifests))
 }
 
+func PatchDeploymentWithInKcpModeFlag(manifestObjs []ctrlClient.Object, isInKcpMode bool) error {
+	var deployment *appsv1.Deployment
+	for _, manifest := range manifestObjs {
+		manifestJSON, err := getManifestJSONForDeployment(manifest)
+		if err != nil || manifestJSON == nil {
+			return err
```
lindnerby, line 91: If `manifestJSON == nil` is evaluated to true there won't be an error and this function will return nil.

internal/deploy/bootstrap.go
```diff
@@ -63,16 +65,97 @@ func Bootstrap(
...
+			dryRun, force, defaultRetries, defaultInitialBackoff}, manifestObjs,
+	)
+	if err != nil {
 		return false, err
 	}
 
 	return hasKyma(string(manifests))
 }
 
+func PatchDeploymentWithInKcpModeFlag(manifestObjs []ctrlClient.Object, isInKcpMode bool) error {
```
lindnerby, line 86: Can be pkg private

internal/deploy/bootstrap.go
```diff
@@ -63,16 +65,97 @@ func Bootstrap(
...
+		if err != nil || manifestJSON == nil {
+			return err
+		}
+		deployment, err = getDeployment(manifestJSON)
+		if err != nil {
+			return err
+		}
+		deploymentArgs := deployment.Spec.Template.Spec.Containers[0].Args
+		hasKcpFlag := checkDeploymentHasKcpFlag(deploymentArgs)
+		if !hasKcpFlag && isInKcpMode {
```
lindnerby, line 102: If this function is called with `isInKcpMode` false, nothing will happen, so why have this parameter even? The caller should decide to call this if needed.

lindnerby (changes requested):

internal/deploy/bootstrap.go
nesmabadr: yes makes sense, will only call the function if isInKcpMode is true

lindnerby (approved):

PR #1666, opened 2023-05-16
Make v1beta2 default

lindnerby:
**Description**

- Since the API has been upgraded to `v1beta2` for CRDs of Kyma, ModuleTemplate, Manifest and Watcher, we need to draw-in using the latest API version into CLI to the module and kyma delivery alpha commands commands

**Related issue(s)**


internal/deploy/kyma.go
```diff
@@ -33,10 +33,10 @@ func Kyma(
 		kyma.SetName("default-kyma")
 		kyma.SetNamespace(namespace)
 		kyma.SetAnnotations(map[string]string{"cli.kyma-project.io/source": "deploy"})
-		kyma.SetLabels(map[string]string{"operator.kyma-project.io/managed-by": "lifecycle-manager"})
+		kyma.SetLabels(map[string]string{v1beta2.ManagedBy: "lifecycle-manager"})
 		kyma.Spec.Channel = channel
-		kyma.Spec.Sync.Enabled = false
-		kyma.Spec.Modules = []v1beta1.Module{}
+		kyma.SetLabels(map[string]string{v1beta2.SyncLabel: v1beta2.DisableLabelValue})
```
nesmabadr, line 38: Shouldn't this label depend on the `target` flag as well? 
We patch the `in-kcp-mode` during the LM deployment if the target is set to `remote` [here](https://github.com/kyma-project/cli/pull/1663) but then disabling the label is not consistent with this behavior.

lindnerby: True, will wait for https://github.com/kyma-project/cli/pull/1663 to be merged and incorporate usage of `isInKcpMode` then here 👍

lindnerby: Added, please have a look :)

nesmabadr (approved):

PR #1679, opened 2023-06-06
Handle optional serverless dependencies in cli

kwiatekus:
**Description**

- check which modules will be required to apply function, error on missing module only when needed

**Related issue(s)**
https://github.com/kyma-project/serverless-manager/issues/84
Depends on https://github.com/kyma-project/hydroform/pull/413


cmd/kyma/apply/function/function.go
```diff
@@ -103,32 +105,45 @@ func (c *command) Run() error {
...
-	if err != nil {
-		step.Failure()
-		return err
-	}
+	operators := []operator.Operator{}
 
-	kymaAddress, err := c.kymaHostAddress()
-	if err != nil {
-		step.LogErrorf("%s\n%s", err, "Check if your cluster is available and has Kyma installed.")
+	if configuration.Subscriptions != nil {
```
pPrecel, line 110: The definition of the Subscriptions field looks like this:

```go
Subscriptions []Subscription    `yaml:"subscriptions,omitempty"`
```

It does mean that in this case, we should react on non-zero value as well ( not nil but empty map ). Example here:
https://go.dev/play/p/b44TCFzbzNw

```suggestion
	if len(configuration.Subscriptions) != 0 {
```

cmd/kyma/apply/function/function.go
```diff
@@ -103,32 +105,45 @@ func (c *command) Run() error {
...
+		}
+		operators = append(operators, operator.NewSubscriptionOperator(client.Resource(subscriptionGVR).Namespace(configuration.Namespace),
+			configuration.Name, configuration.Namespace, subscriptions...))
 	}
 
-	apiRules, err := resources.NewAPIRule(configuration, kymaAddress)
-	if err != nil {
-		step.Failure()
-		return err
+	if configuration.APIRules != nil {
```
pPrecel, line 128: Same thing as for the `configuration.Subscriptions` field

```suggestion
	if len(configuration.APIRules) != 0 {
```

cmd/kyma/apply/function/function.go
```diff
@@ -103,32 +105,45 @@ func (c *command) Run() error {
...
 	}
 
-	subscriptions, err := resources.NewSubscriptions(configuration)
-	if err != nil {
-		step.Failure()
-		return err
-	}
+	operators := []operator.Operator{}
 
-	kymaAddress, err := c.kymaHostAddress()
```
pPrecel, line 112: So we totally remove this functionality? Am I right?

kwiatekus: Yes. API rules do not require to pass the full runtime domain. The controller is assuming it based on kyma gateway internally

pPrecel (dismissed):

pPrecel (approved):

Tomasz-Smelcerz-SAP (approved):

PR #1692, opened 2023-06-27
module command: Add Error Code for Kyma Warning State

### LINKED ISSUE #1683: Adjust exit codes for the `Warning` state
**Description**

The new introduced `Warning` state of Kyma is treated by the CLI like an error and leads to an exit-code 1.

Unfortunately, the Kyma module release pipeline relies on the exit-code of the CLI to determine whether a Kyma module was properly installed.

The `Warning` state is indicating that a module is correctly installed, but cannot work properly caused by missing configuration parameters (or other required pre-requisites).

We should agree how the CLI will reflect the `Warning` state in the return-code. I would see these two options:

1. `Warning` is handled like `Ready` and leads to an exit-code 0 by the CLI
2. `Warning` is handle individually by the CLI and returns a dedicated exit code, e.g.
    1. `Ready` = exit code 0
    2. `Warning` = exit code 2 (or any other preferred digit >1)
    3. `Error` = exit code 1

Please agree how the CLI will handle the `Warning` state on the exit-code level and inform @the1bit about your agreement. Finally, please update the CLI implementation to respond with the appropriate exit-code if Kyma ends up in a `Warning` state.

**Expected result**

It's possible to detect the Kyma `Warning` state by evaluating the exit-code of the Kyma CLI.

**Actual result**

Kyma CLI returns always exit code 0 if the Kyma state is not `Ready` (or `Error`).

**Steps to reproduce**

Install BTP-operator module via the Kyma CLI. This leads to a Kyma instance in `Warning` state.

**Troubleshooting**

LeelaChacha:
Closes #1683

### DIFF
--- a/cmd/kyma/alpha/disable/module/module.go
+++ b/cmd/kyma/alpha/disable/module/module.go
@@ -139,10 +139,8 @@ func (cmd *command) run(ctx context.Context, l *zap.SugaredLogger, moduleName st
 
 	if cmd.opts.Wait {
 		waitStep := cmd.NewStep("Waiting for Kyma to become Ready")
-		if err = moduleInteractor.WaitUntilReady(
-			ctx,
-		); err != nil {
-			waitStep.Failuref("Kyma did not get Ready")
+		if err = moduleInteractor.WaitUntilReady(ctx); err != nil {
+			waitStep.Failuref("kyma did not get ready: %s", err)
 			return err
 		}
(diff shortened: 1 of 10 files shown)


LeelaChacha:
Skipping e2e test as mentioned in the AC

lindnerby (approved):
lgtm 👍

PR #1707, opened 2023-07-19
Enable type assertion linter

lindnerby:
**Description**

- Bump golangci-lint version
- Add checks for type assertions

**Related issue(s)**
* https://github.com/kyma-project/lifecycle-manager/issues/694


cmd/kyma/alpha/list/module/module.go
```diff
@@ -214,12 +214,16 @@ func (cmd *command) printKymaActiveTemplates(ctx context.Context, kyma *unstruct
...
+		annotations := tpl.GetAnnotations()
+		if annotations == nil {
+			annotations = make(map[string]string)
 		}
-		anns["state.cmd.kyma-project.io"] = item["state"].(string)
-		tpl.SetAnnotations(anns)
+		stateValue, ok := item["state"].(string)
+		if !ok {
+			continue
+		}
```
Tomasz-Smelcerz-SAP, line 224: It looks here you're using type assertion to verify if the map entry exists... If it doesn't, `item["state"]` resolves to `nil` and then the type assertion fails. IMO this is slower and less readable than explicit handling, like this:
```suggestion
		stateValue, exists := item["state"]
		if !exists {
			continue
		}
		stateValueString, ok := stateValue.(string)
		if !ok {
			continue
		}
```

pkg/errs/common.go
```diff
@@ -0,0 +1,5 @@
+package errs
+
+import "errors"
+
+var ErrTypeAssertKyma = errors.New("raw can not be cast to v1beta2.Kyma")
```
Tomasz-Smelcerz-SAP, line 5: Perhaps just a: `not a v1beta2.Kyma type` ?
```suggestion
var ErrTypeAssertKyma = errors.New("not a v1beta2.Kyma type")
```

Tomasz-Smelcerz-SAP:
I have found two more places without proper type assertion handling:
```
cmd/kyma/alpha/list/module/module.go
195:            item, _ := statusItems[i].(map[string]interface{})
```

```
internal/deploy/bootstrap_test.go
115:                    manifestJSON, _ := json.Marshal(obj.(*unstructured.Unstructured).Object)
```

Tomasz-Smelcerz-SAP (changes requested):
I have found two places where type assertion is not handled properly, take a look at my comments.

lindnerby:
> I have found two more places without proper type assertion handling:
> 
> ```
> cmd/kyma/alpha/list/module/module.go
> 195:            item, _ := statusItems[i].(map[string]interface{})
> ```
> 
> ```
> internal/deploy/bootstrap_test.go
> 115:                    manifestJSON, _ := json.Marshal(obj.(*unstructured.Unstructured).Object)
> ```

Thanks for the findings. Pushed a change for these as well. Please have a look

pkg/errs/common.go
lindnerby: Changed the error

cmd/kyma/alpha/list/module/module.go
lindnerby: Nice find, split

PR #1719, opened 2023-07-31
Enable cache for the setup-go action

lindnerby:
**Description**

- It's the default settings

**Related issue(s)**


lindnerby:
Disabling cache on the setup-go action seems to cause problems for the golangci-lint action seems to cause problems.

PR #1739, opened 2023-08-21
WIP: Use simple mode as default

jeremyharisch:
**Description**

- Introduce new flag `--kubebuilder-project`, which needs to be provided it module should be build form kubebuilder project instead simple mode
- Simple-mode is now default
- Errors if config-file is not provided in simple-mode

**Related issue(s)**


nesmabadr:
I believe this will break the neighbors pipeline, so we need to communicate this change to them

nesmabadr (approved):

NHingerl (approved):

PR #1746, opened 2023-08-24
Better User-Feedback: Fail the sec scan info step if error appears

jeremyharisch:
**Description**

- If the security config file cannot be found, everything was skipped and no feedback was given to the user, thus quite confusing why this whole process was "skipped"
- Now, the step `Configuring security scanning...` is starting before the security file is fetched, thus when the file is not found, the user will get the feedback that it failed -> Better Userfeedback

Before this change:
Screenshot 2023-08-24 at 15 19 57

After this change:
Screenshot 2023-08-24 at 15 14 44
**Related issue(s)**

### DIFF
--- a/cmd/kyma/alpha/create/module/module.go
+++ b/cmd/kyma/alpha/create/module/module.go
@@ -299,7 +299,7 @@ func (cmd *command) Run(ctx context.Context) error {
 			}
 			cmd.CurrentStep.Successf("Security scanning configured")
 		} else {
-			l.Errorf("Security scanning configuration was skipped: %s", err.Error())
+			l.Warnf("Security scanning configuration was skipped: %s", err.Error())
 			cmd.CurrentStep.Failure()
 		}
 	}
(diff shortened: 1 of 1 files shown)


cmd/kyma/alpha/create/module/module.go
```diff
@@ -299,7 +299,8 @@ func (cmd *command) Run(ctx context.Context) error {
 			}
 			cmd.CurrentStep.Successf("Security scanning configured")
 		} else {
-			l.Warnf("Security scanning configuration was skipped: %s", err.Error())
+			l.Errorf("Security scanning configuration was skipped: %s", err.Error())
```
ruanxin, line 302: can we change this back to `Warnf`? in fact it's not an error.

medmes:
Tested and it works as expected!
Maybe The message should be a hint warnng not an error. If it's an error the command should exit with a status 1.

ruanxin (approved):

PR #1749, opened 2023-08-25
Add the option to override default namespace

### LINKED ISSUE #1727: Add a possibility to override the default namespace
**Description**

Creating a module results in a template yaml being built with the `kcp-system` namespace set as default.

**Expected result**

There should be a possibility to override the namespace for the command (use the config file).

**Actual result**

`kcp-system` is the namespace chosen by default.

**Steps to reproduce**

1. Use the `kyma alpha create module` command with the metadata file as an input
2. The generated Module Template has `namespace: kcp-system`

nesmabadr:
**Description**

- Add the option to override the namespace using the config file with the Create Module command

**Related issue(s)**
Resolves #1727

### DIFF
--- a/cmd/kyma/alpha/create/module/module.go
+++ b/cmd/kyma/alpha/create/module/module.go
@@ -378,9 +378,6 @@ func (cmd *command) Run(ctx context.Context) error {
 		if modCnf != nil && modCnf.Namespace != "" {
 			namespace = modCnf.Namespace
 		}
-		if namespace == "" {
-			namespace = kcpSystemNamespace
-		}
 
 		t, err := module.Template(componentVersionAccess, resourceName, namespace, channel, modDef.DefaultCR, labels, annotations)
 
--- a/cmd/kyma/alpha/create/module/module.go
+++ b/cmd/kyma/alpha/create/module/module.go
@@ -211,6 +211,8 @@ type validator interface {
 	Run(ctx context.Context, log *zap.SugaredLogger) error
 }
 
+const kcpSystemNamespace = "kcp-system"
+
 func (cmd *command) Run(ctx context.Context) error {
 	osFS := osfs.New()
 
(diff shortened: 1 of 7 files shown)


cmd/kyma/alpha/create/module/moduleconfig.go
```diff
@@ -28,6 +29,7 @@ const (
 	//taken from "github.com/open-component-model/ocm/resources/component-descriptor-v2-schema.yaml"
 	moduleNamePattern = "^[a-z][-a-z0-9]*([.][a-z][-a-z0-9]*)*[.][a-z]{2,}(/[a-z][-a-z0-9_]*([.][a-z][-a-z0-9_]*)*)+$"
 	moduleNameMaxLen  = 255
+	namespaceMaxLen   = 253
```
lindnerby, line 32: Let's be consistent and extract the namespace regex pattern below as well into const?

pkg/module/template.go
```diff
@@ -41,7 +41,7 @@ spec:
 `
 )
 
-func Template(remote ocm.ComponentVersionAccess, moduleTemplateName, channel string, data []byte, labels, annotations map[string]string) ([]byte, error) {
+func Template(remote ocm.ComponentVersionAccess, moduleTemplateName, namespace string, channel string, data []byte, labels, annotations map[string]string) ([]byte, error) {
```
lindnerby, line 44: Can we start to create some unit tests for this func? Should be nice and easy to make it table driven?

lindnerby (changes requested):

cmd/kyma/alpha/create/module/module.go
```diff
@@ -211,6 +211,8 @@ type validator interface {
 	Run(ctx context.Context, log *zap.SugaredLogger) error
 }
 
+const kcpSystemNamespace = "kcp-system"
```
lindnerby, line 214: I think also need to provide the namespace flag to the cmd itself right? And there you can set the default value

lindnerby (changes requested):

mmitoraj (approved):

Issue #1723, opened 2023-08-09
Remove the git requirement for create module command


janmedrek:
**Description**

Right now, the alpha create module command uses the `.git` directory to import sources description for the ModuleTemplate, this is implicit and not passed as an argument/metadata field.

The goal is to remove that requirement, it should be optional. The sources part is optional in the OCM libraries.

**Acceptance Criteria**

- [x] Alpha create module command does not require to be ran in the git directory
- [x] With non-git repository, security scan configuration should be ignored 
    - Warn the user by letting fail the step, but not error out
      - in CI mode, print warning message in terminal, but not blocked
      - in non CLI mode, print warning message, block and waiting for user interactive, press `enter` or `y` to continue, `n` to quit. 
    -  Source layer should not be created

medmes:
**PR**: https://github.com/kyma-project/cli/pull/1751

PR #1751, opened 2023-08-27
Allow create module for no git directory

medmes:
**Description**
The purpose of this PR is to remove the git requirement and support local filesysten directory as well.

- Remove and refactor old git package.
- Add source package to implement git source and fs source.
- 

**Related issue(s)**
#1723


internal/files/files.go
```diff
@@ -71,3 +72,28 @@ func FileType(fs vfs.FileSystem, path string) (string, error) {
...
+
+// FindDirectoryContaining returns the path to the directory containing the targetFolderName
+// and a bool indicating if the directory was found or not.
+// If the directory was not found, the error will be os.ErrNotExist.
+// The search starts at the path and continues to the root of the filesystem.
+// The targetFolderName can be a relative path.
+// Example:
+//
+//	FindDirectoryContaining("` + path + `", "` + targetFolderName + `")
+func FindDirectoryContaining(path, targetFolderName string) (string, bool, error) {
```
lindnerby, line 84: The error can already indicates if the directory exists or not. Seems the bool return value is redundant?

pkg/module/build.go
```diff
@@ -79,16 +81,27 @@ func CreateArchive(fs vfs.FileSystem, path string, def *Definition) (*comparch.C
...
+			cd.Sources = append(cd.Sources, *src)
+		} else {
+			cd.Sources[idx] = *src
+		}
 	}
 
 	return nil
 }
+
+func fetchSource(ctx cpi.Context, src source.Source, path, repo, version string) (*ocm.Source, error) {
```
lindnerby, line 101: Don't think you need to wrap this again. This function does nothing else but calling another function.

pkg/module/source/source.go
```diff
@@ -0,0 +1,15 @@
...
+	ocm "github.com/open-component-model/ocm/pkg/contexts/ocm/compdesc"
+	"github.com/open-component-model/ocm/pkg/contexts/ocm/cpi"
+)
+
+const (
+	ocmIdentityName = "module-sources"
+	ocmVersion      = "v1"
+)
+
+type Source interface {
```
lindnerby, line 13: If you want to use an interface here, it should probably be defined by the consumer rather than the implementing package.

internal/files/files.go
```diff
@@ -71,3 +72,28 @@ func FileType(fs vfs.FileSystem, path string) (string, error) {
...
+	if path == string(filepath.Separator) {
+		return "", false, os.ErrNotExist
+	}
+
+	targetPath := filepath.Join(path, targetFolderName)
+	_, err := os.Stat(targetPath)
+	if err != nil {
+		if errors.Is(err, os.ErrNotExist) {
+			return FindDirectoryContaining(filepath.Dir(path), targetFolderName)
```
lindnerby, line 93: Why do we need recursive lookup here? The required path flag of the command already points to either an absolute or relative path or?

pkg/module/git/source.go
```diff
@@ -1,113 +0,0 @@
-package git
```
lindnerby, line 1: Any reason why the package needs to be changed? This code is ok and only needs to be either refactored a bit internally and only be called if module dir contains `.git`

pkg/module/source/git.go
```diff
@@ -0,0 +1,123 @@
...
+	}
+
+	return &ocm.Source{
+		SourceMeta: sourceMeta,
+		Access:     access,
+	}, nil
+}
+
+func (g GitSource) determineRepositoryURL(repo string) (string, error) {
+	if repo == "" {
```
lindnerby, line 72: In general, if you invert conditions like this you can work with early returns to reduce nesting and increase readability/maintainability.

lindnerby (dismissed):

cmd/kyma/alpha/create/module/module.go
```diff
@@ -273,12 +276,36 @@ func (cmd *command) Run(ctx context.Context) error {
...
-	if err != nil {
-		cmd.CurrentStep.Failure()
-		return err
+
+	var archive *comparch.ComponentArchive
+	gitPath, err := files.SearchForTargetDirByName(modDef.Source, ".git")
+	if gitPath == "" || err != nil {
+		l.Warnf("could not find git repository root, using %s directory", modDef.Source)
+		fmt.Println()
+		nice.NewNice().PrintImportant("! CAUTION: The target folder is not a git repository. Therefore, the security scan will not be performed and the module sources will not be added to the layer")
```
ruanxin, line 285: security scan config file in this step has not been detected yet, We don't know if user provide this config file, for open source users, they don't need security scan file to generate module. 

The condition to have this message is when CLI detect is not git repo **and** security scan config file exists.

ruanxin (changes requested):

cmd/kyma/alpha/create/module/module.go
```diff
@@ -320,7 +323,7 @@ func (cmd *command) Run(ctx context.Context) error {
 	// Security Scan
 	if cmd.opts.SecurityScanConfig != "" && gitPath != "" { // security scan is only supported for target git repositories
 		cmd.NewStep("Configuring security scanning...")
-		if _, err = osFS.Stat(cmd.opts.SecurityScanConfig); err == nil {
+		if err = files.IsFileExists(cmd.opts.SecurityScanConfig); err == nil {
```
ruanxin, line 326: This function I believe can't handle relative path, say if I provide `../btp-manager/sec-scanners-config.yaml`, it complains file not exists

cmd/kyma/alpha/create/module/module.go
```diff
@@ -273,12 +276,39 @@ func (cmd *command) Run(ctx context.Context) error {
...
-		return err
+
+	var archive *comparch.ComponentArchive
+	gitPath, err := files.SearchForTargetDirByName(modDef.Source, ".git")
+	if gitPath == "" || err != nil {
+		l.Warnf("could not find git repository root, using %s directory", modDef.Source)
+		n := nice.NewNice()
+		n.PrintImportant("! CAUTION: The target folder is not a git repository. The sources will be not added to the layer")
+		if err = files.IsFileExists(cmd.opts.SecurityScanConfig); err == nil {
+			n.PrintImportant("  The security scan configuration file was not found. The security scan will be skipped")
```
ruanxin, line 287: This message is wrong, in this condition, security scan configuration file exists, but the message shows "was not found"

ruanxin (dismissed):
Please check my comments, still exist two problems, hope it's clear after our video call.

cmd/kyma/alpha/create/module/module.go
```diff
@@ -273,12 +276,39 @@ func (cmd *command) Run(ctx context.Context) error {
...
-		return err
+
+	var archive *comparch.ComponentArchive
+	gitPath, err := files.SearchForTargetDirByName(modDef.Source, ".git")
+	if gitPath == "" || err != nil {
+		l.Warnf("could not find git repository root, using %s directory", modDef.Source)
+		n := nice.NewNice()
+		n.PrintImportant("! CAUTION: The target folder is not a git repository. The sources will be not added to the layer")
+		if files.IsFileExists(cmd.opts.SecurityScanConfig) {
+			n.PrintImportant("  The security scan configuration file was provided. But the security scan will be skipped")
```
ruanxin, line 287:
```suggestion
			n.PrintImportant("  The security scan configuration file has been provided, but it will be skipped due to the absence of repository information.")
```

cmd/kyma/alpha/create/module/module.go
```diff
@@ -273,12 +276,39 @@ func (cmd *command) Run(ctx context.Context) error {
...
-	if err != nil {
-		cmd.CurrentStep.Failure()
-		return err
+
+	var archive *comparch.ComponentArchive
+	gitPath, err := files.SearchForTargetDirByName(modDef.Source, ".git")
+	if gitPath == "" || err != nil {
+		l.Warnf("could not find git repository root, using %s directory", modDef.Source)
+		n := nice.NewNice()
+		n.PrintImportant("! CAUTION: The target folder is not a git repository. The sources will be not added to the layer")
```
jeremyharisch, line 285: No need to use nice. Please log a Warning - but for user interaction please fail the step using `.Failuref("<some message to the user>")`

internal/nice/nice.go
```diff
@@ -12,6 +12,11 @@ type Nice struct {
 	NonInteractive bool
 }
 
+// NewNice returns a new Nice instance
+func NewNice() *Nice {
+	return &Nice{}
+}
+
```
jeremyharisch, line 19: Not needed as stated in other comments
```suggestion

```

cmd/kyma/alpha/create/module/module.go
```diff
@@ -273,12 +276,39 @@ func (cmd *command) Run(ctx context.Context) error {
...
+		if files.IsFileExists(cmd.opts.SecurityScanConfig) {
+			n.PrintImportant("  The security scan configuration file has been provided, but it will be skipped due to the absence of repository information.")
```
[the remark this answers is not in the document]
ruanxin: This information is useful to indicate to users that if they provide a security scan config, it will be skipped, because they don't know that no sources means can't add security scan information.

cmd/kyma/alpha/create/module/module.go
ruanxin: I think using Nice here is good, please check `PrintImportant`, for interactive mode, it changes the message colour to better interact with user, it should be the default way if waiting for user input is needed.

But the problem here is this  `NewNice()` lack of initialised `NonInteractive` field, which is bad, I refactored in later commit.

internal/nice/nice.go
ruanxin: as mentioned above, the problem here is this `NewNice()` lack of initialised `NonInteractive` field.

jeremyharisch (approved):

PR #1758, opened 2023-09-01
upgrade go mod and smoke test for alpha deploy

ruanxin:
- upgraded dependencies to use latest lifecycle-manager, hydroform
- upgraded kustomize
- regular upgrade to latest k8s.io
- add smoke test for alpha deploy command


internal/deploy/bootstrap.go
```diff
@@ -70,6 +71,7 @@ func Bootstrap(
 	if err != nil {
 		return false, err
 	}
+	manifestObjs = filterCRD(manifestObjs)
```
Tomasz-Smelcerz-SAP, line 74: `filter` functions are usually used to return a subset of some collection. Here we're modifying objects. Perhaps `fixTimestamps` or something similar is a better choice.

Tomasz-Smelcerz-SAP (approved):

PR #1768, opened 2023-09-13
use ociblob instead of ocilocalblob for oci v2

ruanxin:
ocilocalblob are marked as [deprecated](https://github.com/open-component-model/ocm/blob/53d94beeac09abb4d35c3780b3f21636296b4224/pkg/contexts/ocm/accessmethods/localociblob/method.go#L33) and offically removed in 0.3.0

In order to prepare for ocm upgrade from 0.3.0-rc1 to 0.3.0, we should stop generating module template with v2 schema using `ocilocalblob`, instead of using `ociblob`.

v2 schema support [ociblob](https://ocm.software/spec/schema-v2.html#component_resources_items_anyOf_i0_access)

Changes in this PR:
- refactor code to use `ociblob`
- remove `compatattr.Set` related code to remove compatible for legacy format
- add e2e test to verify new format
- the new format are verified in LM smoke testing https://github.com/kyma-project/lifecycle-manager/pull/862

 
related issue https://github.com/kyma-project/lifecycle-manager/issues/841


tests/e2e/kyma_create_module_test.go
```diff
@@ -0,0 +1,70 @@
...
+	assert.Contains(t, localblobAccessSpec.LocalReference, "sha256:")
+
+	// test descriptor.component.sources
+	assert.Equal(t, len(descriptor.Sources), 1)
+	source := descriptor.Sources[0]
+	sourceAccessSpec, err := ocm.DefaultContext().AccessSpecForSpec(source.Access)
+	assert.Nil(t, err)
+	githubAccessSpec, ok := sourceAccessSpec.(*github.AccessSpec)
+	assert.Equal(t, githubAccessSpec.Type, github.Type)
+	//assert.Equal(t, githubAccessSpec.RepoURL, testRepoURL)
```
ruanxin, line 69: To Reviewer:

This test condition can be uncommitted when this [PR](https://github.com/kyma-project/cli/pull/1770) get merged

pkg/module/build.go
```diff
@@ -63,14 +57,10 @@ func CreateArchive(fs vfs.FileSystem, path, gitRemote string, def *Definition, i
 		return nil, err
 	}
 
-	if compatattr.Get(ctx) {
-		cd.Provider = v1.Provider{Name: "internal"}
-	} else {
-		cd.Provider = v1.Provider{Name: "kyma-project.io", Labels: v1.Labels{*builtByCLI}}
-	}
+	cd.Provider = v1.Provider{Name: "kyma-project.io", Labels: v1.Labels{*builtByCLI}}
```
Tomasz-Smelcerz-SAP, line 60: What was the role of this `internal` Provider?

pkg/module/init.go
```diff
@@ -4,16 +4,15 @@ import (
 	"github.com/open-component-model/ocm/pkg/contexts/oci"
 	"github.com/open-component-model/ocm/pkg/contexts/oci/repositories/ocireg"
 	"github.com/open-component-model/ocm/pkg/contexts/ocm"
-	"github.com/open-component-model/ocm/pkg/contexts/ocm/compdesc"
-	"github.com/open-component-model/ocm/pkg/contexts/ocm/compdesc/versions/ocm.software/v3alpha1"
+	// nolint:revive
+	_ "github.com/open-component-model/ocm/pkg/contexts/ocm/compdesc/versions"
```
Tomasz-Smelcerz-SAP, line 8: This import registers two schemes: `v3alpha1` and `v2`:
```
import (
        _ "github.com/open-component-model/ocm/pkg/contexts/ocm/compdesc/versions/ocm.software/v3alpha1"
        _ "github.com/open-component-model/ocm/pkg/contexts/ocm/compdesc/versions/v2"
)
```
The existing code registered just `v3alpha1`, in the removed line number 16.
It's the same in the ocm library version `v0.3.0`
Maybe it's safer to do unnamed import of just `v3alpha1`, like this:
```
_ "github.com/open-component-model/ocm/pkg/contexts/ocm/compdesc/versions/ocm.software/v3alpha1"
```
What do you think?

pkg/module/gitsource/git.go
```diff
@@ -8,22 +8,17 @@ import (
...
 	ref, commit, err := g.getGitInfo(path)
 	if err != nil {
 		return nil, err
 	}
 
-	sourceType := "git"
-	if !compatattr.Get(ctx) {
-		sourceType = github.CONSUMER_TYPE
-	}
+	sourceType := github.CONSUMER_TYPE
```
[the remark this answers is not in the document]
ruanxin: The `Type = "gitHub"` is used for [`access.type`](https://github.com/kyma-project/lifecycle-manager/blob/main/tests/smoke_test/module_upgrade/moduletemplate_template_operator_fast.yaml#L51), not for `source.type`, we should not mix it.

I can't find a detail definition of this source.type in ocm document, but by checking the definition of [SourceRef](https://github.com/open-component-model/ocm/blob/v0.3.0/pkg/contexts/ocm/compdesc/componentdescriptor.go#L430-L440), it has `IdentitySelector`, which defines the identity that is used to match a source, so a reasonable design behind this type should be an identity type.

And you can find this CONSUMER_TYPE is [used](https://github.com/open-component-model/ocm/blob/v0.3.0-rc.1/pkg/contexts/ocm/accessmethods/github/method.go#L224-L227) as an identity type in `0.3.0-rc.1`
```
id := credentials.ConsumerIdentity{
identity.ID_TYPE:     CONSUMER_TYPE,
identity.ID_HOSTNAME: hostname,
}
``` 

btw, `Github` as identity type can be found in different [package](https://github.com/open-component-model/ocm/blob/v0.3.0/pkg/contexts/credentials/builtin/github/identity/identity.go#L17) in 0.3.0

Tomasz-Smelcerz-SAP (approved):

PR #1775, opened 2023-09-20
Do not remove `.git` if exists

jeremyharisch:
**Description**
To stick to Github's default, we should not remove the `.git` suffix from the repository URL.

Default schema: `https://github.com/<project name>/<repository name>.git`

For example:
Screenshot 2023-09-20 at 12 16 23

- Do not remove `.git` suffix
- Adapt tests

**Related issue(s)**

### DIFF
--- a/pkg/module/gitsource/git_test.go
+++ b/pkg/module/gitsource/git_test.go
@@ -42,6 +42,14 @@ func Test_fetchRepoURLFromRemotes(t *testing.T) {
 			want:    "",
 			wantErr: true,
 		},
+		{name: "Should return unchanged string -  no scheme format",
+			args: args{
+				gitRemotes: createTestRemotes(),
+				remoteName: "origin",
+			},
+			want:    "github.com/user-test/test",
+			wantErr: false,
+		},
 		{name: "Should return return error due invalid URL",
 			args: args{
 				gitRemotes: createTestRemotes(),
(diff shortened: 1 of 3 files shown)


pkg/module/gitsource/git_test.go
```diff
@@ -18,12 +18,20 @@ func Test_fetchRepoURLFromRemotes(t *testing.T) {
...
+		},
+		{name: "Should return correct GitHub URL -  https format",
+			args: args{
+				gitRemotes: createTestRemotes(),
+				remoteName: "upstream-https",
 			},
-			want:    "github.com/kyma-test/test",
+			want:    "https://github.com/kyma-test/test.git",
 			wantErr: false,
 		},
```
Tomasz-Smelcerz-SAP, line 36: It would be useful to show what happens if there's a string without scheme prefix at all, like in the linked issue, where they used: `github.com/kyma-project/kyma-module`
There is a test repository in the `git_test.go` that uses such pseudo-URL, called `origin`, but there's no test case for it...
How about adding one? It would look like this:
```
        {name: "Should return unchanged string -  no scheme format",
                args: args{
                        gitRemotes: createTestRemotes(),
                        remoteName: "origin",
                },
                want:    "github.com/user-test/test",
                wantErr: false,
        },
```
What do you think?

jeremyharisch: Will add it

Tomasz-Smelcerz-SAP (approved):
Please take a look at my comment. We nee

PR #1776, opened 2023-09-20
Updating the security scan labels according to the new fields in the security config files

### LINKED ISSUE #1755: Kyma CLI `module create` command has to consider new fields in security configuration file
**Description**

Caused by changing requirements on the security backend side of Kyma, following new fields (which are both not related to one of the scanners) were added to the Kyma security configuration file:

* `/dev-branch` (type: string)
* `/rc-tag` (type: string)
(Should be added as labels in sources)

This field was removed in the meantime:
*  `/whitsource/subprojects` (type: boolean)

Summary/Example of security config file:
```
module-name: btp-operator
rc-tag: 0.5.0.           #<< new field
dev-branch: main.        #<< new field
protecode:
  - europe-docker.pkg.dev/kyma-project/prod/btp-manager:0.5.0
whitesource:
  language: golang-mod
  subprojects: false     #<<< This field is no longer used!
  exclude:
    - "**/*_test.go"
```
Exmaple of rendered moduletemplate:
```
     - access:
          commit: abcd1234
          type: gitHub
        labels:
        - name: git.kyma-project.io/ref
          value: refs/heads/main
          version: v1
        - name: scan.security.kyma-project.io/language
          value: golang-mod
          version: v1
        - name: scan.security.kyma-project.io/dev-branch
          value: main
          version: v1
        - name: scan.security.kyma-project.io/rc-tag
          value: 0.5.0
          version: v1
        - name: scan.security.kyma-project.io/exclude
          value: '**/test/**'
          version: v1
```

The rendering of Kyma module templates should consider these changes and remove the dropped fields but also add the new fields.

AC:
- [x] New added fields (see above) in Kyma security file are added to the generated Kyma module template
- [x] Removed fields (see above) in Kyma security file are not longer added to the generated Kyma module template

nesmabadr:
**Description**

- Reflect the fields changes in the security config files when generating ModuleTemplate using create module command
- Remove `/whitsource/subprojects` field 
- Add `rc-tag` and `dev-branch` fields

**Related issue(s)**
Resolves #1755

### DIFF
--- a/pkg/module/security_scan.go
+++ b/pkg/module/security_scan.go
@@ -16,12 +16,12 @@ import (
 var ErrFailedToParseImageURL = errors.New("error parsing protecode image URL")
 
 const (
-	secScanLabelKey = "scan.security.kyma-project.io"
+	SecScanLabelKey = "scan.security.kyma-project.io"
 	secLabelKey     = "security.kyma-project.io"
 	secScanEnabled  = "enabled"
 )
 
-var labelTemplate = secScanLabelKey + "/%s"
+var labelTemplate = SecScanLabelKey + "/%s"
 var globalLabelTemplate = secLabelKey + "/%s"
 
 func AddSecurityScanningMetadata(descriptor *ocm.ComponentDescriptor, securityConfigPath string) error {
--- a/pkg/module/security_scan.go
+++ b/pkg/module/security_scan.go
@@ -37,17 +37,26 @@ func AddSecurityScanningMetadata(descriptor *ocm.ComponentDescriptor, securityCo
 	if err != nil {
 		return err
 	}
+
 	if len(descriptor.Sources) == 0 {
 		return errors.New("found no sources in component descriptor")
 	}
 	//add whitesource sec scan labels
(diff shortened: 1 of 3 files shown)


pkg/module/security_scan.go
```diff
@@ -117,11 +126,12 @@ type SecurityScanCfg struct {
 	ModuleName  string            `json:"module-name"`
 	Protecode   []string          `json:"protecode"`
 	WhiteSource WhiteSourceSecCfg `json:"whitesource"`
+	DevBranch   string            `json:"dev-branch"`
```
ruanxin, line 129: now we can test the security scan label in [e2e](https://github.com/kyma-project/cli/blob/main/tests/e2e/kyma_create_module_test.go)

This test use template-operator [security scan config file](https://github.com/kyma-project/template-operator/blob/main/sec-scanners-config.yaml), please update it with this new format first.

ruanxin (changes requested):
test is missing

ruanxin (approved):

PR #1797, opened 2023-10-05
fix: Error out if create module is executed with an existing version without the overwrite flag

### LINKED ISSUE #1741: create module to the existing version should not continue and warning user
When the user creates module with existing version without `module-archive-version-overwrite`, although it will not overwrite existing oci image, it still finished successfully and will return back an existing component descriptor in the module template, which is misleading.

AC:
- [x] When the user creates module with an existing version without `module-archive-version-overwrite`, exits processing with error, and displays some warning sentence to indicate user this version already exists, use the `module-archive-version-overwrite` option to overwrite the existing version if needed.

nesmabadr:
**Description**

- Error out if create module is executed with an existing version without the overwrite flag
- Add E2E test

**Related issue(s)**
Fixes #1741


pkg/module/remote.go
```diff
@@ -136,6 +136,16 @@ func (r *Remote) Push(archive *comparch.ComponentArchive, overwrite bool) (ocm.C
 		return nil, err
 	}
 
+	if !overwrite {
+		versionAlreadyExists, _ := repo.ExistsComponentVersion(
```
jeremyharisch, line 140: Any specific reasoning why to ignore the error returned by `ExistsComponentVersion` - If not I would usggest to not ignore it and have a proper error handling

nesmabadr: @jeremyharisch The reason I ignored it is that this function returns an error if the version does not exist.

jeremyharisch (approved):

PR #1799, opened 2023-10-06
fix: Add module version annotation to the ModuleTemplate

nesmabadr:
**Description**

- Add `operator.kyma-project.io/module-version` annotation to the ModuleTemplate
- Validate the annotations in the create module E2E test

**Related issue(s)**
https://github.com/kyma-project/lifecycle-manager/issues/910

### DIFF
--- a/cmd/kyma/alpha/create/module/module.go
+++ b/cmd/kyma/alpha/create/module/module.go
@@ -4,10 +4,12 @@ import (
 	"context"
 	"errors"
 	"fmt"
+	"maps"
 	"os"
 	"path/filepath"
 	"strings"
 
+	"github.com/kyma-project/lifecycle-manager/api/v1beta2"
 	"github.com/mandelsoft/vfs/pkg/memoryfs"
 	"github.com/mandelsoft/vfs/pkg/osfs"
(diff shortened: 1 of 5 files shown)


cmd/kyma/alpha/create/module/module.go
```diff
@@ -407,6 +407,12 @@ func (cmd *command) Run(ctx context.Context) error {
 			annotations["operator.kyma-project.io/is-cluster-scoped"] = "false"
 		}
 
+		version := cmd.opts.Version
+		if modCnf != nil {
+			version = modCnf.Version
+		}
+		annotations["operator.kyma-project.io/module-version"] = version
```
lindnerby, line 414: If you have time, this whole annotations and label part would be a nice candidate to extract and make this crazy long function body shorter. Also possible to make it nicer in terms of using consts for labels, etc. WDYT?

lindnerby (changes requested):

cmd/kyma/alpha/create/module/module.go
nesmabadr: I agree specifically with having the labels and annotations as const :D

lindnerby (approved):

PR #1808, opened 2023-10-18
deps: Upgrade open-component-model/ocm to v0.4.0

lindnerby:
**Description**

- Upgrade OCM archive and descriptor generation:
   - The archive was made inmutable in the latest version after archive creation so the order of descriptor and archive creation needed to change
   - Also many APIs have changed that needed to be adapted, especially repository base types
   - Registration of some repository can be removed, they are already registered as defaults in the library
- Refactoring

**Related issue(s)**
* https://github.com/kyma-project/lifecycle-manager/issues/841


cmd/kyma/alpha/create/module/module.go
```diff
@@ -333,85 +331,83 @@ func (cmd *command) Run(ctx context.Context) error {
...
-	if err := module.AddResources(archive, modDef, l, osFS, cmd.opts.RegistryCredSelector); err != nil {
+	if err = module.AddResources(archive, modDef, l, osFS, cmd.opts.RegistryCredSelector); err != nil {
 		cmd.CurrentStep.Failure()
 		return err
 	}
+	cmd.CurrentStep.Successf("Layers successfully added to archive")
 
-	cmd.CurrentStep.Success()
+	if cmd.opts.RegistryURL == "" {
+		return nil
```
ruanxin, line 350: `return nil` is not expected, right?

lindnerby: It is, the flow just got inverted since this is the last step it can be skipped if there is no reg given

pkg/module/gitsource/git.go
```diff
@@ -28,7 +38,7 @@ func (g GitSource) FetchSource(path, repo, version string) (*ocm.Source, error)
 	access := github.New(repo, "", commit)
 
 	sourceMeta := ocm.SourceMeta{
-		Type: sourceType,
```
Tomasz-Smelcerz-SAP, line 31: I think it is invalid. The ocm library doesn't help here - the field is just a `string` without any docs.
But...
1) The `CONSUMER_TYPE` constant has been moved to the identity package. This package has relation to authn/authz code in the ocm library (e.g: repository credentials). Here we're just defining a type of sources repository, it's something else.
2) The closer look at the ocm library shows this:
```
 ✓ .../github.com/open-component-model/ocm $ ag -Q '"github"'
docs/reference/ocm_add_source-configuration.md
787:$ ocm add source-config sources.yaml --name sources --type filesystem --access '{ "type": "gitHub", "repoUrl": "github.com/open-component-model/ocm", "commit": "xyz" }'

cmds/ocm/commands/ocmcmds/sourceconfig/add/cmd.go
47:$ ocm add source-config sources.yaml --name sources --type filesystem --access '{ "type": "gitHub", "repoUrl": "github.com/open-component-model/ocm", "commit": "xyz" }'

pkg/contexts/ocm/accessmethods/github/method.go
34:     Type   = "gitHub"
39:     LegacyType   = "github"

pkg/contexts/credentials/builtin/github/identity/identity.go
17:const CONSUMER_TYPE = "Github"
``` 

There is a command that `"add a source specification to a source config file"`. And in this command usage doc you can see the `"type": "gitHub"` is used. This constant matches the one defined in the `pkg/contexts/ocm/accessmethods/github/method.go`.

Conclusion: I think we should use one of the constants defined in the `pkg/contexts/ocm/accessmethods/github/method.go` file here, not the one from the `identity` package.
What do you think?

Tomasz-Smelcerz-SAP: OK, I think we should create another issue to verify that - it's also used in Lifecycle-Manager. This PR is not about changing this particular constant - probably it's not important anyway. I still think this particular value is incorrect, but it was like that for a long time.
Neither OCM code nor the spec is clear on that topic, so I am not going spend more time on it.

cmd/kyma/alpha/create/module/module.go
ruanxin: but if no regstry given, this create module command is meaningless, right? It can't push image to remote.

pkg/module/resources.go
```diff
@@ -75,24 +55,23 @@ func AddResources(
 			log.Debugf("Added input blob from %q", resource.Input.Path)
 		}
 	}
-	compdesc.DefaultResources(descriptor)
```
Tomasz-Smelcerz-SAP, line 78: Why are these removed?

pkg/module/resources.go
```diff
@@ -111,10 +90,10 @@ func generateResources(log *zap.SugaredLogger, version string, credLabel []byte,
 			r.Input.Type = "file"
 		}
 
-		if len(credLabel) != 0 {
+		if len(credMatchLabels) != 0 {
```
Tomasz-Smelcerz-SAP, line 93: I don't know why, but it looks strange to me, I'd prefer the following:
```suggestion
		if len(credMatchLabels) > 0 {
```

pkg/module/remote.go
```diff
@@ -100,9 +100,7 @@ func (r *Remote) getCredentials(ctx cpi.Context) credentials.Credentials {
 	return creds
 }
 
-// See: github.com/open-component-model/ocm/pkg/contexts/credentials/repositories/dockerconfig/repository.go#IsEmptyAuthConfig()
```
Tomasz-Smelcerz-SAP, line 103: Maybe leave this  comment? Tt indicates that the function is written as the one linked.

pkg/module/build.go
```diff
@@ -39,58 +31,45 @@ func CreateArchive(fs vfs.FileSystem, path, gitRemote string, def *Definition, i
...
+		ctx,
+		accessobj.ACC_CREATE, archiveFs, nil,
+		nil,
+		vfs.ModePerm,
+	)
 }
 
-// addSources adds the sources to the component descriptor. If the def.Source is a git repository
-func addSources(cd *ocm.ComponentDescriptor, def *Definition, gitRemote string) error {
-	if strings.HasSuffix(def.Source, ".git") {
```
Tomasz-Smelcerz-SAP, line 81: This condition was added as part of #1751. Is it still working?

pkg/module/build.go
```diff
@@ -39,58 +31,45 @@ func CreateArchive(fs vfs.FileSystem, path, gitRemote string, def *Definition, i
...
 	ctx := cpi.DefaultContext()
 
-	archive, err := comparch.New(
-		ctx,
-		accessobj.ACC_CREATE, archiveFs,
-		nil,
-		nil,
-		vfs.ModePerm,
-	)
 	if err != nil {
```
Tomasz-Smelcerz-SAP, line 34: This check has no effect

pkg/module/build.go
```diff
@@ -39,58 +31,45 @@ func CreateArchive(fs vfs.FileSystem, path, gitRemote string, def *Definition, i
...
 		return nil, err
 	}
-
-	cd.Provider = v1.Provider{Name: "kyma-project.io", Labels: v1.Labels{*builtByCLI}}
-
-	if isTargetDirAGitRepo {
-		if err := addSources(cd, def, gitRemote); err != nil {
-			return nil, err
-		}
+	file, err := archiveFs.Create("component-descriptor.yaml")
```
Tomasz-Smelcerz-SAP, line 45: Maybe I am wrong, but it looks like it's bypassing the ocm functionality - in the previous version this file was created by the library, isn't it?

pkg/module/build.go
```diff
@@ -39,58 +31,45 @@ func CreateArchive(fs vfs.FileSystem, path, gitRemote string, def *Definition, i
...
-
-	if err := ocm.Validate(cd); err != nil {
-		return nil, fmt.Errorf("unable to validate component descriptor: %w", err)
+	_, err = file.Write(data)
+	if err != nil {
+		return nil, err
 	}
-
-	return archive, nil
+	defer file.Close()
```
Tomasz-Smelcerz-SAP, line 53: I would move this even before `file.Write` in line 49.

pkg/module/resources.go
```diff
@@ -45,23 +40,8 @@ func AddResources(
...
-	if registryCredSelector != "" {
-		selector, err := metav1.ParseToLabelSelector(registryCredSelector)
-		if err != nil {
-			return err
-		}
-		matchLabels, err = json.Marshal(selector.MatchLabels)
-		if err != nil {
-			return err
-		}
-		descriptor.SetLabels([]ocmv1.Label{{
```
Tomasz-Smelcerz-SAP, line 59: Why is this removed?

pkg/module/remote.go
lindnerby: The link does not work, that's why I removed it.

pkg/module/resources.go
lindnerby: Labels are added here: https://github.com/kyma-project/cli/blob/4bbba5934c0e77cbb38b6a7162eb4c182bf40a98/pkg/module/resources.go#L69

cmd/kyma/alpha/create/module/module.go
lindnerby: i guess it's like this because the command could also be used for just persisting the archive locally to inspect layers? i don't know, the registry flag is also not mandatory. i don't know the intentions/requirements when this command was implemented, but agree that this should have a second look at some point.
and it's not only the registry flow, also the generation of the template could be interesting to be used without registry, but at the moment the flow ends here.

Tomasz-Smelcerz-SAP (approved):

PR #1864, opened 2023-11-24
fix: Fix Zip Slip Vulnerability

jeremyharisch:
**Description**

- Fixes https://github.com/kyma-project/cli/security/code-scanning/1
- Goland auto aligned the code with the Code-Style preferences we discussed in the team

**Related issue(s)**

### DIFF
--- a/internal/deploy/istioctl/istio.go
+++ b/internal/deploy/istioctl/istio.go
@@ -36,7 +36,7 @@ const (
 )
 
 var (
-	ErrIstioSourcepath = errors.New("istio source path contains `..`")
+	errIstioSourcepath = errors.New("istio source path contains `..`")
 )
 
 type operatingSystem struct {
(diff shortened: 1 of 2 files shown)


internal/deploy/istioctl/istio.go
```diff
@@ -334,6 +342,9 @@ func unGzip(source, target string, deleteSource bool) error {
 }
 
 func unTar(source, target string, deleteSource bool) error {
+	if strings.Contains(source, "..") {
+		return ErrIstioSourcepath
```
lindnerby, line 346:
```suggestion
		return errIstioSourcepath
```

lindnerby (commented):
Small suggestion first, but can you explain how this actually fixes the vulnerability in istioctl? Basically the details of the added check `strings.Contains(source, "..")` ?

lindnerby (approved):

PR #1865, opened 2023-11-24
fix: Allow creating module for existing versions

### LINKED ISSUE #1862: Allow create module generate moduletemplate if version already exists.
**Description**

We have the use case in channel switch, different channels use same version due to the version upgrade strategy. However at the moment, without using `module-archive-version-overwrite` flag, CLI will fail if the version already exists in remote OCI repo, that's why currently in the module submission pipeline, regardless of the version, it will always use this overwrite flag.

It's not a problem to overwrite version if the content is same, because the layer sha value will be consistent, however, if the content is different, then the layer of old moduletemplate will be replaced, which leads to the old moduletemplate becomes invalid.

**AC**
Given a module config with version which exists in remote OCI registry
And the content of `manifest` is same between remote and local
Then execute `alpha create module`
Then CLI fetch component descriptor from remote, instead of push local content
And generate module template

Given a module config with version which exists in remote OCI registry
And the content of `manifest` is different between remote and local
Then execute `alpha create module`
Then CLI print out the reason to indicate there is diff between the content, can't push to remote.
And stop the process with error

- [x] Include test cases for the CLI E2E tests
- [x] Dedicated unit/integration test to validate there is no push if the repo content is the same

nesmabadr:
**Description**

- If no overwrite flag is provided and version already exists:
   - if the local and remote contents are the same: allow ModuleTemplate creation from the remote image without pushing the local one
   - if the contents are different: display an error
- Adjust E2E test for the module creation
- Add relevant tests 

**Related issue(s)**
Resolves #1862


pkg/module/remote.go
```diff
@@ -126,45 +125,47 @@ func NoSchemeURL(url string) string {
 	return regex.ReplaceAllString(url, "")
 }
 
-// Push picks up the archive described in the config and pushes it to the provided registry.
+// Push picks up the archive described in the config and pushes it to the provided registry if not existing.
```
Tomasz-Smelcerz-SAP, line 128: I think this entire godoc deserves some improvement.
Maybe we should also change the method name to reflect it's current logic? Something like `PushIfNotExists` ?
Another idea would be to keep the `Push` method for pushing - without any conditions, and move all the resources comparison to another method.
To summarize: I think this method now does too much, it's hard to even write a good docs for it.

Tomasz-Smelcerz-SAP (approved):

Issue #1742, opened 2023-08-23
Command to create a "scaffolding" for module creation


janmedrek:
**Description**

Some time ago we introduced [a new way of creating modules](https://github.com/kyma-project/cli/issues/1684) - by providing a specific metadata file and basic information on the module.

It requires quite a vast input in the form of different files, it would be great to have a command for creating empty files that can be populated with data by the module creator.

**Reasons**

We want to have a smooth experience for the module creation.

**Acceptance Criteria**
- [x] Command to generate input scaffolding available
- [x] Metadata YAML is always created
- [x] For the required fields add an option to pass the values inline and have them already filled in the generated files
- [x] Parameters to control which additional files are generated:
  - defaultCR YAML
  - manifest YAML
      - Empty yaml file named `manifest.yaml`
  - security scanners config YAML
  - [x] File references in metadata YAML are provided upfront and point to generated files (depending on the listed above)
  - [x] Generated files have a generated structure/description of what should be put into them
    - think about creating a file structure and adding field explanations
- [x] Fields and files have their explanations and placeholders in them, for example (see examples in the Attachments section)
- [x] File descriptions and explanations aligned with the Technical Writers
- [x] Find a way to introduce proper testing related to the create command chaining with scaffolding of modul-config (write a test or create a ticket with a proposal)
  
**Attachments**

Example for generated metadata YAML (**make sure it is up-to-date with the features create module command**):
```
name:         # required (string), the name of the Module
channel:           # required (string), channel that should be used in the ModuleTemplate
version:           # required (string), the version of the Module
# resourceName:      # optional (string), default={NAME}-{CHANNEL}, the name for the ModuleTemplate that will be created

manifest:         # required (string), reference to the manifests, can be a name of a file or whole directory
# security:          # optional (string), name of the security scanners config file
# defaultCR:       #optional (string), reference to a YAML file containing the default CR for the module, must be a relative file name

#internal:            # optional (bool), default=false, determines whether the ModuleTemplate should have the internal flag or not
#beta:                # optional (bool), default=false, determines whether the ModuleTemplate should have the beta flag or not

#labels:                   # optional, additional labels for the ModuleTemplate
#  - operator.kyma-project.io/label={LABEL}
#annotations:              # optional, additional annotations for the ModuleTemplate
#  - operator.kyma-project.io/doc-url={DOCUMENTATION_URL}
```

Example for generated metadata defaultCR file (**make sure it is up-to-date with the features create module command**):
```
[...]

PR #1866, opened 2023-11-24
feat: Introduce Mandatory Spec field for ModuleTemplate

jeremyharisch:
**Description**

- Implements https://github.com/kyma-project/lifecycle-manager/issues/1051
- Adjusting the scaffolding is not included, since this task is not implemented until not, I will create a follow-up for it (https://github.com/kyma-project/cli/issues/1742)
- Since dependecy to Lifecycle-manger was upgrades, following changes have been introduced:
    - Use `shared` pkg for States
    - refactored Interactor interface for better testability
    - Rewrote unit test to tabledriven, use new mock, cover one more test-case, decouple from Lifecycle-Manager testutils pkg
- Cover new feature in e2e test

**Related issue(s)**


cmd/kyma/alpha/create/module/moduleconfig.go
```diff
@@ -17,6 +17,7 @@ type Config struct {
 	Version           string                     `yaml:"version"`          //required, the version of the Module
 	Channel           string                     `yaml:"channel"`          //required, channel that should be used in the ModuleTemplate
 	ManifestPath      string                     `yaml:"manifest"`         //required, reference to the manifests, must be a relative file name.
+	Mandatory         bool                       `yaml:"mandatory"`        //required, indicates whether the module is mandatory to be installed on all clusters.
```
nesmabadr, line 20: I think we need to indicate that this is by default set to false. Also, in this case, it should be optional not required.

nesmabadr: Yes I mean, we need to indicate this (that by default it is set to false) in the documentation

jeremyharisch: And no, Mandatory should be required here. because if it is not given than it will default to false

nesmabadr: The thing is if mandatory is not provided in the moduleconfig, the command will not fail, it will just set it to false. That's why I believe we should add that the default value is false here.

pkg/module/template.go
```diff
@@ -48,6 +49,7 @@ spec:
 {{- end}}
   descriptor:
 {{yaml .Descriptor | printf "%s" | indent 4}}
+  mandatory: {{.Mandatory}}
```
nesmabadr, line 52: I suggest moving the Mandatory field right after Channel (before data and descriptor) so that it can be easily found in the ModuleTemplate

nesmabadr (approved):

PR #1909, opened 2024-01-12
chore: Remove Module Validation from create module command

### LINKED ISSUE #1846: [Discussion] If we remove validation Module CR in CLI
At the moment, we introduce envTest into CLI to do module CR validation.

Cons:
- It introduces a test framework into production code, which is wired, also takes time to provision a in-memory cluster to do validation.
- To make the log work as expected, it must initialize a different logger, can't use log from CLI. https://github.com/kyma-project/cli/issues/1763

Does this validation bring value? @janmedrek 
Basically, this validation can verified in Module Teams' testing pipeline or module submission pipeline, is it really worth this effort to do validation during module template creation?

If have benefit from user side, we can think some static validation implementation:
https://github.com/kubernetes/apiextensions-apiserver/blob/v0.28.3/pkg/apiserver/validation/validation.go

nesmabadr:
**Description**

- Remove module validation
- Remove envtest

**Related issue(s)**
Resolves #1846


lindnerby:
@nesmabadr Can you check if you can also remove `sigs.k8s.io/controller-runtime/pkg/envtest` from deps now please?

lindnerby (approved):

Issue #1907, opened 2024-01-11
Upgrade OCM library to same version as used in KLM (0.6.0)


jeremyharisch:
**Description**
To mitigate the identified security risks, we propose upgrading the OCM package to version 0.6.0 (same as used in KLM), which includes patches and updates addressing the reported vulnerabilities.

**Acceptance Criterias**
- [ ] Update the OCM package in the project's dependency file.
- [ ] Fix breaking changes.
- [ ] Test the application thoroughly to ensure compatibility with the new version.
- [ ] Update documentation to reflect the new package version and any breaking changes.
- [ ]  Resolve this [dependabot PR](https://github.com/kyma-project/cli/pull/1900)
- [ ] Prepare knowledge sharing about OCM pkg findings.

janmedrek:
Stopping the development according to: https://github.com/kyma-project/community/issues/872

PR #1934, opened 2024-01-29
chore: Bump ocm library to v0.6.0

Tomasz-Smelcerz-SAP:
**Description**

Update the [ocm](https://github.com/open-component-model/ocm) library dependency from `v0.4.0` to `v0.6.0`

- Upgrade OCM dependency
- Necessary code fixes

**Related issue(s)**
https://github.com/kyma-project/cli/issues/1907

**See also:**
Related OCM issue created as a result of this PR: https://github.com/open-component-model/ocm/issues/643

### DIFF
--- a/pkg/module/resources.go
+++ b/pkg/module/resources.go
@@ -14,6 +14,7 @@ import (
 	"github.com/open-component-model/ocm/pkg/common/accessobj"
 	"github.com/open-component-model/ocm/pkg/contexts/ocm/compdesc"
 	ocmv1 "github.com/open-component-model/ocm/pkg/contexts/ocm/compdesc/meta/v1"
+	"github.com/open-component-model/ocm/pkg/contexts/ocm/cpi"
 	"github.com/open-component-model/ocm/pkg/contexts/ocm/repositories/comparch"
 	"github.com/pkg/errors"
 	"go.uber.org/zap"
@@ -117,7 +118,7 @@ func addBlob(fs vfs.FileSystem, archive *comparch.ComponentArchive, resource *Re
 		return err
 	}
 
-	return archive.SetResource(&resource.ResourceMeta, blobAccess)
+	return archive.SetResource(&resource.ResourceMeta, blobAccess, cpi.ModifyResource(true))
 }
 
 func (rd ResourceDescriptor) String() string {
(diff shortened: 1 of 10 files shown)


pkg/module/security_scan.go
```diff
@@ -97,7 +97,7 @@ func appendProtecodeImagesLayers(descriptor *ocm.ComponentDescriptor, config *Se
 					Labels:  []ocmv1.Label{*imageTypeLabel},
 					Version: imageTag,
 				},
-				Type:     ociimage.TYPE,
+				Type:     itociartifact.LEGACY_TYPE, //#561
```
ruanxin, line 100: what's the meaning of this `#561`? a PR number? do we really need this in the comment?

pkg/module/sign.go
```diff
@@ -112,7 +113,7 @@ func privateKey(pathToPrivateKey string) (interface{}, error) {
 		return nil, fmt.Errorf("unable to open key file: %w", err)
 	}
 
-	key, err := signing.ParsePrivateKey(privateKeyFile)
+	key, err := signutils.ParsePublicKey(privateKeyFile)
```
ruanxin, line 116:
```suggestion
	key, err := signutils.ParsePrivateKey(privateKeyFile)
```

pkg/module/security_scan.go
Tomasz-Smelcerz-SAP: Yes, PR number from OCM library that introduced that change. Leftover from investigating, removed.

pkg/module/resources.go
```diff
@@ -117,7 +118,7 @@ func addBlob(fs vfs.FileSystem, archive *comparch.ComponentArchive, resource *Re
 		return err
 	}
 
-	return archive.SetResource(&resource.ResourceMeta, blobAccess)
+	return archive.SetResource(&resource.ResourceMeta, blobAccess, cpi.ModifyResource(true))
```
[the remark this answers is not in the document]
Tomasz-Smelcerz-SAP: The answer is: I don't know (yet). Without it, the current code fails with a message (I am recalling from my memory, these aren't exact words): "changes to archive would invalidate the digest signature"
It may be that the way we're constructing the component archive is not "optimal" - it is different from what OCM library authors assume as a "right path". And once we change our code, this option is no longer necessary.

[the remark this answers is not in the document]
ruanxin: I reproduced it with same error

ruanxin (approved):

PR #2006, opened 2024-03-19
chore: Make Nodejs20 default runtime

halamix2:
**Description**

- Make Nodejs20 default runtime

**Related issue(s)**

- https://github.com/kyma-project/serverless/issues/748


cmd/kyma/init/function/function_test.go
```diff
@@ -52,13 +52,13 @@ func TestFunctionFlags(t *testing.T) {
 
 	err = c.ParseFlags([]string{
 		"-d", "/tmpfile",
-		"-r", "nodejs18",
+		"-r", "nodejs20",
```
anoipm, line 55: IMO in this place better use non default runtime. if we have bug and ignore `-r` flag, result has default runtime and test is useless.

anoipm (approved):

grego952 (approved):

Issue #2001, opened 2024-03-18
Implement tests for importing docker images into kyma's docker registry


kwiatekus:
**Description**

Add unit tests to #1979

PR #2008, opened 2024-03-20
Test the `transport.go` file

pPrecel:
**Description**

- add tests to cover all the most important cases in the `transport.go` file

**Related issue(s)**

- https://github.com/kyma-project/cli/issues/2001


internal/registry/portforward/transport_test.go
```diff
@@ -0,0 +1,203 @@
...
+			remotePort: "8080",
+			remoteConn: fixConnectionMock(t,
+				fixBrokenErrStreamMock(t, "test error message", io.EOF),
+				fixDataStreamMock(t, testRequest, &testResponse),
+			),
+		}
+
+		got, err := pft.dialRemote(testRequest)
+
+		require.Error(t, err)
```
anoipm, line 85: Do we expect error message and check it? The same question for other tests.

anoipm (approved):

PR #2009, opened 2024-03-20
Test the `image.go` file

pPrecel:
**Description**

- add tests to cover all the most important cases in the `image.go` file

**Related issue(s)**

- https://github.com/kyma-project/cli/issues/2001

### DIFF
--- a/internal/registry/image.go
+++ b/internal/registry/image.go
@@ -11,6 +11,7 @@ import (
 	"github.com/google/go-containerregistry/pkg/v1/daemon"
 	"github.com/google/go-containerregistry/pkg/v1/remote"
 	"github.com/kyma-project/cli.v3/internal/registry/portforward"
+	"k8s.io/apimachinery/pkg/util/httpstream"
 	"k8s.io/client-go/rest"
 )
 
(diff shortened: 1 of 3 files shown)


internal/registry/image.go
```diff
@@ -23,13 +24,28 @@ type ImportOptions struct {
 	RegistryPodPort      string
 }
 
+// for testing
+type utils struct {
+	daemonImage        func(name.Reference, ...daemon.Option) (v1.Image, error)
+	portforwardNewDial func(config *rest.Config, podName, podNamespace string) (httpstream.Connection, error)
+	remoteWrite        func(ref name.Reference, img v1.Image, options ...remote.Option) (rerr error)
```
anoipm, line 31: Do we need name of returned value (`rerr`) in this place?

internal/registry/image_test.go
```diff
@@ -0,0 +1,181 @@
...
+						mock := automock.NewConnection(t)
+						mock.On("Close").Return(nil).Once()
+						return mock, nil
+					},
+				},
+			},
+			wantErr: errors.New("failed to push image to the in-cluster registry: registries must be valid RFC 3986 URI authorities: <    >"),
+		},
+		{
+			name: "import image",
```
anoipm, line 141: Does this test import image?

internal/registry/image.go
pPrecel: nope, it's autogenerated. I will remove it

internal/registry/image_test.go
pPrecel: at least tries but then fails with an error :P I will rename it

anoipm (approved):

Issue #1997, opened 2024-03-14
Allow referencing shared service instances via kyma CLI


kwiatekus:
**Description**
Kyma cli should make it easy to create 'reference-instance' that references already existing service instances.
This way users get an open path to binding to existing instances and also deleting kyma environment w/o the need to remove all phisically provisioned service instances

**Execution**
 - [x] try it out manually in btp (creating shared service instance and reference it ( and bind ) in kyma runtime) (1)
 - [x] implement a command that would take service offering name and shared instance name to create a reference service instance

**Reason**
Kyma CLI should enable users an effortless binding of service instances that were already created and marked as "shared"
Cases for instance sharing:
 - An application may consist of multiple microservices running on different platforms. Sharing the same RabbitMQ service instance would allow pubsub communication between those microservices. They can also share the same Postgresql service instance if needed.
 - The application can be moved to a different platform and still use the same service instances that it previously used.

**Attachment**
https://www.youtube.com/watch?v=AYULBewzEwI
https://help.sap.com/docs/btp/btp-cli-command-reference/btp-share-services-instance
https://wiki.one.int.sap/wiki/display/CPC15N/%5BCookbook%5D+Service+Instance+Sharing

kwiatekus:
A target service instance must be "shared" <-- for now we treat this as a prerequisite, i.e user must share it via btp cockpit or [btp cli](https://help.sap.com/docs/btp/btp-cli-command-reference/btp-share-services-instance)

![Image](https://github.com/orgs/kyma-project/projects/26/assets/6783567/3d663475-358b-4740-80f0-8d0a4e1431a4)

A kyma cli command to reference a shared instance in  kyma runtime should create a **service instance** k8s resource  (taking service offering name, and name of the shared instance from user input). It should use fixed ` reference-instance` plan name 

```yaml
spec:
  externalName: object-store-ref
  parameters:
    referenced_instance_id: ''
    selectors:
      instance_label_selector: []
      instance_name_selector: object-store-cf
      plan_name_selector: ''
  serviceOfferingName: objectstore
  servicePlanName: reference-instance
```

and **service binding**

kwiatekus:
```bash
 ./kyma-cli reference-instance --name-selector=kk-object-store --offering-name=objectstore --reference-name=kk-object-store-ref --kubeconfig=/Users/krzysztofkwiatosz/Downloads/cli.yaml
```

I would like to propose available options renaming:
 - ~`name-selector`~ -> `target-name-selector`
[...]

PR #2022, opened 2024-03-27
Add reference-instance command

halamix2:
**Description**

- Add reference-instance command

**Related issue(s)**

- #1997

### DIFF
--- a/internal/cmd/referenceinstance/referenceinstance.go
+++ b/internal/cmd/referenceinstance/referenceinstance.go
@@ -72,6 +72,32 @@ func (pc *referenceInstanceConfig) complete() error {
 }
 
 func runReferenceInstance(config referenceInstanceConfig) error {
+	requestData := fillRequestData(config)
+	u, err := runtime.DefaultUnstructuredConverter.ToUnstructured(&requestData)
+	if err != nil {
+		return err
+	}
+
+	unstructuredObj := &unstructured.Unstructured{
+		Object: u,
+	}
+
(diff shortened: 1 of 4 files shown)


internal/cmd/referenceinstance/referenceinstance.go
```diff
@@ -0,0 +1,127 @@
...
+	if err != nil {
+		return err
+	}
+
+	unstructuredObj := &unstructured.Unstructured{
+		Object: u,
+	}
+
+	unstructuredObj.SetGroupVersionKind(schema.GroupVersionKind{
+		Group:   "services.cloud.sap.com",
```
anoipm, line 86: It's already defined in `internal/btp/operator/gvr.go`

internal/cmd/referenceinstance/referenceinstance.go
```diff
@@ -0,0 +1,127 @@
...
+
+	var err error
+	pc.kubeClient, err = kube.NewClient(pc.kubeconfig)
+
+	return err
+}
+
+func runReferenceInstance(config referenceInstanceConfig) error {
+	requestData := fillRequestData(config)
+	u, err := runtime.DefaultUnstructuredConverter.ToUnstructured(&requestData)
```
anoipm, line 76: Could we extract preparation of `unstructuredObj` to separated method? Now I see two parts of code: data preparation, creating resource on kubernetes. IMO now second operation is hard to find.

anoipm (approved):

Issue #2045, opened 2024-04-15
Resolve hana provisioning state before mapping


kwiatekus:
**Description**
As part od hana map command implement waiting for hana probisioning. Identify Error/Creation Failed conditions to break the waiting loop. 

**Reason**
Before hana mapping can be done, hana provisioning must be finished.

PR #2046, opened 2024-04-15
Better hana map hints

halamix2:
**Description**

- add hints about missing Hana API entitlement
- wait untim HANA API instance and mapping are ready
- do not fail if Hana API or the binding already exists

**Related issue(s)**

- #2045

### DIFF
--- a/internal/cmd/hana/check.go
+++ b/internal/cmd/hana/check.go
@@ -53,10 +53,10 @@ func NewHanaCheckCMD(kymaConfig *cmdcommon.KymaConfig) *cobra.Command {
 }
 
 type somethingWithStatus struct {
-	Status status
+	Status Status
 }
 
-type status struct {
+type Status struct {
 	Conditions []metav1.Condition
 	Ready      string
 	InstanceID string
@@ -85,7 +85,7 @@ func runCheck(config *hanaCheckConfig) error {
 }
 
 func checkHanaInstance(config *hanaCheckConfig) error {
-	u, err := getServiceInstance(config)
+	u, err := getServiceInstance(config, config.name)
 	return handleCheckResponse(u, err, "Hana instance", config.namespace, config.name)
 }
 
@@ -111,7 +111,7 @@ func handleCheckResponse(u *unstructured.Unstructured, err error, printedName, n
 		}
 	}
 
-	ready, error := isReady(u)
+	ready, _, error := isReady(u)
 	if error != nil {
 		return &clierror.Error{
 			Message: "failed to check readiness of Hana resources",
@@ -133,7 +133,7 @@ func handleCheckResponse(u *unstructured.Unstructured, err error, printedName, n
 	return nil
 }
 
-func getServiceInstance(config *hanaCheckConfig) (*unstructured.Unstructured, error) {
+func getServiceInstance(config *hanaCheckConfig, name string) (*unstructured.Unstructured, error) {
 	return config.KubeClient.Dynamic().Resource(operator.GVRServiceInstance).
 		Namespace(config.namespace).
 		Get(config.Ctx, config.name, metav1.GetOptions{})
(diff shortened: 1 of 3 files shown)


internal/cmd/hana/check.go
```diff
@@ -143,10 +143,10 @@ func handleCheckResponse(u *unstructured.Unstructured, err error, printedName, n
 	return nil
 }
 
-func getServiceInstance(config *hanaCheckConfig) (*unstructured.Unstructured, error) {
+func getServiceInstance(config *hanaCheckConfig, name string) (*unstructured.Unstructured, error) {
```
anoipm, line 146: I see incostistency with params. IMO we should pass `(config)` or `(ctx, name)`. Now we have struct and next argument from this struct.

internal/cmd/hana/map.go
```diff
@@ -66,6 +66,13 @@ func runMap(config *hanaCheckConfig) error {
 }
 
 func createHanaAPIInstance(config *hanaCheckConfig) error {
+	//check if instance exists, skip API creation if it does
+	instance, err := getServiceInstance(config, hanaBindingAPIName(config.name))
```
anoipm, line 70: Could we extract this code to another function? This check is not main responsibility of `createHanaAPIInstance`.

internal/cmd/hana/map.go
```diff
@@ -80,6 +87,13 @@ func createHanaAPIInstance(config *hanaCheckConfig) error {
 }
 
 func createHanaAPIBinding(config *hanaCheckConfig) error {
+	//check if instance exists, skip API creation if it does
```
anoipm, line 90: Could we extract this code to another function? This check is not main responsibility of `createHanaAPIBinding`.

internal/cmd/hana/check.go
```diff
@@ -155,25 +155,46 @@ func getServiceBinding(config *hanaCheckConfig, name string) (*unstructured.Unst
...
 			Details: err.Error(),
 		}
 	}
-	status := instance.Status
+
+	return instance.Status, nil
+}
+
+// isReady returns readiness status, and failed status if at least one of the contitions has failed, or an error was returned
+func isReady(u *unstructured.Unstructured) (bool, bool, error) {
```
anoipm, line 171: Why this function return 3 values? IMO we have mixed two different functions `isReady` and `isFailed` and I don't see pros of mixing them together.

internal/cmd/hana/map.go
```diff
@@ -51,8 +49,8 @@ func NewMapHanaCMD(kymaConfig *cmdcommon.KymaConfig) *cobra.Command {
 
 var (
 	mapCommands = []func(config *hanaCheckConfig) error{
-		createHanaAPIInstance,
-		createHanaAPIBinding,
+		checkAndCreateHanaAPIInstance,
```
anoipm, line 52: IMO this prefix of this name say almost nothing. What we check? Maybe we `createHanaAPIInstanceIfNeeded`, `createHanaAPIInstanceIfNotExists` or something like this?

internal/cmd/hana/map.go
```diff
@@ -183,28 +204,15 @@ func getClusterID(config *hanaCheckConfig) (string, error) {
...
-	}
-}
-
 func getHanaID(config *hanaCheckConfig) (string, error) {
-	instance := somethingWithStatus{}
-
 	// wait for until Hana instance is ready, for default setting it should take 5 minutes
 	fmt.Print("waiting for Hana instance to be ready... ")
-	err := wait.PollUntilContextTimeout(config.Ctx, 10*time.Second, config.timeout, true, isHanaReady(config))
+	err := wait.PollUntilContextTimeout(config.Ctx, 10*time.Second, config.timeout, true, kube.IsInstanceReady(config.kubeClient, config.Ctx, config.namespace, config.name))
```
anoipm, line 210: IMO this row is hard to read because of length. Could we calculate `kube.IsInstanceReady(...` befor this line?

anoipm (approved):

Issue #2036, opened 2024-04-09
Produce token based kubeconfig originating from ServiceAccount via kyma cli


kwiatekus:
**Description**

Provide a command that produces a kubeconfig file that includes token for a specially created service account that is bound to a cluster-admin role.

Command should (using `--kubeconfig`):
 - create a new service account
 - bind sevice account to referred cluster role
 - issue a [token request](https://kubernetes.io/docs/reference/kubernetes-api/authentication-resources/token-request-v1/) or create secret of type `kubernetes.io/service-account-token` ([tutorial](https://developers.sap.com/tutorials/kyma-create-service-account.html))
 - return a kubeconfig file where the token is used as user context  

Sth like
`kyma alpha access --clusterrole={clusterRoleName} --kubeconfig={inputKubeconfig} --output=kubeconfig-sa.yaml --name={my-service-account}`, where:

 - `--name` - name of the new SA to be created by kyma cli (mandatory)
 - `--clusterrole` - mandatory - decides which role to bind to the new SA
 - `--kubeconfig` - mandatory (can be consumed from `KUBECONFIG` env)
 - `--output` - path where to save the output kubeconfig file (optional. if skipped print the kubeconfig (?))
 - `--namespace` - optional (by default create service account in default namespace)
 - `--time` - optional (by deafult grant access for 1h); accepts integer in hours or days. For example `--time 3h` or `--time 2d`
 - `--permanent`- optional; Creates an access token based on secret without any time limit
 
 Similar future exists in kyma dashboard on service account detail view

**Reasons**
It doesn't yet solve getting a kubeconfig for a freshly created cluster in automated way, but it reduces effort of manual configuration for the time until we finally solve the problem.  Kyma users are forced to do it (see [example](https://github.com/kyma-project/kyma-environment-broker/issues/284#issuecomment-1905997723)) until there is a real solution

**Attachments**

PR #2067, opened 2024-05-08
Add Kyma Access command to CLI

Cortey:
**Description**
Adds a command that creates a kubeconfig that includes token and certificate for specially crafted service account with admin access rights

**Related issue(s)**

- #2036


internal/cmd/access/access.go
```diff
@@ -0,0 +1,226 @@
...
+			return runAccess(&cfg)
+		},
+	}
+
+	cfg.KubeClientConfig.AddFlag(cmd)
+
+	cmd.Flags().StringVar(&cfg.name, "name", "", "Name of the Service Account to be created")
+	cmd.Flags().StringVar(&cfg.clusterrole, "clusterrole", "", "Name of the cluster role to bind the Service Account")
+	cmd.Flags().StringVar(&cfg.output, "output", "", "Path to the output kubeconfig file")
+	cmd.Flags().StringVar(&cfg.namespace, "namespace", "default", "Namespace ")
```
pPrecel, line 49: I would suggest to improve the `--namespace` flag :)

internal/cmd/access/access.go
```diff
@@ -0,0 +1,226 @@
...
+		Rules: []rbacv1.PolicyRule{
+			{
+				APIGroups: []string{"", "extensions", "batch", "apps", "gateway.kyma-project.io", "servicecatalog.k8s.io"},
+				Resources: []string{"deployments", "replicasets", "pods", "jobs", "configmaps", "apirules", "serviceinstances", "servicebindings", "services", "secrets"},
+				Verbs:     []string{"create", "update", "patch", "delete", "get", "list"},
+			},
+		},
+	}
+	_, err := cfg.KubeClient.Static().RbacV1().ClusterRoles().Create(cfg.Ctx, &cRole, metav1.CreateOptions{})
+	if client.IgnoreNotFound(err) == nil {
```
pPrecel, line 197: I would suggest to replace `client.IngoreNotFound` with `errors.IsAlreadyExists` from the `k8s.io/apimachinery/pkg/api/errors`. I think the `Create` func will never return a `not found` error because in this case this func will just create it

pPrecel (commented):
In general, looks really ok. The only doubt I have is if it's not better to extract most functions that are creating Kubeconfig to the separate cluster and test it. But I don't want to force it in this PR. Maybe it would be better to create a separate PR.

pPrecel (approved):

Issue #2070, opened 2024-05-10
Adjust secret lookup logic for `image-import` command 


kwiatekus:
**Description**
`image-import` command should be able to lookup docker registry secret from [docker-registry](https://github.com/kyma-project/docker-registry/) module.

Currently it looks up secret by name `serverless-registry-config-default` and in the scenario when docker registry is installed it returns error
```
./kyma-cli image-import  kwiatekus/test:latest
failed to load in-cluster registry configuration: secrets "serverless-registry-config-default" not found
```

Adjust the secret lookup logic so that:
 - [x] secret announced in docker registry status  is selected as the first choice

Additionally please improve the success message after importing.. Looks to me that the string templating is mixes up strings
```
Importing kwiatekus/redispop:latest

Successfully imported image
Use it as 'dockerregistry-config' and use the localhost:32137/kwiatekus/redispop:latest secret.
```
Should be:
```
Use it as localhost:32137/kwiatekus/redispop:latest and use the 'dockerregistry-config' as image pull secret.
```

PR #2072, opened 2024-05-13
Support DockerRegistry CR in the `image-import` command

pPrecel:
**Description**

- use the DockerRegistry CR to import image
- if DR does not exist, try to import image to the old (serverless based) registry
- improve the success message after importing

**Related issue(s)**

- https://github.com/kyma-project/cli/issues/2070#issue-2289027892

### DIFF
--- a/internal/registry/config.go
+++ b/internal/registry/config.go
@@ -4,47 +4,87 @@ import (
 	"context"
 	"errors"
 	"fmt"
-	"strconv"
 	"strings"
 
+	"github.com/kyma-project/cli.v3/internal/kube"
 	corev1 "k8s.io/api/core/v1"
 	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
+	"k8s.io/client-go/dynamic"
 	"k8s.io/client-go/kubernetes"
 )
 
 const (
-	RegistrySecretName  = "serverless-registry-config-default"
-	serverlessNamespace = "kyma-system"
+	LegacyRegistrySecretName  = "serverless-registry-config-default"
+	LegacyServerlessNamespace = "kyma-system"
 )
 
 type RegistryConfig struct {
-	DockerConfigJson string
-	Username         string
-	Password         string
-	PullRegAddr      string
-	PushRegAddr      string
-	IsInternal       bool
+	SecretName string
+	SecretData *SecretData
+	PodMeta    *RegistryPodMeta
 }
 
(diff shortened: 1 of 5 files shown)


internal/registry/config.go
```diff
@@ -4,47 +4,87 @@ import (
...
+	SecretName string
+	SecretData *SecretData
+	PodMeta    *RegistryPodMeta
+}
+
+func GetConfig(ctx context.Context, client kube.Client) (*RegistryConfig, error) {
+	var secretConfig *SecretData
+	var secretName string
+
+	dockerRegistry, err := getServedDockerRegistry(ctx, client.Dynamic())
```
anoipm, line 31: IMO this part of code is hard to read. Whole method is long.

internal/registry/config.go
```diff
@@ -114,3 +154,24 @@ func isPodReady(pod corev1.Pod) bool {
...
+	}
+
+	for _, item := range list.Items {
+		var dockerRegistry DockerRegistry
+		err = kube.FromUnstructured(&item, &dockerRegistry)
+		if err != nil {
+			return nil, err
+		}
+
+		if dockerRegistry.Status.Served == "True" {
```
anoipm, line 171: Why you look on `Served`? IMO better could be `State == "Ready"`.

internal/registry/types.go
```diff
@@ -0,0 +1,26 @@
+package registry
+
+import (
+	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
+	"k8s.io/apimachinery/pkg/runtime/schema"
+)
+
+var (
+	DockerRegistryGVR = schema.GroupVersionResource{
```
anoipm, line 9: Now we have file `internal/btp/operator/gvr.go`. Isn't it good place for this definition?

pPrecel: internal/btp package is related to the `btp` platform. And all sub-directories should be have the same destiny. Of course I can move these lines into the `internal/registry/gvr.go` file, but I think `types.go` file is generic enough

internal/registry/config.go
pPrecel: done

internal/registry/config.go
pPrecel: done

internal/registry/config.go
pPrecel: good point, thanks

PR #2091, opened 2024-05-15
Implement own interface for the `clierror` struct

pPrecel:
**Description**

- as in the title
- disable possibility to wrap `clierrors` by the `errors.Wrap` func - it's not supported and may cause really weird situations
- implement own `clierror.Check` func that prints error and exits application with status `1` if error exists

**Related issue(s)**

- depends on https://github.com/kyma-project/cli/pull/2090


pPrecel:
changes are in [this PR](https://github.com/kyma-project/cli/pull/2092)

PR #2112, opened 2024-05-22
New Kyma Access CMD - Working ver

Cortey:
**Description**
Adds a command that creates a kubeconfig that includes token and certificate for specially crafted service account with admin access rights

**Related issue(s)**

- #2036
https://github.com/kyma-project/cli/pull/2119

### DIFF
--- a/internal/cmd/access/access.go
+++ b/internal/cmd/access/access.go
@@ -12,6 +12,8 @@ import (
 	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
 	"k8s.io/client-go/tools/clientcmd"
 	"k8s.io/client-go/tools/clientcmd/api"
+	"strconv"
+	"strings"
 )
 
 type accessConfig struct {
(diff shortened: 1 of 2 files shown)


internal/cmd/access/access.go
```diff
@@ -84,28 +90,40 @@ func runAccess(cfg *accessConfig) clierror.Error {
 	return nil
 }
 
-func prepareKubeconfig(cfg *accessConfig) (*api.Config, error) {
-	secret, err := cfg.KubeClient.Static().CoreV1().Secrets(cfg.namespace).Get(cfg.Ctx, cfg.name, metav1.GetOptions{})
+func prepareKubeconfig(cfg *accessConfig) (*api.Config, clierror.Error) {
+	currentCtx := cfg.KubeClient.ApiConfig().CurrentContext
+	clusterName := cfg.KubeClient.ApiConfig().Contexts[currentCtx].Cluster
+	token, duration, err := getServiceAccountToken(cfg)
```
kwiatekus, line 96: shouldn't the token be requested like this only if permanent!=true ?

Cortey: Adressed in followup PR

Issue #2093, opened 2024-05-15
POC: Command to access btp kyma runtime with workflow OIDC token


kwiatekus:
**Description**

Prototype command(s) allowing to generate a kubeconfig file from reference to kyma instance and user provided GH token. Validate ( with the prototype) how cli can get all the data (environment ID) to be able to download kubeconfig file for a provisioned (via CLI) cluster and produce a changed version of kubeconfig file that includes a custom token.

![Screenshot 2024-05-15 at 13 40 33](https://github.com/kyma-project/cli/assets/6783567/cf655e6e-d2c3-4114-8da5-df65f7d20c32)

Assumptions:
Kyma cli can produce a kubeconfig file for a selected kyma runtime (provisioned via kyma cli):
   - kyma cli should retrieve kyma environment details (using ENVIRONMENT ID ). Learn from where we could get environment ID in the cli. Maybe provisioning command should get it from CIS and  return it.
   - kyma cli should be able to download kubeconfig file based on the environment details (see attached btp mission 👇 )
   - kyma cli should replace the user context of the downloaded file with the token that is passed as a command option

The following is a proposed draft of the flow. We should be able to test it on DEV landscape where we can manually (*) manipulate an SKR so that it will grant access when presented with GH token:

1) `kyma provision --cluster-name=foo --plan --region --credentails ..` (in the future (*) here the extra config for OIDC issuer will be provided that will recognise the token )
2) (*) This is needed only temporarily - until necessary changes are introduced in kyma provisioning  
  a) Open gardener dashboard on DEV locate the cluster created in point 1 and enable the [`shoot-oidc-extension`](https://github.com/pPrecel/gardener-oidc-extension-poc/blob/main/.github/workflows/setup-serverless-on-gardener.yaml#L48)
  b)In the shoot cluster apply OIDC resource for github issuer (poc [reference](https://github.com/pPrecel/gardener-oidc-extension-poc/blob/main/.github/workflows/setup-serverless-on-gardener.yaml#L115))
  c) in the shoot cluster apply cluster role binding to cluster-admin role for the github identity (identified by org, repo, ref) (poc [reference](https://github.com/pPrecel/gardener-oidc-extension-poc/blob/main/.github/workflows/setup-serverless-on-gardener.yaml#L159))
3) `kyma access --with-token={token} --kyma-env-id={ENVIRONMENT ID AS RECOGNISED BY CIS} (or --cluster-name) --output=./action-kubeconfig.yaml`
[...]

PR #2121, opened 2024-05-28
Move kubeconfig saving/printing to a common package

halamix2:
**Description**

- Move kubeconfig saving/prointing to a common package

**Related issue(s)**

- #2093

### DIFF
--- a/internal/cmd/access/access.go
+++ b/internal/cmd/access/access.go
@@ -5,6 +5,7 @@ import (
 
 	"github.com/kyma-project/cli.v3/internal/clierror"
 	"github.com/kyma-project/cli.v3/internal/cmdcommon"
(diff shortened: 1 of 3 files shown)


internal/cmd/access/access.go
```diff
@@ -80,6 +81,10 @@ func runAccess(cfg *accessConfig) clierror.Error {
 		fmt.Println(string(message))
 
 	}
+	err = kube.SaveConfig(enrichedKubeconfig, cfg.output)
```
anoipm, line 84: Do we save config twice (first time in rows 70-83, second in 84-87)?

halamix2: I forgot to remove it from this command, fixed

anoipm (approved):

Issue #2117, opened 2024-05-27
POC: Kyma Modules command to list modules


kwiatekus:
**Description**
Prototype a new CLI command(s) that shows a module catalog, that is a list of all known modules. Implement classification of modules to distinguish offered modules from installed modules. From the installed ones, distinguish managed ones from non-managed ones. Implement filtering, ie: 

 - [x] `kyma modules --catalog ` - The list should list all available modules (with repo link and latest version)
  - [x]  `kyma modules --managed` - (or similar) - shows only those modules that are managed via central control-plane (those listed in `Kyma` CR) [Managed only]
 - [x]  `kyma modules --installed` - (or similar) - shows only the modules that are installed (doesnt show modules offered from catalog that are not currently installed) [Both managed and unmanaged]
 - [x] propose how to project all three classifications (catalog, installed, managed) in one list (`kyma modules`) : -->`kyma modules`
  
Until we have a final module catalog we can use this one: https://raw.githubusercontent.com/kyma-project/community-modules/main/model.json

Detection of installed modules should be based on the workload labels ( `kyma-project.io/module={module_name}` https://github.com/kyma-project/community/issues/864) We should verify consistency across all modules first.

**Reason**
A Kyma user needs a concise, comprehensive information about a catalog of modules installable in his kyma runtime, as well as those which are already installed (in which version) and weather the installed module is managed via control plane centrally or not.

kwiatekus:
As agreed before `kyma  alpha modules` should be projecting additional context regarding installed and managed modules on the all availabe module catalog.

When I run this command against a non managed kyma instance (hence no Kyma CR) this command doesnt show me any of the remaining info. It would be much better to skip this axtra context on non managed clusters and show what modules are installed projected on the list of all available ones.

Throwing error eliminates any usage on self-managed k8s clusters:
``` 
Error:
  failed to list modules

Error Details:
  failed to get managed Kyma modules: while getting managed modules: while getting Kyma CR: the server could not find the requested resource`
```
Please change that

kwiatekus:
```
 ./bin/kyma@v3 alpha modules 
          NAME          |                                 REPOSITORY                                  | VERSION INSTALLED | CONTROL-PLANE
------------------------+-----------------------------------------------------------------------------+-------------------+----------------
        telemetry       | https://github.com/kyma-project/telemetry-manager.git                       |                   |
------------------------+-----------------------------------------------------------------------------+-------------------+----------------
```

[...]

PR #2129, opened 2024-06-05
Add "managed" flag to "modules" command

Cortey:
**Description**
Added flag to list all modules managed by the central control plane in the cluster

**Related issue(s)**
https://github.com/kyma-project/cli/issues/2117

### DIFF
--- a/internal/cmd/modules/modules.go
+++ b/internal/cmd/modules/modules.go
@@ -41,6 +41,7 @@ func NewModulesCMD(kymaConfig *cmdcommon.KymaConfig) *cobra.Command {
 			clierror.Check(runModules(&cfg))
 		},
 	}
+	cfg.KubeClientConfig.AddFlag(cmd)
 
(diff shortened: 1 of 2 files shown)


internal/cmd/modules/modules.go
```diff
@@ -89,3 +114,54 @@ func listAllModules() ([]string, clierror.Error) {
...
+		}
+
+		modules, ok := spec["f:modules"].(map[string]interface{})
+		if !ok {
+			continue
+		}
+
+		for key := range modules {
+			if strings.Contains(key, "name") {
+				name := strings.TrimPrefix(key, "k:{\"name\":\"")
```
anoipm, line 160: Is it safe to use string operation on such structures? What will be done for string `"k : { "name" : "abcd" }` (very similar but with spaces)? Maybe we are sure that this string always ha fixed structure and this solution is ok?

Cortey: There could never be a space in there so we think it is okay

anoipm (approved):

PR #2147, opened 2024-06-25
Move v3 command prototypes into alpha command group

### LINKED ISSUE #2143: Move all new v3 commands into alpha group
**Description**

Move new command prototypes into alpha group:
 - kyma alpha provision ..
 - kyma alpha hana provision ...
 -  etc

**Reasons**

Make it clear for the early adopters that some new commands are still prototypes

**Attachments**

kwiatekus:
**Description**

- Move new commands to `alpha` command group

**Related issue(s)**
Fixes #2143


internal/cmd/kyma.go
```diff
@@ -33,14 +20,6 @@ func NewKymaCMD() *cobra.Command {
...
-
-	cmd.AddCommand(hana.NewHanaCMD(config))
-	cmd.AddCommand(imageimport.NewImportCMD(config))
-	cmd.AddCommand(provision.NewProvisionCMD())
-	cmd.AddCommand(referenceinstance.NewReferenceInstanceCMD(config))
-	cmd.AddCommand(access.NewAccessCMD(config))
-	cmd.AddCommand(oidc.NewOIDCCMD(config))
-	cmd.AddCommand(modules.NewModulesCMD(config))
-
+	cmd.AddCommand(alpha.NewAlphaCMD())
```
anoipm, line 23: Do we want only `alpha` commands now?

internal/cmd/alpha/alpha.go
```diff
@@ -0,0 +1,39 @@
+package alpha
```
anoipm, line 1: I need discuss about moving all code for commands to `alpha` directory. I'm not sure that it's necessary.

internal/cmd/kyma.go
anoipm: We decided that now all commands will be `alpha`

internal/cmd/alpha/alpha.go
anoipm: We decided that directory `alpha` is ok for us.

anoipm (approved):

PR #2151, opened 2024-06-26
Partial tests for Community Modules package

Cortey:
**Description**

- ...
- ...
- ...

**Related issue(s)**
#2117

### DIFF
--- a/internal/communitymodules/merge_test.go
+++ b/internal/communitymodules/merge_test.go
@@ -1,0 +1,60 @@
+package communitymodules
+
+import (
+	"github.com/stretchr/testify/require"
+	"testing"
+)
+
+func TestMergeRowMaps(t *testing.T) {
+	moduleMapCatalog := moduleMap{
+		"serverless": {
+			Name:       "serverless",
+			Repository: "github.com/kyma-project/serverless",
+		},
+	}
+	moduleMapManaged := moduleMap{
+		"serverless": {
+			Name:    "serverless",
+			Managed: "Managed",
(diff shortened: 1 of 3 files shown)


internal/communitymodules/merge_test.go
```diff
@@ -0,0 +1,60 @@
+package communitymodules
+
+import (
+	"github.com/stretchr/testify/require"
+	"testing"
+)
+
+func TestMergeRowMaps(t *testing.T) {
```
anoipm, line 8: I can't see tests for other cases i.e:
- more than one key inside `moduleMap`,
- merging maps with some different keys.

Now such data are probably impossible but we don't know about future. Tested module is generic library that should support such cases.

internal/communitymodules/render_test.go
```diff
@@ -0,0 +1,33 @@
+package communitymodules
+
+import (
+	"github.com/stretchr/testify/require"
+	"testing"
+)
+
+func TestRender(t *testing.T) {
```
anoipm, line 8: What if `convertRow...` doesn't work for `moduleMap` with not one row?

internal/communitymodules/merge_test.go
Cortey: Added case covering those

internal/communitymodules/render_test.go
Cortey: Added cases for those

anoipm (approved):

PR #2146, opened 2024-06-25
restore go linter job

halamix2:
**Description**

Changes proposed in this pull request:

- restore go linter job

**Related issue(s)**

- #2142

PR #2159, opened 2024-06-28
Lint fixes

halamix2:
**Description**

- fix some of the linting issues

**Related issue(s)**

- #2146

### DIFF
--- a/internal/btp/cis/client.go
+++ b/internal/btp/cis/client.go
@@ -72,14 +72,6 @@ func (c *httpClient) post(url string, options requestOptions) (*http.Response, e
 	return c.genericRequest(http.MethodPost, url, options)
 }
 
-func (c *httpClient) put(url string, options requestOptions) (*http.Response, error) {
-	return c.genericRequest(http.MethodPut, url, options)
-}
-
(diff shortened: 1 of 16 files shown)


internal/btp/cis/client.go
```diff
@@ -72,14 +72,6 @@ func (c *httpClient) post(url string, options requestOptions) (*http.Response, e
 	return c.genericRequest(http.MethodPost, url, options)
 }
 
-func (c *httpClient) put(url string, options requestOptions) (*http.Response, error) {
```
anoipm, line 75: Do we should remove these (probably unused) methods? IMO this client is like library and can have methods for potential use.

halamix2: I'd delete them, because I see two solutions:
- add linter ignore comment
    - when these function will get used, we may forget to remove this comment
- remove these functions
    - when somebody would want to use them, they would need to re-add the function (basically copy-paste of the post function)
Since this is a part of CIS package, where 9at least for now) we won't use any put or patch (or delete) methods, I'd remove these

anoipm (approved):

PR #2164, opened 2024-07-01
add test for modules managed command

anoipm:
**Description**

- add test for modules managed command

**Related issue(s)**
See also #2117

### DIFF
--- a/internal/communitymodules/modules_test.go
+++ b/internal/communitymodules/modules_test.go
@@ -1,7 +1,15 @@
 package communitymodules
 
 import (
+	"context"
+	"encoding/json"
+	"github.com/kyma-project/cli.v3/internal/cmdcommon"
+	kube_fake "github.com/kyma-project/cli.v3/internal/kube/fake"
+	"github.com/stretchr/testify/assert"
 	"github.com/stretchr/testify/require"
+	"k8s.io/apimachinery/pkg/apis/meta/v1/unstructured"
+	"k8s.io/apimachinery/pkg/runtime"
+	dynamic_fake "k8s.io/client-go/dynamic/fake"
 	"net/http"
 	"net/http/httptest"
 	"testing"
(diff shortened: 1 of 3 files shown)


internal/communitymodules/modules_test.go
```diff
@@ -86,9 +94,125 @@ func Test_modulesCatalog(t *testing.T) {
 	})
 }
 
+func Test_ManagedModules(t *testing.T) {
+	t.Run("ok", func(t *testing.T) {
+		expectedResult := moduleMap{
+			"module1": row{
+				Name:    "module1",
+				Managed: "Managed",
```
MichalKalke, line 102: Please change it to `True` as the new value of the `MANAGED` field

internal/communitymodules/modules_test.go
```diff
@@ -86,9 +94,125 @@ func Test_modulesCatalog(t *testing.T) {
...
+			},
+			"module2": row{
+				Name:    "module2",
+				Managed: "Managed",
```
MichalKalke, line 106: As above

internal/communitymodules/modules_test.go
```diff
@@ -86,9 +94,125 @@ func Test_modulesCatalog(t *testing.T) {
...
+			},
+			"module3": row{
+				Name:    "module3",
+				Managed: "Managed",
```
MichalKalke, line 110: As above

internal/communitymodules/modules_test.go
anoipm: done

internal/communitymodules/modules_test.go
anoipm: done

internal/communitymodules/modules_test.go
anoipm: done

MichalKalke (approved):

Issue #2217, opened 2024-09-12
Provide new command that outputs `config.json` file to be used in docker CLI against internal docker registry


kwiatekus:
**Description**

Add a new command that would output docker registry config in form of docker's `config.json` containing kyma's docker registry external url and auth data:
```
{
	"auths": {
		"{external url of docker registry}": {
			"auth": "{base64 encoded `user:password`}"
		}
	}
}
```
If the docker registry secret is not present (i.e docker-registry module is not installed), then output information saying that there is no docker registry module installed.

```sh
kyma alpha registry config --dockerconfig (--output config.json)
URL = $(kyma alpha registry config --externalurl )
docker --config {PATH to config.json} push ${URL}/my-app:{tag}
```

**Reasons**

Kyma can be provisioned with internal docker registry.
The registry will prove useful only if it is easy to use it with docker CLI locally or in a CI workflow context. 

**Attachments**
https://lumaks.medium.com/creating-docker-config-json-for-use-in-external-systems-a5d14aa81c41

**Related topics**:
https://github.com/kyma-project/kyma/issues/18555
https://github.com/kyma-project/kyma/issues/18198

kwiatekus:
Works.
It is already used in the sample  pipeline for kyma automation

PR #2220, opened 2024-09-26
Registry config command

Cortey:
**Description**

- Added a command to save registry config.json to a local file

**Related issue(s)**
https://github.com/kyma-project/cli/issues/2217

### DIFF
--- a/internal/cmd/alpha/registry/config/config.go
+++ b/internal/cmd/alpha/registry/config/config.go
@@ -1,0 +1,65 @@
+package config
+
+import (
+	"fmt"
+	"github.com/kyma-project/cli.v3/internal/clierror"
+	"github.com/kyma-project/cli.v3/internal/cmdcommon"
+	"github.com/kyma-project/cli.v3/internal/registry"
+	"github.com/spf13/cobra"
+	"os"
+)
+
+type cfgConfig struct {
+	*cmdcommon.KymaConfig
+	cmdcommon.KubeClientConfig
+
+	dockerconfig bool
+	externalurl  bool
+	output       string
+}
+
+func NewConfigCMD(kymaConfig *cmdcommon.KymaConfig) *cobra.Command {
+	cfg := cfgConfig{
+		KymaConfig:       kymaConfig,
+		KubeClientConfig: cmdcommon.KubeClientConfig{},
+	}
+
+	cmd := &cobra.Command{
+		Use:   "config",
+		Short: "Saves Kyma registry dockerconfig to a file",
+		Long:  "Use this command to save Kyma registry dockerconfig to a file",
+		PreRun: func(_ *cobra.Command, _ []string) {
+			clierror.Check(cfg.KubeClientConfig.Complete())
+		},
+		Run: func(_ *cobra.Command, _ []string) {
+			clierror.Check(runConfig(&cfg))
+		},
+	}
+
+	cfg.KubeClientConfig.AddFlag(cmd)
+	cmd.Flags().BoolVar(&cfg.dockerconfig, "dockerconfig", false, "Generate a docker config.json file for the Kyma registry")
+	cmd.Flags().BoolVar(&cfg.externalurl, "externalurl", false, "External URL for the Kyma registry")
+	cmd.Flags().StringVar(&cfg.output, "output", "config.json", "Path where the output file should be saved to")
+
+	return cmd
+}
+
+func runConfig(cfg *cfgConfig) clierror.Error {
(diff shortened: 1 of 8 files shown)


internal/cmd/alpha/registry/config/config.go
```diff
@@ -0,0 +1,65 @@
...
+	cfg.KubeClientConfig.AddFlag(cmd)
+	cmd.Flags().BoolVar(&cfg.dockerconfig, "dockerconfig", false, "Generate a docker config.json file for the Kyma registry")
+	cmd.Flags().BoolVar(&cfg.externalurl, "externalurl", false, "External URL for the Kyma registry")
+	cmd.Flags().StringVar(&cfg.output, "output", "config.json", "Path where the output file should be saved to")
+
+	return cmd
+}
+
+func runConfig(cfg *cfgConfig) clierror.Error {
+	registryConfig, err := registry.GetConfig(cfg.Ctx, cfg.KubeClient)
```
pPrecel, line 48: We can't use the `registry.GetConfig` func here because it returns secrets containing data used for local (in-cluster) connection, but users may expect config they can use for external connection.

internal/cmd/alpha/registry/config/config.go
```diff
@@ -0,0 +1,65 @@
...
+	if cfg.dockerconfig {
+		writeErr := os.WriteFile(cfg.output, []byte(registryConfig.SecretData.DockerConfigJSON), os.ModePerm)
+		if writeErr != nil {
+			return clierror.New("failed to write docker config to file")
+		}
+		fmt.Print("Docker config saved to ", cfg.output)
+	}
+
+	if cfg.externalurl {
+		fmt.Print(registryConfig.SecretData.PushRegAddr)
```
pPrecel, line 62: I would suggest using `fmt.Println`, or adding `\n` at the end of the format because when I've specified both flags the output looks weird:

```bash
./kyma-cli alpha registry config --dockerconfig --externalurl

Docker config saved to config.jsondockerregistry.kyma-system.svc.cluster.local:5000
```

internal/cmd/alpha/registry/config/config.go
```diff
@@ -0,0 +1,65 @@
...
+		PreRun: func(_ *cobra.Command, _ []string) {
+			clierror.Check(cfg.KubeClientConfig.Complete())
+		},
+		Run: func(_ *cobra.Command, _ []string) {
+			clierror.Check(runConfig(&cfg))
+		},
+	}
+
+	cfg.KubeClientConfig.AddFlag(cmd)
+	cmd.Flags().BoolVar(&cfg.dockerconfig, "dockerconfig", false, "Generate a docker config.json file for the Kyma registry")
```
pPrecel, line 40: I would suggest removing this flag and dealing with its behavior as a default for the `registry config` command. Right now the `registry config` default is useless without passing additional flags which may be really confusing

internal/cmd/alpha/registry/config/config.go
```diff
@@ -0,0 +1,65 @@
...
+	cmd.Flags().BoolVar(&cfg.externalurl, "externalurl", false, "External URL for the Kyma registry")
```
pPrecel, line 41: I would suggest replacing the `--externalurl` flag with the `--external-url` one.

Also, the description is confusing because it does not describe what is causing this flag. Maybe this would be better:

```
Print docker registry external URL
```

pPrecel: In general, I don't get the goal of this flag, because it's not cooperating with the `--output` flag and exactly the same information will be located in the `config.json` after running the `registry config --dockerconfig` command. Maybe we should remove this flag? or move it to a command command like `registry external-up`? or I don't know...

But this is only a suggestion. For sure this flag should be re-designed

pPrecel (changes requested):

internal/cmd/alpha/registry/config/config.go
Cortey: Done

internal/cmd/alpha/registry/config/config.go
Cortey: Fixed

internal/cmd/alpha/registry/config/config.go
Cortey: Flag removed

internal/cmd/alpha/registry/config/config.go
Cortey: Done, can't do both at once now

pPrecel (approved):

Issue #2239, opened 2024-11-08
Create k8s service when pushing application via kyma cli


kwiatekus:
**Description**

Create k8s svc instance for workloads pushed via kyma cli:

For example
`kyma alpha app push --name mynginx --image nginx:latest --containerPort xx`
should also create k8s service that would enable routing to the exposed container port of the workload. 

**Reasons**
k8s svc is an important k8s resource essential to internal k8s routing.

Part of https://github.com/kyma-project/cli/issues/2236
Depends on https://github.com/kyma-project/cli/issues/2238

PR #2250, opened 2024-11-19
Create service when pushing an application via `kyma alpha app push --containerport`

Cortey:
**Description**
Creates a service with a container port exposed when using the --containerport flag

**Related issue(s)**
https://github.com/kyma-project/cli/issues/2239


halamix2 (approved):
/pony

PR #2267, opened 2024-11-29
feat: Add dry-run flag to prevent pushing module template to registry

KacperMalachowski:
**Description**

In neighbors team, we need to validate only the possibility to push module template to registry.
It's required to write correct check for submission pipeline. See: https://github.tools.sap/kyma/test-infra/issues/469

- Add `dry-run` flag to `kyma alpha create module` to preventing push of module template

**Related issue(s)**

See: https://github.tools.sap/kyma/test-infra/issues/469


Sawthis:
The content of the default directory is now in https://github.com/kyma-project/template-operator/tree/main/config/overlays/deployment .

CC: @KacperMalachowski @nesmabadr

KacperMalachowski:
FIxing the e2e tests (at least at level of my knowledge for cli) without breaking it require code breaking change of: https://github.com/kyma-project/cli/blob/release-2.20/pkg/module/kubebuilder/project.go#L21

I really doubt if we can safely fix it for now.
CC: @Sawthis

PR #2320, opened 2025-01-20
Add the `alpha <resource> create` command

pPrecel:
**Description**

- make the `create` template command

**How to run it**

1. add the following configmap to the cluster:

```yaml
apiVersion: v1
data:
  rootCommand: |
    name: function
    description: Set of commands for managing Functions
    descriptionLong: Use this command manage Functions
  resource: |
    scope: namespace
    kind: Function
    version: v1alpha2
    group: serverless.kyma-project.io
  templateCommands: |
    create:
      description: create test resource
      descriptionLong: use this command to create test resource
      customFlags:
      - type: "string"
        name: "runtime"
        description: "function runtime"
        shorthand: "r"
        path: ".spec.runtime"
        default: "nodejs22"
      - type: "int64"
        name: "replicas"
        description: "function replicas"
        path: ".spec.replicas"
        default: 1
      - type: "path"
        name: "source"
        description: "function source file path"
        shorthand: "s"
        path: ".spec.source.inline.source"
        default: |
          module.exports = {
            main: function(event, context) {
              return 'Hello World!'
            }
          }
      - type: "path"
        name: "dependencies"
        description: "function dependencies file path"
        shorthand: "d"
        path: ".spec.source.inline.dependencies"
kind: ConfigMap
metadata:
  labels:
    kyma-cli/extension: resource
    kyma-project.io/module: function
  name: serverless.function.cli
  namespace: kyma-system

```

2. build the cli from this PR:

```bash
go build --output kyma-dev main.go
```

3. use command

```bash
./kyma-dev alpha --help
```

**Related issue(s)**

- https://github.com/kyma-project/serverless/issues/1247

### DIFF
--- a/internal/cmd/alpha/templates/parameters/default.go
+++ b/internal/cmd/alpha/templates/parameters/default.go
@@ -1,0 +1,26 @@
+package parameters
+
+// this func makes sure that types used as an default follows YAML standards
+// converts:
+// int8,int16,int32,int64,int to int64
+// string, []byte to string
+func sanitizeDefaultValue(defaultValue interface{}) interface{} {
+	switch value := defaultValue.(type) {
+	case int8:
+		return int64(value)
+	case int16:
+		return int64(value)
+	case int32:
+		return int64(value)
+	case int64:
+		return value
+	case int:
+		return int64(value)
+	case string:
+		return value
+	case []byte:
+		return string(value)
+	default:
+		return nil
+	}
+}
--- a/internal/cmd/alpha/templates/types/create.go
+++ b/internal/cmd/alpha/templates/types/create.go
@@ -1,0 +1,35 @@
+package types
+
+type CreateCommand struct {
+	// short description of the command
+	Description string `yaml:"description"`
+	// long description of the command group
+	DescriptionLong string `yaml:"descriptionLong"`
+	// custom flags used to build command and set values for specific fields
+	CustomFlags []CreateCustomFlag `yaml:"customFlags"`
+}
+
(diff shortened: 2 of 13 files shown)


internal/cmd/alpha/templates/parameters/default.go
```diff
@@ -0,0 +1,26 @@
...
+		return value
+	case int:
+		return int64(value)
+	case string:
+		return value
+	case []byte:
+		return string(value)
+	default:
+		return nil
+	}
```
halamix2, line 25: do we need to convert bools to `true/false` as well?

internal/cmd/alpha/templates/types/create.go
```diff
@@ -0,0 +1,35 @@
...
+	// description of the custom flag
+	Description string `yaml:"description"`
+	// optional shorthand of the custom flag
+	Shorthand string `yaml:"shorthand"`
+	// path to the resource fild that will be updated with value from the flag
+	Path string `yaml:"path"`
+	// default value for the flag
+	DefaultValue interface{} `yaml:"default"`
+	// mark if flag is required
+	Required bool `yaml:"required"`
```
halamix2, line 34: Do we want to add `omitempty` here to at least signify this field is optional?

internal/cmd/alpha/templates/parameters/default.go
pPrecel: not right now because we do not support this kind while setting custom fields. We should add this together with new functionality

halamix2 (approved):

Issue #2325, opened 2025-01-21
Remove redundant `--module` flag for `manage/unmanage` commands


kwiatekus:
**Description**

Commands to manage/unmanage module offers different way to specify target  module than commands to  add/delete module 

`kyma@v3 alpha module add keda --default-cr`
vs
`kyma@v3 alpha module manage --module keda`
`kyma@v3 alpha module unmanage --module keda`

Remove the `--module` flag for `manage/unmanage` commands as it is redundant.
Make it:
```
`kyma@v3 alpha module manage keda`
`kyma@v3 alpha module unmanage keda`
```

**Reasons**

Consistency in this matter will be good for user experience

**Attachments**

kwiatekus:
Please print out msg that confirms succesful operation:

"Module <module name> set to un-/managed"

PR #2334, opened 2025-01-23
Add succesful operation message after manage/unmanage

FilipRudy:
**Description**

- As in title

**Related issue(s)**

- #2325


internal/cmd/alpha/module/manage.go
```diff
@@ -59,5 +59,8 @@ func runManage(cfg *manageConfig) clierror.Error {
 	if err != nil {
 		return clierror.Wrap(err, clierror.New("failed to check module state"))
 	}
+
+	fmt.Printf("Module %s set to managed\n\n", cfg.module)
```
pPrecel, line 63: I would suggest using a single new line char to make output consistent with other commands

```suggestion
	fmt.Printf("Module %s set to managed\n", cfg.module)
```

internal/cmd/alpha/module/unmanage.go
```diff
@@ -46,5 +47,8 @@ func runUnmanage(cfg *unmanageConfig) clierror.Error {
 	if err != nil {
 		return clierror.Wrap(err, clierror.New("failed to check module state"))
 	}
+
+	fmt.Printf("Module %s set to unmanaged\n\n", cfg.module)
```
pPrecel, line 51: I would suggest using a single new line char to make output consistent with other commands
```suggestion
	fmt.Printf("Module %s set to unmanaged\n", cfg.module)
```

pPrecel (approved):

Issue #2330, opened 2025-01-21
List installed modules with statuses


kwiatekus:
**Description**

Show installed modules based on Kyma CR's status.

**Reasons**

Listing modules should in the first place list installed modules.
Listing available modules should be moved to separate command (or triggered by option).

Comparing UX to busola: user first sees modules installed in the cluster. Showing catalog is triggered by an extra action (`Edit` shows available modules to be added)

**Ideas**

List all installed modules using Kyma CR (`status.modules`) as a source of truth with
```
kyma alpha module list
```

List availalable modules (module catalog) [using moduletemplates and modulereleasemeta CRs](https://github.com/kyma-project/cli/blob/main/internal/modules/list.go#L47-L55), enriched with information about installed versions (if any) with:

```
kyma alpha module catalog 
```

PR #2344, opened 2025-02-12
Replace render table lib

pPrecel:
**Description**

- replace lib that allows us to render tables (the previous one was too restrictive) to the [jedib0t/go-pretty
](https://github.com/jedib0t/go-pretty/tree/main/table)
- improve `kyma alpha <resource> get` command to print namespace optionally only if user use the `-A` flag
- add the `-n` shorthand for the `--namespace` flag

**Related issue(s)**

- https://github.com/kyma-project/cli/issues/2330


internal/render/table.go
```diff
@@ -3,28 +3,44 @@ package render
...
+	}
 )
 
 // Table renders the table with the provided headers and data
-func Table(writer io.Writer, modulesData [][]string, headers []string) {
-	twTable := setTable(writer)
-	twTable.AppendBulk(modulesData)
-	twTable.SetHeader(headers)
-	twTable.Render()
+func Table(writer io.Writer, headers []interface{}, rows [][]interface{}) {
```
anoipm, line 29: Why do you use `interface{}` instead of `string`?

pPrecel: because new library requires such input data types

anoipm: ok

anoipm (approved):

PR #2504, opened 2025-06-03
Installation script default dir

pPrecel:
**Description**

- set defailt installation tmp dir if the `TMPDIR` env is empty


hack/install_cli_latest.sh
```diff
@@ -2,6 +2,7 @@
 
 set -e
 
+TMPDIR=${TMPDIR:-/tmp}
 cd ${TMPDIR}
```
halamix2: can we `cd` always or should we add `mkdir -p ${TMPDIR}` ?

pPrecel: I decided to get rid of the `cd` command

PR #2507, opened 2025-06-09
Add Community Module

### LINKED ISSUE #2466: Adding community module
**Description**

Add community module as part of `kyma alpha module add ` - estimate [5]

**AC**
   - If the module template does not have `operator.kyma-project.io/managed-by=kyma` label than CLI should apply `.spec.resources`  on the cluster
  - `--default-cr` and `--cr-path` flags handling should be the same as for non-community modules 
  - nn managed clusters inform the user that there is no binding SLA for community modules and prompt for confirmation (overridable by `--auto-approve`)
  - if there are multiple versions of communitu module template available on the cluster, user should be promted to pick a version (overridable by `--version`).

musztardem:
**Description**

- Implemented two prompting mechanisms:
Zrzut ekranu 2025-06-11 o 11 22 41
Zrzut ekranu 2025-06-11 o 11 23 06
- Added `--community`, `--auto-approve` and `--version` flags to `module add` command
- Implemented mechanisms that installs community modules (not managed by the lifecycle manager) on the cluster

**Related issue(s)**

Resolves #2466

### DIFF
--- a/internal/cmdcommon/prompt/bool.go
+++ b/internal/cmdcommon/prompt/bool.go
@@ -26,12 +26,16 @@ func NewBool(message string, defaultValue bool) *Bool {
 func (b *Bool) Prompt() (bool, error) {
 	var userInput string
 	fmt.Fprintf(b.writer, "%s %s: ", b.message, b.defaultValueDisplay())
-	_, err := fmt.Fscan(b.reader, &userInput)
+	_, err := fmt.Fscanln(b.reader, &userInput)
 
 	// If the user just presses Enter, Fscan returns the EOF error
 	if err != nil && err == io.EOF {
 		// Treat as empty input, use default value
 		return b.defaultValue, nil
+	}
+
+	if err != nil {
+		return false, err
 	}
 
 	parsedUserInput, err := b.validateUserInput(userInput)
--- a/internal/cmdcommon/prompt/bool_test.go
+++ b/internal/cmdcommon/prompt/bool_test.go
@@ -2,7 +2,10 @@ package prompt
 
 import (
 	"bytes"
+	"errors"
+	"io"
 	"testing"
+	"testing/iotest"
 
 	"github.com/stretchr/testify/require"
 )
@@ -10,68 +13,75 @@ import (
 func TestBoolPrompt_Table(t *testing.T) {
 	tests := []struct {
 		name         string
-		input        string
+		inputReader  io.Reader
 		defaultValue bool
 		expectResult bool
 		expectErr    string
 	}{
 		{
 			name:         "Yes short",
-			input:        "y\n",
+			inputReader:  bytes.NewBufferString("y\n"),
 			defaultValue: true,
 			expectResult: true,
 		},
 		{
 			name:         "Yes long",
-			input:        "yes\n",
+			inputReader:  bytes.NewBufferString("yes\n"),
 			defaultValue: false,
 			expectResult: true,
 		},
 		{
 			name:         "No short",
-			input:        "n\n",
+			inputReader:  bytes.NewBufferString("n\n"),
 			defaultValue: true,
 			expectResult: false,
 		},
 		{
 			name:         "No long",
-			input:        "no\n",
+			inputReader:  bytes.NewBufferString("no\n"),
 			defaultValue: true,
 			expectResult: false,
 		},
 		{
 			name:         "Default true with empty input",
-			input:        "\n",
+			inputReader:  bytes.NewBufferString(""),
 			defaultValue: true,
 			expectResult: true,
 		},
 		{
 			name:         "Default true with whitespaces",
-			input:        "  \t\n",
+			inputReader:  bytes.NewBufferString("  \t\n"),
 			defaultValue: true,
-			expectResult: true,
+			expectResult: false,
+			expectErr:    "unexpected newline",
 		},
 		{
 			name:         "Default false with empty input",
-			input:        "\n",
+			inputReader:  bytes.NewBufferString(""),
 			defaultValue: false,
 			expectResult: false,
 		},
 		{
 			name:         "Invalid input",
-			input:        "maybe\n",
+			inputReader:  bytes.NewBufferString("maybe\n"),
 			defaultValue: false,
 			expectResult: false,
 			expectErr:    "invalid input, please enter 'y' or 'n'",
+		},
+		{
+			name:         "Erroneous input",
+			inputReader:  iotest.ErrReader(errors.New("test error")),
+			defaultValue: true,
+			expectResult: false,
+			expectErr:    "test error",
 		},
 	}
 
 	for _, tc := range tests {
 		t.Run(tc.name, func(t *testing.T) {
-			input := bytes.NewBufferString(tc.input)
 			output := bytes.NewBuffer([]byte{})
 			b := Bool{
-				reader:       input,
+				reader:       tc.inputReader,
 				writer:       output,
 				message:      "Proceed?",
 				defaultValue: tc.defaultValue,
--- a/internal/cmdcommon/prompt/oneofstringlist.go
+++ b/internal/cmdcommon/prompt/oneofstringlist.go
@@ -39,9 +39,13 @@ func NewCustomOneOfStringList(reader io.Reader, writer io.Writer, message, promp
 func (l *OneOfStringList) Prompt() (string, error) {
 	var userInput string
 	fmt.Fprintf(l.writer, "%s\n%s\n\n%s", l.message, l.valuesListString(), l.promptText)
-	_, err := fmt.Fscan(l.reader, &userInput)
+	_, err := fmt.Fscanln(l.reader, &userInput)
 	if err != nil && err == io.EOF {
 		return "", fmt.Errorf("no value was selected")
+	}
+
+	if err != nil {
+		return "", err
 	}
 
 	validatedUserInput, err := l.validateUserInput(userInput)
--- a/internal/cmdcommon/prompt/oneofstringlist_test.go
+++ b/internal/cmdcommon/prompt/oneofstringlist_test.go
@@ -2,7 +2,10 @@ package prompt_test
 
 import (
 	"bytes"
+	"errors"
+	"io"
 	"testing"
+	"testing/iotest"
 
 	"github.com/kyma-project/cli.v3/internal/cmdcommon/prompt"
 	"github.com/stretchr/testify/require"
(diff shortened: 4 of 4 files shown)


internal/modules/list.go
```diff
@@ -127,6 +127,29 @@ func ListCatalog(ctx context.Context, client kube.Client) (ModulesList, error) {
...
+	catalog, err := ListCatalog(ctx, client)
+	if err != nil {
+		return nil, err
+	}
+
+	var module Module
+
+	for _, m := range catalog {
+		if m.CommunityModule == isCommunity {
+			module = m
```
pPrecel, line 140: If you expect one result only, then I would suggest breaking the loop after this line:

```suggestion
			module = m
			break
```

internal/cmd/module/add.go
```diff
@@ -43,6 +50,9 @@ func newAddCMD(kymaConfig *cmdcommon.KymaConfig) *cobra.Command {
 	cmd.Flags().StringVarP(&cfg.channel, "channel", "c", "", "Name of the Kyma channel to use for the module")
 	cmd.Flags().StringVar(&cfg.crPath, "cr-path", "", "Path to the custom resource file")
 	cmd.Flags().BoolVar(&cfg.defaultCR, "default-cr", false, "Deploys the module with the default CR")
+	cmd.Flags().BoolVar(&cfg.autoApprove, "auto-approve", false, "Automatically approve community module installation")
+	cmd.Flags().StringVar(&cfg.version, "version", "", "Specify version of the community module to install")
+	cmd.Flags().BoolVar(&cfg.community, "community", false, "Install a community module (no official support, no binding SLA)")
```
pPrecel, line 56: This PR adds a new use case for the `add` command, so now some flags will be used for the specific case only. I would suggest updating [flags validation](https://github.com/kyma-project/cli/pull/2507/files#diff-88c49af705f117e097eae37c93c30c9e97ad814212bd1a7cda5acb652fb4dc39R41) to prevent incorrect input configurations. [Here](https://github.com/kyma-project/cli/blob/main/internal/cmd/alpha/app/push.go#L50-L57) is an example of another command having complex validation.

internal/cmdcommon/prompt/bool.go
```diff
@@ -0,0 +1,60 @@
...
+	_, err := fmt.Scanln(&userInput)
+
+	// If the user just presses Enter, Scanln returns an error, but userInput remains empty.
+	if err != nil && userInput == "" {
+		// Treat as empty input, use default value
+		return b.defaultValue, nil
+	}
+	if err != nil {
+		return false, err
+	}
```
pPrecel: what is the difference between these ifs? Is the userInput not empty in the second case? Maybe the error is some specific kind when user pressed Enter (EOF or something like this)?

internal/cmdcommon/prompt/bool.go
```diff
@@ -0,0 +1,60 @@
...
+	return &Bool{
+		message:      message,
+		defaultValue: defaultValue,
+	}
+}
+
+func (b *Bool) Prompt() (bool, error) {
+	var userInput string
+	fmt.Printf("%s %s: ", b.message, b.defaultValueDisplay())
+	_, err := fmt.Scanln(&userInput)
```
pPrecel: The `Scanln` func reads input from the stdin. If you want to simplify tests (replace stdin with test buffer/reader) I would suggest using the `Fscanln` func

internal/cmdcommon/prompt/bool.go
```diff
@@ -0,0 +1,60 @@
...
+	}
+	return "[y/N]"
+}
+
+func (b *Bool) validateUserInput(userInput string) (bool, error) {
+	switch strings.TrimSpace(strings.ToLower(userInput)) {
+	case "y":
+		return true, nil
+	case "n":
+		return false, nil
```
pPrecel: I would support adding hidden support for full words like `yes` and `no`

```suggestion
	case "y" || "yes":
		return true, nil
	case "n" || "no":
		return false, nil
```

internal/cmdcommon/prompt/bool_test.go
```diff
@@ -0,0 +1,90 @@
...
+func TestBoolPrompt_Yes(t *testing.T) {
+	b := prompt.NewBool("Proceed?", false)
+	withInput("y\n", func() {
+		result, err := b.Prompt()
+		if err != nil {
+			t.Fatalf("expected no error, got: %v", err)
+		}
+		if !result {
+			t.Errorf("expected true for 'y' input, got false")
+		}
```
pPrecel: The `"github.com/stretchr/testify/require"` may help you with checing final results

```suggestion
                 require.NoError(t, err)
                 require.True(t, result)
```

Same suggestion for all other tests

internal/cmdcommon/prompt/bool.go
pPrecel: Also, the error coming from this line is not tested. I would suggest test this scenario

internal/flags/validate.go
```diff
@@ -135,7 +135,7 @@ func MarkMutuallyExclusive(flags ...string) Rule {
 		})
 
 		if len(listChangedFlags(flagSet, flags...)) > 1 {
-			return fmt.Errorf("only one flag from groud [%s] can be used at the same time, used [%s]",
+			return fmt.Errorf("only one flag from group [%s] can be used at the same time, used [%s]",
```
pPrecel, line 138: thanks for catching this :)

internal/modules/install.go
```diff
@@ -0,0 +1,192 @@
...
+		return clierror.Wrap(err, clierror.New("failed to install community module"))
+	}
+
+	if err := applyDefaultCustomResource(ctx, client, existingModule, data.IsDefaultCRApplicable); err != nil {
+		return clierror.Wrap(err, clierror.New("failed to apply default custom resource"))
+	}
+
+	if err := applyCustomResourcesFromFile(ctx, client, data.CustomResources); err != nil {
+		return clierror.Wrap(err, clierror.New("failed to apply custom resource files"))
+	}
```
pPrecel: I would suggest moving this functionality to a separate function called `applyCustomResource`, and inside it, decide how and from which CR should be applied. Reading this code from top to bottom is quite confusing.

internal/modules/install.go
```diff
@@ -0,0 +1,192 @@
...
+			return err
+		}
+		installedResources = append(installedResources, installedResource)
+	}
+	return nil
+}
+
+func applyResourceWithRollback(ctx context.Context, client kube.Client, resourceYamlStr string, installedResources []map[string]any) (map[string]any, error) {
+	var obj map[string]any
+	if err := yaml.Unmarshal([]byte(resourceYamlStr), &obj); err != nil {
```
pPrecel: I would suggest unmarshalling all resources first and then applying them. It will simplify functionality because we will use rollback only on client error

internal/modules/install.go
```diff
@@ -0,0 +1,192 @@
...
+}
+
+func applyDefaultCustomResource(ctx context.Context, client kube.Client, existingModule *kyma.ModuleTemplate, isDefaultCRApplicable bool) error {
+	if !isDefaultCRApplicable {
+		return nil
+	}
+
+	defaultCustomResourceUnstructured := existingModule.Spec.Data
+
+	if err := client.RootlessDynamic().Apply(ctx, &defaultCustomResourceUnstructured, false); err != nil {
```
pPrecel, line 166: Should we rollback module resources when this func failed?

internal/modules/install.go
```diff
@@ -0,0 +1,192 @@
...
+		if rollbackErr != nil {
+			return nil, fmt.Errorf("failed to parse module resource: %w; rollback also failed: %v", err, rollbackErr)
+		}
+		return nil, fmt.Errorf("failed to parse module resource: %w", err)
+	}
+
+	if err := client.RootlessDynamic().Apply(ctx, &unstructured.Unstructured{Object: obj}, false); err != nil {
+		rollbackErr := rollback(ctx, client, installedResources)
+		if rollbackErr != nil {
+			return nil, fmt.Errorf("failed to apply resource: %w; rollback also failed: %v", err, rollbackErr)
```
pPrecel: I would suggest inform user about this error but do not stop the process

internal/modules/install.go
```diff
@@ -0,0 +1,192 @@
...
+}
+
+func rollback(ctx context.Context, client kube.Client, resources []map[string]any) error {
+	if len(resources) == 0 {
+		return nil
+	}
+
+	for _, resource := range resources {
+		err := client.RootlessDynamic().Remove(ctx, &unstructured.Unstructured{Object: resource}, false)
+		if err != nil {
```
pPrecel, line 198: I would suggest continue on error

grego952:
I'd change `Are you sure you want to proceed with installing the community module?` to `Are you sure you want to proceed with the installation?`

grego952:
Instead of `Type your choice` I'd go with `Type the version number`

internal/cmdcommon/prompt/bool.go
```diff
@@ -0,0 +1,63 @@
...
+func (b *Bool) Prompt() (bool, error) {
+	var userInput string
+	fmt.Fprintf(b.writer, "%s %s: ", b.message, b.defaultValueDisplay())
+	_, err := fmt.Fscan(b.reader, &userInput)
+
+	// If the user just presses Enter, Fscan returns the EOF error
+	if err != nil && err == io.EOF {
+		// Treat as empty input, use default value
+		return b.defaultValue, nil
+	}
```
pPrecel: I think we should handle the case when the error is not equal to the `io.EOF` and then return it

internal/cmdcommon/prompt/oneofstringlist.go
```diff
@@ -0,0 +1,72 @@
...
+	}
+}
+
+func (l *OneOfStringList) Prompt() (string, error) {
+	var userInput string
+	fmt.Fprintf(l.writer, "%s\n%s\n\n%s", l.message, l.valuesListString(), l.promptText)
+	_, err := fmt.Fscan(l.reader, &userInput)
+	if err != nil && err == io.EOF {
+		return "", fmt.Errorf("no value was selected")
+	}
```
pPrecel: I think we should handle the case when the error is not equal to the `io.EOF` and then return it

internal/cmdcommon/prompt/bool_test.go
```diff
@@ -0,0 +1,92 @@
...
+			input:        "maybe\n",
+			defaultValue: false,
+			expectResult: false,
+			expectErr:    "invalid input, please enter 'y' or 'n'",
+		},
+	}
+
+	for _, tc := range tests {
+		t.Run(tc.name, func(t *testing.T) {
+			input := bytes.NewBufferString(tc.input)
```
pPrecel: We can test the case when the Read method returns an error. For example, to cover the case when the user runs the `add` command on the CI/CD system, where STDIN is not configured. To do so, I will suggest using the `ErrReader` function from [this package](https://go.dev/src/testing/iotest/reader.go).

internal/cmdcommon/prompt/oneofstringlist_test.go
```diff
@@ -0,0 +1,72 @@
...
+			values:     []string{"apple", "banana", "orange"},
+			want:       "",
+			wantErr:    "no value was selected",
+			wantOutput: "Select a fruit:\n - apple\n - banana\n - orange\n\nType the version number: ",
+		},
+	}
+
+	for _, tc := range tests {
+		t.Run(tc.name, func(t *testing.T) {
+			input := bytes.NewBufferString(tc.input)
```
pPrecel: We can test the case when the Read method returns an error. For example, to cover the case when the user runs the `add` command on the CI/CD system, where STDIN is not configured. To do so, I will suggest using the `ErrReader` function from [this package](https://go.dev/src/testing/iotest/reader.go).

PR #2534, opened 2025-07-23
Added Community Modules support for clusters not managed by KLM

musztardem:
**Description**

- Fixed `module catalog` and `module list` commands to support scenarios when the cluster is not managed by the Kyma Lifecycle Manager

**Related issue(s)**


internal/modules/list.go
```diff
@@ -155,6 +172,42 @@ func ListCatalog(ctx context.Context, client kube.Client) (ModulesList, error) {
...
+
+		if i := getModuleIndex(modulesList, moduleName, isCommunityModule(&moduleTemplate)); i != -1 {
+			modulesList[i].Versions = append(modulesList[i].Versions, version)
+		} else {
+			modulesList = append(modulesList, Module{
+				Name: moduleName,
+				Versions: []ModuleVersion{
+					version,
+				},
+				CommunityModule: isCommunityModule(&moduleTemplate), // always true
```
anoipm, line 198:
```suggestion
				CommunityModule: true, // always true
```

anoipm: we keep it for code clarity

PR #2570, opened 2025-09-03
Improve some of the error messages

kwiatekus:
**Description**

- Make error messages more informative
- Use `kyma environment` instead of `cluster`
- Add `SAP` before `Hana`
- hide technical details like `cis`

**Related issue(s)**
Related to feedback from one of the users

### DIFF
--- a/internal/modules/enable.go
+++ b/internal/modules/enable.go
@@ -23,7 +23,12 @@ func Enable(ctx context.Context, client kube.Client, repo repo.ModuleTemplatesRe
 
 func enable(writer io.Writer, ctx context.Context, client kube.Client, repo repo.ModuleTemplatesRepository, module, channel string, defaultCR bool, crs ...unstructured.Unstructured) clierror.Error {
 	if err := validateModuleAvailability(ctx, client, repo, module); err != nil {
-		return clierror.Wrap(err, clierror.New("invalid module name or version"))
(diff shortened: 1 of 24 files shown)


internal/modules/enable.go
```diff
@@ -23,7 +23,7 @@ func Enable(ctx context.Context, client kube.Client, repo repo.ModuleTemplatesRe
 
 func enable(writer io.Writer, ctx context.Context, client kube.Client, repo repo.ModuleTemplatesRepository, module, channel string, defaultCR bool, crs ...unstructured.Unstructured) clierror.Error {
 	if err := validateModuleAvailability(ctx, client, repo, module); err != nil {
-		return clierror.Wrap(err, clierror.New("module invalid"))
+		return clierror.Wrap(err, clierror.New("invalid module name or version"))
```
pPrecel: It's not possible to use the `--version` flag while enabling (not community) module. I would suggest removing information about the version and adding the module name tothe  error:

```suggestion
		return clierror.Wrap(err, clierror.New(fmt.Sprintf("failed to verify module: %s", module)))
```

kwiatekus: It is shown where we cannot find an available module by name.. I dont think `failed to verify module: qwerty` will be accurate

pPrecel (approved):

PR #2581, opened 2025-09-08
Update Kyma CLI tutorials

grego952:
**Description**

- ...
- ...
- ...

**Related issue(s)**

### DIFF
--- a/docs/user/tutorials/01-10-add-delete-modules.md
+++ b/docs/user/tutorials/01-10-add-delete-modules.md
@@ -9,7 +9,13 @@ This tutorial shows how you can add and delete a new module using Kyma CLI.
 
 ### Adding a New Module
 
-To add a new module with the default policy set to `CreateAndDelete`, use the following command:
+1. Check the list of modules that can be added:
+
+```bash
+kyma module catalog
+```
+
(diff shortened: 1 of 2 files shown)


docs/user/tutorials/01-10-add-delete-modules.md
```diff
@@ -9,7 +9,13 @@ This tutorial shows how you can add and delete a new module using Kyma CLI.
 
 ### Adding a New Module
 
-To add a new module with the default policy set to `CreateAndDelete`, use the following command:
+1. Check the list of modules that can be added:
```
IwonaLanger:
```suggestion
1. Check the list of modules that you can add:
```

IwonaLanger: Please indent the code blocks to make the points more visible. The options in point 2 could be bullet points. WDYT? If not, please indent them as well.

IwonaLanger: If you decide to use bullet points, you might also rephrase: `You can also specify which channel the module should use with the -c {CHANNEL-NAME} flag:` to `To specify which channel the module should use, add the -c {CHANNEL-NAME} flag:`

IwonaLanger (approved):

PR #2585, opened 2025-09-09
Implement the `call_files_to_save` action

pPrecel:
**Description**

- implement an action that allows calling a pod from the connected cluster, receive a file list, a nd write them on the machine

**Related issue(s)**

- https://github.com/kyma-project/cli/issues/2587


internal/extensions/actions/call_files_to_save.go
```diff
@@ -0,0 +1,141 @@
...
+
+func (a *callFilesToSaveAction) Run(cmd *cobra.Command, _ []string) clierror.Error {
+	clierr := a.Cfg.validate()
+	if clierr != nil {
+		return clierror.WrapE(clierr, clierror.New("invalid action configuration"))
+	}
+
+	if !filepath.IsLocal(a.Cfg.OutputDir) {
+		// output dir is not a local path, ask user for confirmation
+		clierr = getUserAcceptance(a.Cfg.OutputDir)
```
musztardem: `getUserAcceptance` is defined in the `function_init.go` - if user won't accept the prompt then the returned value will be `function init aborted` which might be misleading

internal/extensions/call/resonse.go
```diff
@@ -0,0 +1,43 @@
+package call
```
musztardem: there's a typo in the filename - `response.go`

Issue #2437, opened 2025-03-28
Enable users to export diagnostics data via Kyma CLI


kwiatekus:
**Description**

Provide a kyma CLI command that would output diagnostic data, such as:
 - Non-ready statuses of Kyma Module CRs (including conditions)
 - resource saturation of the nodes
 - k8s error/warning events ~with kyma-system namespace~
 - cluster metadata (i.e SAP BTP global account, subaccount, provider (azure, aws, etc..), api server url)

For example:

```
kyma alpha diagnose (--output {filepath}) (--fromat {yaml|json})
```

Module providers can further extend the command  by adding sub-commands for their modules:
I.e `kyma alpha diagnose keda` could output [keda specific metrics](https://keda.sh/docs/2.16/integrations/prometheus/#prometheus-exporter-metrics)  
I.e `kyma alpha keda docs` could open url in browser

**Reasons**

When users report defects we asked them to provide kyma installation details in order to better understand the problem. This feature will provide a standardised way of providing diagnostic data, which will help us better support the users, but also may expose some symptoms to the user himself.

kwiatekus:
As we discussed: 
 - please extend the warning events to all namespaces (the requirements adjusted)
 - please rename the `moduleCustomResourceStates` to `moduleErrors`

PR #2600, opened 2025-09-22
Create alpha diagnose command

musztardem:
**Description**

- Implemented command that gathers information about the cluster for the diagnostic purposes.
- Gathers the following information:
    - Resource saturation of the nodes
    - Warnings from the `kyma-system` namespace
    - Cluster metadata

**Related issue(s)**

#2437

### DIFF
--- a/internal/kubeconfig/kubeconfig.go
+++ b/internal/kubeconfig/kubeconfig.go
@@ -36,8 +36,7 @@ func PrepareWithToken(apiBase *api.Config, token string) *api.Config {
 }
 
 func Prepare(ctx context.Context, client kube.Client, name, namespace, time, output string, permanent bool) (*api.Config, clierror.Error) {
-	currentCtx := client.APIConfig().CurrentContext
-	clusterName := client.APIConfig().Contexts[currentCtx].Cluster
+	clusterName := getKubeconfigCurrentClusterName(client.APIConfig())
 	var tokenData authv1.TokenRequestStatus
 	var certData []byte
 	var err clierror.Error
@@ -85,13 +84,13 @@ func Prepare(ctx context.Context, client kube.Client, name, namespace, time, out
 			},
 		},
 		Contexts: map[string]*api.Context{
-			currentCtx: {
+			clusterName: {
 				Cluster:   clusterName,
 				Namespace: namespace,
 				AuthInfo:  name,
 			},
 		},
-		CurrentContext: currentCtx,
+		CurrentContext: clusterName,
 		Extensions:     nil,
 	}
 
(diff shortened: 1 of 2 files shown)


internal/diagnostics/diagnose.go
```diff
@@ -0,0 +1,26 @@
...
+	"context"
+	"os"
+
+	"github.com/kyma-project/cli.v3/internal/kube"
+)
+
+type DiagnosticData struct {
+	Metadata      Metadata
+	Warnings      KymaSystemWarnings
+	NodeResources []NodeResourceInfo
```
pPrecel: I would suggest renaming this field to `nodes`

internal/cmd/alpha/diagnose/diagnose.go
```diff
@@ -0,0 +1,109 @@
...
+	fmt.Println("Collecting diagnostics data...")
+
+	diagnosticData := diagnostics.GetData(cfg.Ctx, client, cfg.verbose)
+	err := render(&diagnosticData, cfg.outputPath, cfg.outputFormat)
+
+	if err != nil {
+		return clierror.Wrap(err, clierror.New("failed to get diagnostic data"))
+	}
+
+	fmt.Println("Done.")
```
pPrecel: I would suggest printing such information only when the `--output` flag is used to allow users to pipe program output to another app like yq or jq:

```suggestion
	diagnosticData := diagnostics.GetData(cfg.Ctx, client, cfg.verbose)
	err := render(&diagnosticData, cfg.outputPath, cfg.outputFormat)

	if err != nil {
		return clierror.Wrap(err, clierror.New("failed to get diagnostic data"))
	}
```

internal/cmd/alpha/diagnose/diagnose.go
```diff
@@ -0,0 +1,109 @@
...
+
+	return nil
+}
+
+func render(diagData *diagnostics.DiagnosticData, outputFilepath string, outputFormat output.Format) error {
+	var outputWriter io.Writer
+
+	if outputFilepath == "" {
+		outputWriter = os.Stdout
```
pPrecel, line 74: I suggest using the cobra command to get stdout. You can obtain it by using the `cmd. OutOrStdout()` func.

internal/diagnostics/noderesourceinfo.go
```diff
@@ -0,0 +1,251 @@
...
+
+	"github.com/kyma-project/cli.v3/internal/kube"
+	corev1 "k8s.io/api/core/v1"
+	"k8s.io/apimachinery/pkg/api/resource"
+	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
+	"k8s.io/apimachinery/pkg/fields"
+)
+
+type MachineInfo struct {
+	Name             string
```
pPrecel: I would suggest adding additional `yaml:"<name>"` tags for all fields to print yaml in `camelCase` format

pPrecel (commented):
I really like these changes. A few suggestions only

pPrecel (approved):

Issue #2606, opened 2025-09-24
Handle missing optional read priviliges in kyma-system for the app push command


kwiatekus:
**Description**

Looks like app push command takes for granted that the caller has read priovilige to kyma-system's gateways.
If its not the case, the command ends with error, while it should be possible to create api -rule anyway

**Steps to reproduce**

```
kubectl create ns my-app
namespace/my-app created

./bin/kyma alpha kubeconfig generate --serviceaccount test-sa --output namespaced-access.yaml --namespace my-app --clusterrole cluster-admin

Kubeconfig saved to: namespaced-access.yaml

kubectl auth can-i create apirules  -n my-app --kubeconfig namespaced-access.yaml 
yes

kubectl auth can-i get gateways  -n kyma-system --kubeconfig namespaced-access.yaml
no

kyma app push --image kennethreitz/httpbin:latest --container-port 80 --expose --istio-inject true  --name test --namespace my-app --kubeconfig namespaced-access.yaml
Extensions Warning:
failed to fetch all extensions from the target Kyma environment. Use the '--show-extensions-error' flag to see more details.


Creating deployment my-app/test

Creating service my-app/test

Creating API Rule my-app/test
Error:
  failed to domain address of the Kyma environment from gateway

Error Details:
  while getting Gateway %s in namespace %s: gateways.networking.istio.io "kyma-gateway" is forbidden: User "system:serviceaccount:my-app:test-sa" cannot get resource "gateways" in API group "networking.istio.io" in the namespace "kyma-system"

Hints:
  - Make sure Istio module is installed
  - kyma-gateway
  - kyma-system 
```

**Troubleshooting**

kwiatekus:
I noticed that the same read operation is enforced in UI.

@mrCherry97 do you plan to handle it in kyma dashboard?

<img width="796" height="511" alt="Image" src="https://github.com/user-attachments/assets/fac9b97c-cf93-4945-a59d-1a4f0f33e6a8" />

mrCherry97:
API Rules view is not our responsibility. We have introduced the `canI` function for extensibility, where users can provide a resource and check if it's available in each scope/RBAC/user. So I'll suggest reaching out to the goat team and asking them to check if it's working on the presets.

PR #2611, opened 2025-09-26
Check permission prior to reading gateway

kwiatekus:
**Description**

- handle missing permissions

**Related issue(s)**
#2606

### DIFF
--- a/internal/cmd/app/push.go
+++ b/internal/cmd/app/push.go
@@ -178,18 +178,28 @@ func runAppPush(cfg *appPushConfig) clierror.Error {
 	if cfg.expose {
 		fmt.Printf("\nCreating API Rule %s/%s\n", cfg.namespace, cfg.name)
 		var domain string
-		domain, clierr = client.Istio().GetClusterAddressFromGateway(cfg.Ctx)
-		if clierr != nil {
-			return clierror.WrapE(clierr, clierror.New("failed to domain address of the Kyma environment from gateway", "Make sure Istio module is installed"))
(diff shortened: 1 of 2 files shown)


internal/cmd/app/push.go
```diff
@@ -178,18 +178,28 @@ func runAppPush(cfg *appPushConfig) clierror.Error {
...
 
-		host := fmt.Sprintf("%s.%s", cfg.name, domain)
-		err = resources.CreateAPIRule(cfg.Ctx, client.RootlessDynamic(), cfg.name, cfg.namespace, host, uint32(*cfg.containerPort.Value))
+		err = resources.CreateAPIRule(cfg.Ctx, client.RootlessDynamic(), cfg.name, cfg.namespace, cfg.name, uint32(*cfg.containerPort.Value))
 		if err != nil {
 			return clierror.Wrap(err, clierror.New("failed to create API Rule resource", "Make sure API Gateway module is installed", "Make sure APIRule CRD is available in v2 version"))
 		}
 
-		fmt.Printf("\nThe %s app is available under the https://%s/ address\n", cfg.name, host)
+		if domain == "" {
```
pPrecel: I think we should change this to:

```suggestion
		if domain != "" {
```

Also, I think we should print the output even when the domain is empty, but with a nicely crafted default, like: `<CLUSTER_DOMAIN>` or something :)

kwiatekus: thx!. done

pPrecel (approved):

Issue #2236, opened 2024-11-05
Implement a MVP for `kyma app push` command


kwiatekus:
**Description**

Implement a new `kyma app push` command that deploys k8s resources that are necessary to transform docker image definition (dockerfile)  into an application.
 
For example  (just a proposal at this stage)
```bash
//will use Dockerfile from the bar folder and use custom port 8080
kyma app push --name foo --dockerfile bar/Dockerfile --container-port 8080

//will not build the image but simply use nginx as app's image
kyma app push --name mynginx  --image nginx:latest
```

**AC**
The command should:
- should be available in command palette as soon as the connection to cluster is established (kubeconfig context is set)
- locate the `dockerfile` (lookup in the current folder by default),
- fail if no dockerfile was found and no `--image` flag was used,
- build the image  if needed 
- push the image to the registry (the command should read the in-cluster's docker registry `config.json`) if needed
- fail if `--dockerfile` option is used but no docker registry module is available in the cluster
- apply deployment (making use of exposed port information read from dockerfile)
- apply k8s service (making use of the port information read from dockerfile)
- (if api-gateway module is enabled) apply api-rule to expose all paths for the default, or user provided port and using the app name as a hostname

The definition of the command should be part of docker-registry module. 
Presence of api-gateway module should extend the command's default capability by adding the api-rule on top.

**Execution**
 - [x] https://github.com/kyma-project/cli/issues/2238
 - [x] https://github.com/kyma-project/cli/issues/2239
 - [x] https://github.com/kyma-project/cli/issues/2240
 - [x] https://github.com/kyma-project/cli/issues/2241

**Reasons**
Kyma users need a simple way to deploy containerised application without having any kubernetes knowledge.

kwiatekus:
Stages:

 - create deployment based on `--name` and `--image` and `--containerPort` (3)
```
kyma alpha app push --name mynginx --image nginx:latest --containerPort xx
```

 - parse dockerfile to read port information and define k8s svc (2)

 - detection and support for api-gateway (3)
```
kyma alpha app push --name mynginx --image nginx:latest --containerPort xx --expose true
``` 

 - support custom image building (using lib, not docker daemon from host machine) (3)
```
kyma alpha app push --name myservice  --containerPort xx --expose true --dockerfile .
```

PR #2626, opened 2025-10-03
Leave essential output in --quiet mode for better automation experience

kwiatekus:
**Description**

- Allow easy extraction of  app URL in automation scenarios

**Related issue(s)**
#2236

### DIFF
--- a/internal/cmd/kyma.go
+++ b/internal/cmd/kyma.go
@@ -25,6 +25,7 @@ func NewKymaCMD() *cobra.Command {
 	extensions.AddCmdPersistentFlags(cmd)
 	cmdcommon.AddCmdPersistentKubeconfigFlag(cmd)
 	cmd.PersistentFlags().BoolP("help", "h", false, "Help for the command")
+	cmd.PersistentFlags().BoolP("quiet", "q", false, "Suppress non-essential output")
 
 	kymaConfig := cmdcommon.NewKymaConfig()
 
--- a/internal/extensions/extensions.go
+++ b/internal/extensions/extensions.go
@@ -6,12 +6,11 @@ import (
 	"io"
 	"os"
 	"slices"
-	"strconv"
-	"strings"
 
 	"github.com/kyma-project/cli.v3/internal/cmdcommon"
 	"github.com/kyma-project/cli.v3/internal/extensions/errors"
 	"github.com/kyma-project/cli.v3/internal/extensions/types"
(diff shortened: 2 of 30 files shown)


internal/cmd/kyma.go
```diff
@@ -25,6 +25,7 @@ func NewKymaCMD() *cobra.Command {
 	extensions.AddCmdPersistentFlags(cmd)
 	cmdcommon.AddCmdPersistentKubeconfigFlag(cmd)
 	cmd.PersistentFlags().BoolP("help", "h", false, "Help for the command")
+	cmd.PersistentFlags().BoolP("quiet", "q", false, "Suppress non-essential output")
```
pPrecel, line 28: Let's use the `BoolVarP` function to allow Cobra to assign a flag value to a specific variable. Such a variable can be stored inside the `kymaConfig` struct defined a few lines below and then read from this config anywhere.

internal/extensions/extensions.go
```diff
@@ -167,27 +166,6 @@ func parseRequiredField[T any](cmData map[string]string, cmKey string) (*T, erro
...
-
-		// example: --show-extensions-error or --show-extensions-error=true
-		if strings.HasPrefix(arg, flag) && !strings.Contains(arg, "false") {
-			return true
-		}
-	}
-
-	return false
-}
-
```
pPrecel, line 190: This method allows us to get the flag value when commands are building, not running. When commands run, we should  use flags defined in the commands definition (like [here](https://github.com/kyma-project/cli/blob/main/internal/cmd/app/push.go#L65-L66))

internal/flags/get.go
```diff
@@ -0,0 +1,28 @@
+package flags
+
+import (
+	"os"
+	"strconv"
+	"strings"
+)
+
+// GetBoolFlagValue checks if a boolean flag is set in the os.Args and returns its value.
+func GetBoolFlagValue(flag string) bool {
```
pPrecel, line 10: I think this method should not be public due to its representing crucial logic, and it's kinda hacky :)

pPrecel (changes requested):
For me, this PR provides inconsistent changes. We want to print only the external address when `-q` flag is used. Why, in this case, did we decide to name this flag `--quiet`? not `--print-external-address-only` or something similar? And why is this flag persistent? I think with such an approach, the `-q` flag may mute more or less (depending on the author's will) on execution, and I think it is super non-consistent

Issue #2627, opened 2025-10-06
Allow establishing GitHub OIDC trust and RBAC using single Kyma CLI command


kwiatekus:
**Description**
Introduce a new command to configure trust between a Kyma cluster and a GitHub repository by:
- Creating an `authentication.gardener.cloud/v1alpha1` `OpenIDConnect` resource.
- Creating a `RoleBinding` or `ClusterRoleBinding` for the GitHub OIDC subject derived from the repository claim.

**Ideas (to refine)**:
`kyma alpha authorize repository`

(We could follow up later on with `kyma alpha authorize user`, `kyma alpha authorize serviceaccount` and extract/re-use part of the logic that we have already for `kyma alpha kubeconfig generate`)

Flags
- --repository string (required): GitHub repo in owner/name format (e.g., kyma-project/cli).
- --client-id string (required): OIDC client ID (audience) expected in the token.
- --issuer-url string (default: https://token.actions.githubusercontent.com): OIDC issuer.
- --prefix string: Username prefix for the repository claim (e.g., gh-oidc:). (Default to OIDC's name followed by `/`: `{metadata.name}/`)
- --namespace string: Namespace for RoleBinding (required if not cluster-wide and binding a Role or namespaced ClusterRole).
- --cluster-wide bool (default: false): If true, create a ClusterRoleBinding; otherwise, a RoleBinding.
- --role string: Role name to bind (namespaced).
- --clusterrole string: ClusterRole name to bind (usable for RoleBinding or ClusterRoleBinding).
- --name string: Name for the OpenIDConnect resource (optional; default derives from client-id; no suffix).
- --dry-run bool (default: false): Print resources without applying.
- -o yaml

Example
kyma alpha authorize repository \
  --repository kwiatekus/my-cap-application \
  --clientId gh-deploy-my-app \
  --prefix gh-oidc-my-app: \
  --namespace test \
  --role can-push-apps

1. Create or edit an existing OpenIDConnect CR:
```yaml  
  apiVersion: authentication.gardener.cloud/v1alpha1
  kind: OpenIDConnect
  metadata:
    name: <name>
  spec:
    issuerURL: <issuerURL>
    clientID: <clientId>
    usernameClaim: repository
    usernamePrefix: <prefix> 
    requiredClaims:
        repository: <repository>
```

2. Compute the Kubernetes user identity as:
  - subject.kind: User
  - subject.name: <prefix><repository> 

3. Bind RBAC:
[...]

PR #2655, opened 2025-10-17
Added kyma alpha authorize command

musztardem:
**Description**

* Added a new "authorize" alpha command that allow users to configure trust between a Kyma cluster and a GitHub repository by creating an OpenIDConnect resource plus a RoleBinding or ClusterRoleBinding.
* Introduced builders for both resources:
    * OIDCBuilder constructs OpenIDConnect resources
    * RBACBuilder constructs RoleBinding or ClusterRoleBinding resources
* Implemented a ResourceApplier that applies the OIDC and RBAC resources to the cluster.

**Related issue(s)**

#2627

### DIFF
--- a/internal/authorization/rbacbuilder.go
+++ b/internal/authorization/rbacbuilder.go
@@ -71,8 +71,7 @@ func (b *RBACBuilder) validateForClusterRoleBinding() error {
 		return fmt.Errorf("repository is required")
 	}
 
-	repoNameParts := strings.Split(b.repository, "/")
-	if len(repoNameParts) != 2 {
+	if !b.isRepoNameValid() {
 		return fmt.Errorf("repository must be in owner/name format (e.g., kyma-project/cli)")
 	}
 
@@ -88,8 +87,7 @@ func (b *RBACBuilder) validateForRoleBinding() error {
 		return fmt.Errorf("repository is required")
 	}
 
-	repoNameParts := strings.Split(b.repository, "/")
-	if len(repoNameParts) != 2 {
+	if !b.isRepoNameValid() {
 		return fmt.Errorf("repository must be in owner/name format (e.g., kyma-project/cli)")
 	}
 
@@ -106,12 +104,13 @@ func (b *RBACBuilder) validateForRoleBinding() error {
 	return nil
 }
 
+func (b *RBACBuilder) isRepoNameValid() bool {
+	repoNameParts := strings.Split(b.repository, "/")
+	return len(repoNameParts) == 2
+}
+
 func (b *RBACBuilder) getSubjectName() string {
-	subjectName := b.repository
-	if b.prefix != "" {
-		subjectName = b.prefix + b.repository
-	}
-	return subjectName
+	return b.prefix + b.repository
 }
 
 func (b *RBACBuilder) getClusterRoleBindingName() string {


internal/authorization/rbacbuilder.go
```diff
@@ -0,0 +1,187 @@
...
+
+	return nil
+}
+
+func (b *RBACBuilder) getSubjectName() string {
+	subjectName := b.repository
+	if b.prefix != "" {
+		subjectName = b.prefix + b.repository
+	}
+	return subjectName
```
anoipm:
```suggestion
	return subjectName b.prefix + b.repository
```

internal/authorization/rbacbuilder.go
```diff
@@ -0,0 +1,187 @@
...
+
+func (b *RBACBuilder) validateForRoleBinding() error {
+	if b.repository == "" {
+		return fmt.Errorf("repository is required")
+	}
+
+	repoNameParts := strings.Split(b.repository, "/")
+	if len(repoNameParts) != 2 {
+		return fmt.Errorf("repository must be in owner/name format (e.g., kyma-project/cli)")
+	}
```
anoipm, line 92: Could we use the same part of code for RB and CRB?

internal/cmd/alpha/authorize/authorize.go
```diff
@@ -0,0 +1,213 @@
...
+	cmd.Flags().StringVar(&cfg.issuerURL, "issuerURL", "https://token.actions.githubusercontent.com", "OIDC issuer")
+	cmd.Flags().StringVar(&cfg.prefix, "prefix", "", "Username prefix for the repository claim (e.g., gh-oidc:)")
+	cmd.Flags().StringVar(&cfg.namespace, "namespace", "", "Namespace for RoleBinding (required if not cluster-wide and binding a Role or namespaced ClusterRole)")
+	cmd.Flags().BoolVar(&cfg.clusterWide, "cluster-wide", false, "If true, create a ClusterRoleBinding; otherwise, a RoleBinding")
+	cmd.Flags().StringVar(&cfg.role, "role", "", "Role name to bind (namespaced)")
+	cmd.Flags().StringVar(&cfg.clusterrole, "clusterrole", "", "ClusterRole name to bind (usable for RoleBinding or ClusterRoleBinding)")
+	cmd.Flags().StringVar(&cfg.name, "name", "", "Name for the OpenIDConnect resource (optional; default derives from clientId)")
+	cmd.Flags().BoolVar(&cfg.dryRun, "dry-run", false, "Print resources without applying")
+	cmd.Flags().VarP(&cfg.outputFormat, "output", "o", "Output format (yaml or json)")
+
```
anoipm, line 70: Do all combinations of parameters are possible? Shouldn't we restrict or validate some of them?

musztardem: This is the list of all parameters that the command can consume. In the `PreRun` function in the command definition I defined additional rules for the parameters:
- `repository` and `clientId` are required
- either `role` or `clusterRole` must be provided
- when `role` is present then `namespace` is also required

Some other requirements are then checked in the runtime as they require validation of specific values, e.g. when clusterRole is provided + clusterWide=false -> namespace is required

PR #2674, opened 2025-10-24
Use the `out` package in whole cli

pPrecel:
**Description**

- as in the title

**Related issue(s)**

- previous PR - https://github.com/kyma-project/cli/pull/2672

### DIFF
--- a/internal/modules/disable.go
+++ b/internal/modules/disable.go
@@ -47,7 +47,7 @@ func GetRunningResourcesOfCommunityModule(ctx context.Context, repo repo.ModuleT
 }
 
 func disableCommunity(printer *out.Printer, ctx context.Context, repo repo.ModuleTemplatesRepository, module string) clierror.Error {
-	printer.Msgfln("removing %s community module from the target Kyma environment\n", module)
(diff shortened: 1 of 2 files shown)


internal/cmd/alpha/diagnose/diagnose.go
```diff
@@ -43,72 +42,75 @@ func NewDiagnoseCMD(kymaConfig *cmdcommon.KymaConfig) *cobra.Command {
 	return cmd
 }
 
-func diagnose(cfg *diagnoseConfig, output io.Writer) clierror.Error {
+func diagnose(cfg *diagnoseConfig) clierror.Error {
+	if cfg.verbose {
+		out.EnableDebug()
```
anoipm, line 47: `diagnose` changes `out`. I doubt if it's proper behavior

pPrecel: It's proper behavior.  The `out` package is designed to be the central configuration for the Standard outputs across the whole CLI. Once configured, it should affect all places in the whole CLI. And this is proper behavior because when a user decides to use the `--verbose` flag, then all places in code should print out verbose messages whenever the code is located.

Issue #2661, opened 2025-10-21
Make --version optional for module template pull (auto-select latest)


kwiatekus:
**Description**
Currently pulling a remote module template requires specifying --version;  Improve UX by allowing omission of the version flag and automatically selecting the newest available version for the given module name.

**Acceptance Criteria**

 - [ ] Running `kyma module pull --name foo` (without --version) applies the highest available version.
 - [ ] Explicit `kyma module pull --name foo --version 1.2.3` behaves as before.
 - [ ] Help description reflects version as optional

**Testing scenario**
 - Unit test covers selection of latest among multiple versions (including pre-release case: release candidates for new releases should not be considered as latest).

PR #2677, opened 2025-10-26
Return latest community module when flag --version is not specified

MichalKalke:
**Description**

- Return latest community module when flag --version is not specified

**Related issue(s)**

- #2661

### DIFF
--- a/internal/cmd/module/pull.go
+++ b/internal/cmd/module/pull.go
@@ -37,6 +37,9 @@ Examples:
   # Pull a specific community module
   kyma module pull community-module-name
 
+  # Pull the latest version of a module into specific namespace
+  kyma module pull community-module-name --namespace module-namespace
+
   # Pull a module with a specific version into specific namespace
   kyma module pull community-module-name --version v1.0.0 --namespace module-namespace`,
 		Args: cobra.ExactArgs(1),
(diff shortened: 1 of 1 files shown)


internal/modules/repo/moduletemplates.go
```diff
@@ -284,17 +284,56 @@ func (r *moduleTemplatesRepo) ExternalCommunityByNameAndVersion(ctx context.Cont
...
+func isNewerVersion(version1, version2 string) bool {
+	// Skip pre-release versions (containing -rc, -alpha, -beta)
+	if strings.Contains(version1, "-") {
+		return false
+	}
+	if strings.Contains(version2, "-") && !strings.Contains(version1, "-") {
+		return true
+	}
+
+	return version1 > version2
```
kwiatekus, line 334: would this always work in edge cases?
why not using `golang.org/x/mod/semver`

internal/modules/repo/moduletemplates_test.go
```diff
@@ -549,6 +549,53 @@ func TestModuleTemplatesRepo_DeleteResourceReturnWatcher(t *testing.T) {
...
+func TestFindLatestVersion(t *testing.T) {
+	t.Run("returns empty module when no modules provided", func(t *testing.T) {
+		result := findLatestVersion([]kyma.ModuleTemplate{})
+
+		require.Equal(t, kyma.ModuleTemplate{}, result)
+	})
+
+	t.Run("returns single module when only one provided", func(t *testing.T) {
+		module := kyma.ModuleTemplate{
+			Spec: kyma.ModuleTemplateSpec{Version: "v1.0.0"},
```
kwiatekus, line 561: Please test w/o `v` prefix. As the code will be working with version without it
https://github.com/kyma-project/community-modules/blob/main/modules/registry-proxy/registry-proxy-0.14.0.yaml#L11

PR #2681, opened 2025-10-27
Use semver lib to find latest release version

MichalKalke:
**Description**

- Use semver lib to find latest release version
- Remove `v` from test cases 

**Related issue(s)**

- #2661


internal/modules/repo/moduletemplates.go
```diff
@@ -7,6 +7,7 @@ import (
 	"net/http"
 	"strings"
 
+	"github.com/Masterminds/semver/v3"
```
halamix2, line 10: run `go mod tidy`, this import is currently indirect

MichalKalke: Thanks, done

PR #2740, opened 2025-11-20
Modules package refactor - `module catalog` PoC

musztardem:
**Description**

- Proposed a new architecture for the `modules`  package.
- This PR is a starting point for the major refactor of the mentioned package.

**Related issue(s)**

#2786

### DIFF
--- a/internal/modulesv2/catalog_test.go
+++ b/internal/modulesv2/catalog_test.go
@@ -8,7 +8,6 @@ import (
 	"github.com/kyma-project/cli.v3/internal/modulesv2/dtos"
 	"github.com/kyma-project/cli.v3/internal/modulesv2/entities"
 	modulesfake "github.com/kyma-project/cli.v3/internal/modulesv2/fake"
-	"github.com/kyma-project/cli.v3/internal/modulesv2/testfactories"
 	"github.com/stretchr/testify/assert"
 	"github.com/stretchr/testify/require"
 )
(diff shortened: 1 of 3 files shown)


internal/di/container.go
```diff
@@ -0,0 +1,76 @@
...
+
+// DIContainer is a simple dependency injection container for singleton instances
+type DIContainer struct {
+	instances map[reflect.Type]interface{}
+	factories map[reflect.Type]Factory
+}
+
+type Factory func(container *DIContainer) (interface{}, error)
+
+func NewDIContainer() *DIContainer {
```
pPrecel: I would suggest renaming this function and the returned structure to something without the `DI` word because it's redundant from the package perspective. If I understand it correctly, this package specializes in dependency injection, and everything inside is about this domain. I think the `NewContainer` for func and `Container` for struct would be enough. From outside of the package, usage will look like this:

```go
di.NewContainer() // func
di.Container // struct
```

internal/modulesv2/testfactories/communitymoduletemplate.go
```diff
@@ -0,0 +1,49 @@
+package testfactories
```
pPrecel: I would suggest moving this package and `fake` to separate one called `testing`, or just `fake`, or something like this, to keep all testing stuff in one place

grego952 (approved):

Issue #2651, opened 2025-10-15
Add  capability for `kyma alpha diagnose` to collect error logs of Kyma module workloads


kwiatekus:
**Description** 
 
Add a subcommand that aggregates recent error logs from Pods in `kyma-system` belonging to given or all enabled Kyma modules. 

**Reason**
 - Allowing users to dump error logs on the spot and attaching them into support ticket
 - Possibility for fast triage of failing module components without manual pod selection.

**Ideas (to refine)**  
```
kyma alpha diagnose logs [flags]
```

Flags  
- `--module <name>` (repeatable): restrict to specific module(s).  
- `--since <duration>` (default: 10m): log time range.  
- `--lines <n>` (default: 200): max lines per container (fallback if `--since` unsupported).  
- `--timeout`

Include only json parsable log entries where level >= warning (`warning` and `error`)

Logs should be printed from the containers of those pods which are managed by deployments labeled with `kyma-project.io/module` label

Edge Cases
- Called w/o --module flag and  Kyma CR missing → warn and exit 0.  
- No matching workloads → warn "No module workloads found."  
- Large clusters: honor interrupt signal (cancellation) / timeout.  
- If both `--since` and `--lines` given, prefer `--since`.

**Acceptance Criteria**  
- [ ] Collects logs only from modules present in Kyma CR unless filtered by --module flag.  
- [ ] JSON output parsable.    
- [ ] skip gracefully if the logs are not json parsable

PR #2773, opened 2025-12-05
Added strict mode to diagnose logs command

musztardem:
**Description**

- Added default (more permissive) and strict (less permissive) logs selection algorithms for `alpha diagnose logs` command.
- Added simpler warning message in a situation when timeout occurs during logs collection.
- Implemented new logs format that stores also `errors` collected for each module

**Related issue(s)**

#2651

### DIFF
--- a/internal/modules/repo/moduletemplates.go
+++ b/internal/modules/repo/moduletemplates.go
@@ -158,23 +158,29 @@ func (r *moduleTemplatesRepo) RunningAssociatedResourcesOfModule(ctx context.Con
 func (r *moduleTemplatesRepo) Resources(ctx context.Context, moduleTemplate kyma.ModuleTemplate) ([]map[string]any, error) {
 	var parsedResources []map[string]any
 
-	for _, resource := range moduleTemplate.Spec.Resources {
-		resourceYamls, err := getFileFromURL(resource.Link)
-		resourceYamlsArr := strings.Split(string(resourceYamls), "---")
-		if err != nil {
-			return nil, fmt.Errorf("failed to fetch resource YAMLs from %s: %w", resource.Link, err)
+	resources := kyma.Resource{}
+	for _, res := range moduleTemplate.Spec.Resources {
+		if res.Name == "rawManifest" {
+			resources = res
+			break
 		}
+	}
 
-		for _, yamlStr := range resourceYamlsArr {
-			if strings.TrimSpace(yamlStr) == "" {
-				continue
-			}
-			var res map[string]any
-			if err := yaml.Unmarshal([]byte(yamlStr), &res); err != nil {
-				return nil, fmt.Errorf("failed to parse module resource YAML for %s:%s - %w", moduleTemplate.Spec.ModuleName, moduleTemplate.Spec.Version, err)
-			}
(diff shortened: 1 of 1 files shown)


internal/diagnostics/logs.go
```diff
@@ -112,70 +129,124 @@ func (c *ModuleLogsCollector) getPodsForModule(ctx context.Context, moduleName s
...
 	if err := scanner.Err(); err != nil {
-		out.Errfln("Error reading logs from pod %s container %s: %v", podName, containerName, err)
+		return collected, fmt.Errorf("error reading logs %v", err)
 	}
 
-	return collected
+	return collected, nil
+}
+
+func (c *ModuleLogsCollector) getParseStrategy() func(string) bool {
```
anoipm: IMO, name of this method is too generic. It returns something like `strictMode`/`getParseStrict`/... but not strategy.

internal/diagnostics/logs.go
```diff
@@ -112,70 +129,124 @@ func (c *ModuleLogsCollector) getPodsForModule(ctx context.Context, moduleName s
...
+			return true
+		}
+	}
+
+	return false
+}
+
+func containsFalsePositives(line string) bool {
+	falsePositives := []string{
+		"\"error\": null", "\"error\":null",
```
anoipm: We could have many falsePositives. For more readable code:
```suggestion
		"\"error\": null", 
		"\"error\":null",
```

internal/diagnostics/logs.go
musztardem: Hmm, this method returns another method that stores logic responsible for logs filtering so simple `strictMode` or `getParseStrict` also doesn't feel good to me because it would look like it's returning some simple value. What do you think about `getLogsFilteringLogic` or something?

internal/diagnostics/logs.go
```diff
@@ -112,70 +129,147 @@ func (c *ModuleLogsCollector) getPodsForModule(ctx context.Context, moduleName s
...
+	var obj map[string]any
+
+	if err := json.Unmarshal([]byte(line), &obj); err != nil {
+		// Not JSON -> ignore
+		return false
+	}
+	lvlRaw, hasLevel := obj["level"]
+	_, hasMessage := obj["message"]
+	_, hasTimestamp := obj["timestamp"]
+	if !hasLevel || (!hasMessage && !hasTimestamp) {
```
anoipm: I don't understand this condition.

internal/diagnostics/logs.go
```diff
@@ -112,70 +129,147 @@ func (c *ModuleLogsCollector) getPodsForModule(ctx context.Context, moduleName s
...
+
+	if onlyContainsFalsePositives(lineLower) {
+		return false
+	}
+
+	for _, keyword := range errorKeywords {
+		if strings.Contains(lineLower, keyword) {
+			return true
+		}
+	}
```
anoipm: This functionality is doubled inside `onlyContainsFalsePositives()`

internal/diagnostics/logs.go
musztardem: Updated the condition to be more strict

Issue #2554, opened 2025-08-21
Add ability to run Kyma Dashboard locally


mrCherry97:
**Description**

Add the ability to run the Kyma Dashboard locally.  
In the previous CLI, there was a `kyma dashboard` command that was responsible for running the Kyma Dashboard Docker image locally and automatically passing the kubeconfig to the Kyma Dashboard (as far as I remember, it was pushed on the `public/kubeconfig/kubeconfig.yaml` path, and the user was logged in using the kubeconfigID path).  

Reintroducing this capability would allow users to quickly access the dashboard without manually setting up and configuring Docker containers.

**Reasons**

- Improves developer experience by restoring a useful CLI command that was available in the past.  
- Simplifies the process of running Kyma Dashboard locally, especially for new contributors and users.  
- Reduces manual steps of configuring Docker and kubeconfig, minimizing chances of user error.  
- Provides an easy way to test and interact with Kyma clusters in a local development environment.  

**Attachments**

- [Busola documentation: Run Kyma Dashboard via Docker](https://github.com/kyma-project/busola/blob/main/docs/contributor/busola-docker.md)

mrCherry97:
If adjustments are needed in Busola, please ping me. Also, if you have other suggestions to fix this problem, please contact me or @pbochynski. We are open to discussion.

kwiatekus:
@mrCherry97 would it run on arm architecture?

when I tried the command, I got an erroro that there is no image for arm

```
docker run --rm -it -p 3001:3001 -v "${KUBECONFIG}":/app/core-ui/kubeconfig/$(basename "${KUBECONFIG}") --pid=host --name busola europe-docker.pkg.dev/kyma-project/prod/busola:latest

Unable to find image 'europe-docker.pkg.dev/kyma-project/prod/busola:latest' locally
latest: Pulling from kyma-project/prod/busola
docker: no matching manifest for linux/arm64/v8 in the manifest list entries
```

If we expose it in cli, we will expose the problem.
Please build a mutli-arch image

mrCherry97:
@kwiatekus PR is merged, I see on post that everything is working, so images are multiarch. Please clear my assaginee.

kwiatekus:
`kyma dashboard` - inteactive, opens browser
`kyma dashboard start` starts in background, opens browser (if container exists, doesnt create new)
 `kyma dashboard stop` stops  the background container

pPrecel:
Here is the code from the previous implementation:
[...]

PR #2785, opened 2025-12-09
Add container functionality, add initial dashboard start functionality

FilipRudy:
**Description**

- As in title

**Related issue(s)**

[#2554](https://github.com/kyma-project/cli/issues/2554)

### DIFF
--- a/internal/busola/container.go
+++ b/internal/busola/container.go
@@ -70,10 +70,7 @@ func (c *Container) Watch(ctx context.Context) error {
 
 // Stop stops the dashboard container.
 func (c *Container) Stop(ctx context.Context) error {
-	if err := c.docker.ContainerStop(ctx, c.id, container.StopOptions{}); err != nil {
-		return err
-	}
-	return nil
+	return c.docker.ContainerStop(ctx, c.id, container.StopOptions{})
 }
 
 func (c *Container) containerOpts(envs []string) docker.ContainerRunOpts {


internal/cmd/dashboard/start.go
```diff
@@ -1,17 +1,20 @@
...
 type dashboardStartConfig struct {
 	*cmdcommon.KymaConfig
-	port          string
-	containerName string
-	//kubeconfigPath string <- to be used in a future PR
-	//verbose        bool	<- to be used in a future PR
+	port           string
+	containerName  string
+	kubeconfigPath string
+	verbose        bool
```
pPrecel: I think these fields are not bound to any flag or input, so they can't be set. Additionally, I would suggest using the kubeconfig from the `--kubeconfig` persistent flag that is included in every command:

```console
$ kyma -h

...

Flags:
...
      --kubeconfig string       Path to the Kyma kubeconfig file
...
```

internal/docker/container.go
```diff
@@ -0,0 +1,102 @@
...
+		port:           port,
+		kubeconfigPath: kubeconfigPath,
+		verbose:        verbose,
+	}
+}
+
+// Start runs the dashboard container.
+func (c *Container) Start() error {
+	var err error
+	c.docker, err = NewClient()
```
pPrecel: I would suggest moving the `NewClient()` function to the `New(name, port, kubeconfigPath string, verbose bool) *Container`. Right now, the `Start()` function must be run first to even the initial Docker client. So, for example, when I would use the Container structure to stop an existing container or watch it, I need to run the Start function to initialise the Docker client first. 

The `Container` structure should be fully configured before running its any functions.

internal/docker/stop.go
```diff
@@ -0,0 +1,18 @@
...
+
+import (
+	"context"
+	"fmt"
+
+	"github.com/docker/docker/api/types/container"
+)
+
+func (c *Client) Stop(ctx context.Context, containerID string, log func(...interface{})) func() {
+	return func() {
```
pPrecel: I would suggest not returning a function and instead doing business logic. This function is doing two things:

1. Stop the container
2. Log error if it occurs

I think Stop func should return raw error, and if you need to handle it in a specific way, then do it in a specialized func closer to the expected logic. This is a part of the Docker client and should be as generic as possible.

internal/docker/containerfollowrun.go
```diff
@@ -0,0 +1,32 @@
...
+	})
+	if err != nil {
+		return err
+	}
+	defer buf.Close()
+
+	dstout, dsterr := io.Discard, io.Discard
+	if forwardOutput {
+		dstout = out.New()
+		dsterr = out.New()
```
pPrecel: Use the default output writer, please:

```suggestion
		dstout = out.Default.MsgWriter()
		dsterr = out.Default.ErrWriter()
```

internal/docker/container.go
```diff
@@ -0,0 +1,102 @@
...
+func (c *Container) Stop(ctx context.Context, log func(...interface{})) error {
+	log(fmt.Sprintf("\r- Removing container %s...\n", c.id))
+
+	if err := c.docker.Stop(ctx, c.id); err != nil {
+		log(err)
+		return err
+	}
+
+	return nil
+}
```
pPrecel: I don't think the Stop function is consistent with other functions in this package. It receives log func as an argument and logs an error before returning it. I would suggest using the `out.Msg` func instead and do not log error before returning:

```suggestion
func (c *Container) Stop(ctx context.Context) error {
	out.Msg(fmt.Sprintf("Removing container %s...\n", c.id))

	if err := c.docker.Stop(ctx, c.id); err != nil {
		return err
	}

	return nil
}
```

internal/busola/container.go
```diff
@@ -0,0 +1,102 @@
...
+
+	var err error
+	if c.id, err = c.docker.PullImageAndStartContainer(context.Background(), opts); err != nil {
+		return fmt.Errorf("unable to start container: %w", err)
+	}
+	return nil
+}
+
+// Opens the kyma dashboard in a browser.
+func (c *Container) Open(path string) error {
```
pPrecel, line 56: I will stop for a while and rethink the concept of this package. If I understand it correctly, it should be a high-level wrapper for the pure Docker client, but right now it's specifically designed to operate the Busola container. I would suggest changing the package name to `busola`.

internal/busola/container.go
```diff
@@ -0,0 +1,90 @@
...
+func (c *Container) Watch(ctx context.Context) error {
+	return c.docker.ContainerFollowRun(ctx, c.id, c.verbose)
+}
+
+// Stop stops the dashboard container.
+func (c *Container) Stop(ctx context.Context) error {
+	if err := c.docker.ContainerStop(ctx, c.id, container.StopOptions{}); err != nil {
+		return err
+	}
+	return nil
```
pPrecel: It can be simplified:

```suggestion
	return c.docker.ContainerStop(ctx, c.id, container.StopOptions{})
```

mmitoraj (approved):

PR #2812, opened 2026-01-11
Add docker package tests

FilipRudy:
**Description**

- As in title

**Related issue(s)**

#2554

### DIFF
--- a/internal/docker/pull.go
+++ b/internal/docker/pull.go
@@ -48,7 +48,7 @@ func (c *Client) PullImageAndStartContainer(ctx context.Context, opts ContainerR
 	})
 }
 
-// pullImageAndStartContainer is the private function that can be tested with mocks
+// PullImageAndStartContainer is the private function that can be tested with mocks
 func pullImageAndStartContainer(ctx context.Context, opts ContainerRunOpts, u utils) (string, error) {
 	config := &container.Config{
 		Env:          opts.Envs,


internal/docker/client_test.go
```diff
@@ -0,0 +1,17 @@
+package docker
+
+import (
+	"testing"
+
+	"github.com/stretchr/testify/require"
+)
+
+func TestNewClient(t *testing.T) {
```
pPrecel: It's ok that you added these tests, but they are not very helpful. We could omit them.

internal/docker/stopcontainer.go
```diff
@@ -46,3 +47,15 @@ func (c *Client) stopContainerOnSigInt(
 	case <-done:
 	}
 }
+
+func (c *Client) StopContainerOnSigInt(sigCh chan os.Signal) error {
```
pPrecel: This function is tested but not used anywhere. Is it expected behaviour?

I see that in this package are two functions with the same name (`StopContainerOnSigInt` and `stopContainerOnSigInt`). If you are looking for a testing strategy, where private func is tested, and a public one is used in code with the default configuration, then please look at [this example](https://github.com/kyma-project/cli/blob/main/internal/registry/image.go#L35-L55)

internal/docker/containerfollowrun_test.go
```diff
@@ -0,0 +1,58 @@
...
+func TestContainerFollowRun(t *testing.T) {
+	t.Run("function writes to provided writers", func(t *testing.T) {
+		outBuf, errBuf := &bytes.Buffer{}, &bytes.Buffer{}
+		cli := &mockClient{
+			outputMessage: "test-output",
+			dstOut:        outBuf,
+			dstErr:        errBuf,
+		}
+
+		err := cli.ContainerFollowRun(true)
```
pPrecel: Are you testing a mock client here? 🤔

internal/docker/pull_test.go
```diff
@@ -0,0 +1,86 @@
...
+)
+
+func TestPullImageAndStartContainer(t *testing.T) {
+	t.Run("successful pulling and starting the container", func(t *testing.T) {
+		cli := &testClient{
+			pullBody: validPullResponse(),
+			createID: "test-container",
+		}
+
+		id, err := cli.PullImageAndStartContainer()
```
pPrecel: Here you are testing the mock client as well 🤫

pPrecel (changes requested):

pPrecel (approved):

PR #2814, opened 2026-01-13
Fixed channels display for module catalog command

musztardem:
**Description**

- `kyma module catalog` command was displaying incomplete information about channels for a specific version, e.g. for the `serverless` module user was getting `1.9.5(experimental)` pair, while the version was also present on the `regular` and `fast` channels.

**Related issue(s)**

https://github.com/orgs/kyma-project/projects/26/views/1?pane=issue&itemId=148229234

### DIFF
--- a/internal/modules/enable.go
+++ b/internal/modules/enable.go
@@ -3,6 +3,7 @@ package modules
 import (
 	"context"
 	"fmt"
+	"slices"
 	"time"
 
 	"github.com/kyma-project/cli.v3/internal/clierror"
@@ -98,10 +99,8 @@ func validateModuleAvailability(ctx context.Context, client kube.Client, repo re
 	}
 
 	for _, v := range availableCoreVersions {
-		for _, ch := range v.Channels {
-			if ch == channel || channel == "" {
-				return nil
-			}
+		if slices.Contains(v.Channels, channel) {
+			return nil
 		}
 	}
 
--- a/internal/modules/manage.go
+++ b/internal/modules/manage.go
@@ -135,10 +135,8 @@ func GetAvailableChannelsAndVersions(ctx context.Context, client kube.Client, re
 
 		assignedChannels := getAssignedChannels(*moduleReleaseMetas, moduleName, coreModuleTemplate.Spec.Version)
 
-		if len(assignedChannels) > 0 {
-			for _, channel := range assignedChannels {
-				channelsAndVersions[channel] = coreModuleTemplate.Spec.Version
-			}
+		for _, channel := range assignedChannels {
+			channelsAndVersions[channel] = coreModuleTemplate.Spec.Version
 		}
 	}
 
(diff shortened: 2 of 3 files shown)


internal/modules/enable.go
```diff
@@ -98,8 +98,10 @@ func validateModuleAvailability(ctx context.Context, client kube.Client, repo re
 	}
```
anoipm, line 99:
```suggestion
    channel = kyma.Spec.Channel
    if len(availableCoreVersions) > 0 {
        return nil
    }
}
```

internal/modules/enable.go
```diff
@@ -98,8 +98,10 @@ func validateModuleAvailability(ctx context.Context, client kube.Client, repo re
...
 
 	for _, v := range availableCoreVersions {
-		if v.Channel == channel || channel == "" {
-			return nil
+		for _, ch := range v.Channels {
+			if ch == channel || channel == "" {
```
anoipm: This condition is not crear and I suggest to extract checking if there is any available channel.

```suggestion
			if ch == channel {
```

internal/modules/manage.go
```diff
@@ -132,10 +133,12 @@ func GetAvailableChannelsAndVersions(ctx context.Context, client kube.Client, re
...
-		assignedChannel := getAssignedChannel(*moduleReleaseMetas, moduleName, coreModuleTemplate.Spec.Version)
+		assignedChannels := getAssignedChannels(*moduleReleaseMetas, moduleName, coreModuleTemplate.Spec.Version)
 
-		if assignedChannel != "" {
-			channelsAndVersions[assignedChannel] = coreModuleTemplate.Spec.Version
+		if len(assignedChannels) > 0 {
+			for _, channel := range assignedChannels {
+				channelsAndVersions[channel] = coreModuleTemplate.Spec.Version
+			}
 		}
```
anoipm:
```suggestion
		for _, channel := range assignedChannels {
			channelsAndVersions[channel] = coreModuleTemplate.Spec.Version
		}
```

internal/modules/render.go
```diff
@@ -147,12 +147,16 @@ func convertInstall(details ModuleInstallDetails) string {
 func convertVersions(versions []ModuleVersion) string {
 	values := make([]string, len(versions))
 	for i, version := range versions {
-		value := version.Version
-		if version.Channel != "" {
-			value += fmt.Sprintf("(%s)", version.Channel)
-		}
+		if len(version.Channels) == 0 {
+			values[i] = version.Version
```
anoipm: Why you use indexation instead of append? It's ok for me but few lines later you use append and IMO it is inconsistent.

internal/modules/enable.go
musztardem: eventually I think this `channel == ""` check is not necessary because if the passed `moduleChannel` is empty then we assign there the default kyma channel which is a required field in the kyma spec so it's always to be present. I'll just get rid of that.

internal/modules/render.go
musztardem: My change involved adding the `else` statement here and I left the index access as it was.
The `values` slice was initialized with `make([]string, len(versions))` which initialized a slice of fixed length with zero-value elements, so appending in this context would cause appending to non-empty slice.
I've changed the implementation so it's concise now and only appending happens 👍

Issue #2819, opened 2026-01-13
[CLI modules v2] kyma alpha module list


musztardem:
**Description**

- Implement the `kyma alpha module list` command responsible for displaying a list of installed modules.
- Design the solution in accordance with the new CLI architecture guideline described here: https://github.com/kyma-project/cli/blob/main/docs/contributor/ADR/004-modules-package-architecture.md
- Keep the interface aligned with the proposed API: https://github.com/kyma-project/cli/blob/main/docs/contributor/ADR/003-module-user-interface.md

PR #2884, opened 2026-04-15
Alpha module list

anoipm:
**Description**

- add (alpha/v2) module list command (without community modules - will be added in a follow-up PR),
- improve json/yaml output:
  - separate version and channel into distinct fields instead of concatenating them,
  - correct managed type - managed is a boolean instead of a string.

**Related issue(s)**
See also:
- #2819


internal/modulesv2/repository/installedmodules.go
```diff
@@ -0,0 +1,38 @@
...
+
+type installedModulesRepository struct {
+	kymaClient kyma.Interface
+}
+
+func NewInstalledModulesRepository(kymaClient kyma.Interface) InstalledModulesRepository {
+	return &installedModulesRepository{kymaClient: kymaClient}
+}
+
+func (r *installedModulesRepository) ListInstalledModules(ctx context.Context) ([]kyma.KymaModuleInfo, error) {
```
musztardem, line 21: Here we leak the raw/low-level type `kyma.KymaModuleInfo` while we should return some entity related to the repository. It's `InstalledModulesRepository` so it suggest that it should return some kind of a module. Currently we don't have any `Module` entity in our domain. Also I wouldn't necessarily like the `Module` name as it's very general name in our case, the whole package is called `Module`. I was thinking about something like `ModuleInstallation`, we could have the `ModuleInstallation` entity + `ModuleInstallationRepository`

internal/modulesv2/repository/moduleinstallationstate.go
```diff
@@ -0,0 +1,121 @@
...
+		return status.State, nil
+	}
+
+	moduleTemplate, err := r.kubeClient.Kyma().GetModuleTemplate(ctx, status.Template.GetNamespace(), status.Template.GetName())
+	if err != nil {
+		if apierrors.IsNotFound(err) {
+			return "", nil
+		}
+		return "", errors.Wrapf(err, "failed to get ModuleTemplate %s/%s", status.Template.GetNamespace(), status.Template.GetName())
+	}
```
musztardem, line 40: Repositories should be responsible for mapping between the outside world (k8s cluster/database/some external rest service) and the inside world (our domain layer) and this code is some business logic responsible for the state calculation. This kind of code should be a part of service

internal/modulesv2/list.go
```diff
@@ -0,0 +1,47 @@
...
+		if err != nil {
+			return nil, err
+		}
+
+		results = append(results, dtos.ListResult{
+			Name:                 module.Status.Name,
+			Version:              module.Status.Version,
+			Channel:              module.Status.Channel,
+			ModuleState:          module.Status.State,
+			Managed:              module.Spec.Managed == nil || *module.Spec.Managed,
```
musztardem, line 40: we are lacking the proper Entities here and we are explicitly describing here some internal behavior or some internal business rule, I would imagine we should be able to have something like `moduleInstallation.IsManaged()`

internal/modulesv2/repository/installedmodules.go
```diff
@@ -0,0 +1,40 @@
...
+
+	modules := make([]entities.ModuleInstallation, len(kymaCR.Status.Modules))
+	for i, status := range kymaCR.Status.Modules {
+		raw := kyma.KymaModuleInfo{Status: status}
+		for _, spec := range kymaCR.Spec.Modules {
+			if spec.Name == status.Name {
+				raw.Spec = spec
+				break
+			}
+		}
```
musztardem: Module entry in `kyma.spec.modules` comes from a user and the entry in `status` is populated by the lifecycle manager. When lifecycle manager fails somehow and won't populate the `status` we might lose information that user wanted to enable the module (i.e. module will be present in spec and will be missing in status). I'm not saying it's very probable scenario but I think it's safer to iterate over the module spec in the first place

musztardem: Ok, I've just realized that when we remove entry from the `spec` then we are still going to have entry in the `status` as it's going to be in `Deleting` state for some time, so it goes both ways:
1. user adds entry in the `spec` - for some time we might have entry only in the `spec` and not in `status` before lifecycle manager gets to reconcile the module
2. user removes entry from the `spec` - then for some time we have entry only in the `status` as it's in the `Deleting` state 

So I believe we should iterate separately the status and spec collections and compose the result

internal/modulesv2/list.go
```diff
@@ -0,0 +1,60 @@
...
+		}
+
+		results = append(results, dtos.ListResult{
+			Name:                 module.Name,
+			Version:              module.Version,
+			Channel:              module.Channel,
+			ModuleState:          module.ModuleState,
+			Managed:              module.IsManaged(),
+			CustomResourcePolicy: module.CustomResourcePolicy,
+			InstallationState:    installationState,
```
musztardem: With that approach we've detached the `installationState` from the `moduleInstallation`. I think the `installationState` should live in the `moduleInstallation` and population of it's value should be handled in the `NewModuleInstallationFromRaw`.

internal/modulesv2/repository/moduleinstallationstate.go
```diff
@@ -0,0 +1,114 @@
...
+
+type moduleInstallationStateRepository struct {
+	kubeClient kube.Client
+}
+
+func NewModuleInstallationStateRepository(kubeClient kube.Client) ModuleInstallationStateRepository {
+	return &moduleInstallationStateRepository{kubeClient: kubeClient}
+}
+
+func (r *moduleInstallationStateRepository) GetInstallationState(ctx context.Context, module entities.ModuleInstallation) (string, error) {
```
musztardem, line 18: This is implementation of a separate repository for a single string value that I believe should live in the `ModuleInstallation` as the installation state belongs to the `moduleInstallation`. I see however that there's lot of extraction logic that might bloat the original repository. 

I think we could introduce some kind of `extractors` or `converters` package that could live in the `repository` package that could have specialized classes responsible for such data transformations

musztardem:
There's some discrepancy in how `module list` and `alpha module list` work for managed modules, there's output from my console:

```terminal
$ ./bin/kyma alpha module list
MODULE                  VERSION               CR POLICY         MANAGED   MODULE STATUS   INSTALLATION STATUS
api-gateway             3.6.2(regular)        CreateAndDelete   true      Ready           Ready
application-connector   1.3.1(fast)           CreateAndDelete   true      Ready           Ready
btp-operator            1.3.2(regular)        CreateAndDelete   true      Ready           Ready
cap-operator            0.28.0(fast)          CreateAndDelete   true      Ready           Ready
cfapi                   0.4.0(experimental)   CreateAndDelete   true      Ready           Ready
cloud-manager           1.7.1(fast)           CreateAndDelete   true      Ready           Ready
connectivity-proxy      1.2.2(fast)           CreateAndDelete   true      Ready           Ready
eventing                1.5.3(fast)           CreateAndDelete   true      Deleting        Deleting
istio                   1.26.7(regular)       CreateAndDelete   true      Ready           Ready
keda                    1.10.1(fast)          CreateAndDelete   true      Ready           Ready
nats                    1.5.2(fast)           CreateAndDelete   true      Deleting        Deleting
serverless              1.12.0(fast)          CreateAndDelete   true      Ready           Ready
telemetry               1.63.3(fast)          CreateAndDelete   true      Ready           Ready
transparent-proxy       1.9.5(fast)           CreateAndDelete   true      Ready           Ready
ztis-agent              0.28.0(fast)          CreateAndDelete   true      Ready           Ready
[12:03:07] [cost 0.409s] ./bin/kyma alpha module list

$ ./bin/kyma module list
2026-05-18T10:05:11.442219Z	info	klog	Waited before sending request	delay=1.1237515s reason=client-side throttling, not priority and fairness verb=GET URL=https://api.ee0bdf1.stage.kyma.ondemand.com/apis/operator.kyma-project.io/v1beta2/namespaces/kyma-system/moduletemplates/istio-1.26.7
NAME                    VERSION               CR POLICY         MANAGED   MODULE STATUS   INSTALLATION STATUS
api-gateway             3.6.2(regular)        CreateAndDelete   true      Ready           Ready
application-connector   1.3.1(fast)           CreateAndDelete   true      Ready           Ready
btp-operator            1.3.2(regular)        CreateAndDelete   true      Ready           Ready
cap-operator            0.28.0(fast)          CreateAndDelete   true      Ready           Ready
cfapi                   0.4.0(experimental)   CreateAndDelete   true      Ready           Ready
cloud-manager           1.7.1(fast)           CreateAndDelete   true      Ready           Ready
connectivity-proxy      1.2.2(fast)           CreateAndDelete   true      Ready           Ready
eventing                1.5.3(fast)           CreateAndDelete   true                      Deleting
istio                   1.26.7(regular)       CreateAndDelete   true      Warning         Ready
keda                    1.10.1(fast)          CreateAndDelete   true      Warning         Ready
nats                    1.5.2(fast)           CreateAndDelete   true      Warning         Deleting
serverless              1.12.0(fast)          CreateAndDelete   true      Ready           Ready
telemetry               1.63.3(fast)          CreateAndDelete   true      Ready           Ready
transparent-proxy       1.9.5(fast)           CreateAndDelete   true      Ready           Ready
ztis-agent              0.28.0(fast)          CreateAndDelete   true      Error           Ready
docker-registry         0.13.0                N/A               false     Ready           Ready
registry-proxy          0.14.0                N/A               false     Ready           Ready
[12:05:16] [cost 7.593s] ./bin/kyma module list
```

internal/modulesv2/repository/installedmodules.go
anoipm: Done — we now merge both collections.

internal/modulesv2/repository/moduleinstallationstate.go
anoipm: Refactored. Extracted extractStateFromObject as a function responsible for state interpretation.

internal/modulesv2/list.go
anoipm: Refactored

internal/modulesv2/repository/modulecrstate.go
```diff
@@ -0,0 +1,111 @@
...
+	"github.com/kyma-project/cli.v3/internal/kube"
+	"github.com/kyma-project/cli.v3/internal/kube/kyma"
+	"github.com/kyma-project/cli.v3/internal/kube/rootlessdynamic"
+	"github.com/kyma-project/cli.v3/internal/modulesv2/entities"
+	"github.com/kyma-project/cli.v3/internal/out"
+	apierrors "k8s.io/apimachinery/pkg/api/errors"
+	"k8s.io/apimachinery/pkg/apis/meta/v1/unstructured"
+)
+
+type ModuleCRStateRepository interface {
```
musztardem: I think this class and `moduleInstallationState` repositories should be a part of `ModuleInstallationRepository` because they're technically not repositories - repositories are tightly related to some entity and manage their lifecycles:
* `ModuleInstallationEntity` is managed by `ModuleInstallationRepository` - get/list/create/update/delete

internal/modulesv2/repository/installedmodules.go
```diff
@@ -0,0 +1,112 @@
+package repository
+
+import (
+	"context"
+
+	"github.com/kyma-project/cli.v3/internal/kube/kyma"
+	"github.com/kyma-project/cli.v3/internal/modulesv2/entities"
+)
+
+type InstalledModulesRepository interface {
```
musztardem: I think the repository naming should reflect an association with the entity it's related to, so I would see it as `ModuleInstallationsRepository`

anoipm: Done — renamed to ModuleInstallationsRepository.

internal/modulesv2/repository/modulecrstate.go
anoipm: Done. ModuleCRStateRepository and ModuleInstallationStateRepository are now internal implementation details of ModuleInstallationsRepository

musztardem (dismissed):

docs/user/gen-docs/kyma_alpha_module_list.md
```diff
@@ -0,0 +1,29 @@
+# kyma alpha module list
+
+Lists installed modules.
+
+## Synopsis
+
+Use this command to list the installed Kyma modules.
+
+NOTE: functionality under construction
```
grego952:
```suggestion
> [!CAUTION] 
> This functionality is still under construction. Community modules are not yet supported.
```

docs/user/gen-docs/kyma_alpha_module_list.md
```diff
@@ -0,0 +1,29 @@
...
+  - community modules not yet supported
```
grego952:
```suggestion
```

anoipm: Changed NOTE to WARNING and updated the message

grego952 (approved):

musztardem (approved):

PR #2899, opened 2026-05-04
Single container app tutorials

kwiatekus:
**Description**

Changes proposed in this pull request:

- shows how to use kyma CLI commands and supporting GH actions to easily push single container apps 

**Related issue(s)**
https://github.com/kyma-project/cli/issues/2410

Issue #2410, opened 2025-03-14
Document single container app development scenario with Kyma CLI @ SAP Development Tutorials


kwiatekus:
**Description**

Document tutorial for kyma app deployment automation with Kyma CLI @ https://developers.sap.com/tutorial-navigator.html

Describe how to automate [the scenario](https://github.com/kyma-project/cli/tree/main/tests/btp)
 - [provision kyma via terraform ](https://github.com/kyma-project/terraform-module/?tab=readme-ov-file#how-to-use-the-terraform-module-for-kyma)
 - connect to backend services like BTP Object Store and SAP Hana DB
 - deploying http exposed Node.js app that serves as REST API 
 
For internal readers, reference [sample CI pipeline](https://github.tools.sap/kyma/ci-automation-demo/).
For all reference the [the scenario](https://github.com/kyma-project/cli/tree/main/tests/btp).

**Area**

 * CLI

**Reasons**
This is an official, public SAP learning source which is often referenced among communities like linkedIn etc.
This will have a positive impact on kyma adoption.

**Assignees**

@kyma-project/technical-writers

**Attachments**

PR #2904, opened 2026-05-06
fix(app): make app push idempotent using server-side apply

### LINKED ISSUE #2905: kyma app push fails on rerun with "already exists" error for Deployment and Service
**Description**

Running `kyma app push` a second time for the same application name fails because `Deployment` and `Service` resources are created using the Kubernetes `Create()` API, which returns an `AlreadyExists` error if the resource already exists. This makes the command non-idempotent and forces users to manually delete resources before redeploying.

**Expected result**

Running `kyma app push` multiple times with the same `--name` should succeed. On the first run it creates the resources; on subsequent runs it updates them in-place — the same behaviour as `kubectl apply`.

**Actual result**

The second run fails with errors such as:

```
Error: failed to create Deployment
deployments.apps "my-app" already exists
```

or

```
Error: failed to create Service
services "my-app" already exists
```

Note: `APIRule` is unaffected because it already uses server-side apply internally.

**Steps to reproduce**

1. Have a Kyma cluster with the in-cluster Docker registry available.
2. Run `kyma app push --name my-app --image eu.gcr.io/my-project/my-app:latest`
3. Run the same command again.
4. Observe the failure on the second run.

**Troubleshooting**

Workaround: manually delete the `Deployment` and `Service` before each rerun:
```bash
kubectl delete deployment my-app -n default
kubectl delete service my-app -n default
```

**Related**
Fixed by kyma-project/cli#2904
Found by #2380

kwiatekus:
## Summary

- `kyma app push` previously failed on rerun with \"already exists\" errors for Deployment and Service
- Adds `ApplyDeployment` and `ApplyService` functions using Server-Side Apply (SSA) via the existing `rootlessdynamic.Apply()` infrastructure — the same pattern already used by `CreateAPIRule`
- `push.go` now calls the new `Apply*` functions; existing `Create*` functions are preserved for backward compatibility

## Changes

- `internal/kube/resources/deployment.go` — add `ApplyDeployment`
- `internal/kube/resources/service.go` — add `ApplyService`
- `internal/cmd/app/push.go` — switch to `ApplyDeployment`/`ApplyService`; update output messages from "Creating" to "Applying"
- Tests cover create (first run) and upsert (rerun) paths for both new functions

## Test plan

- [ ] `go test ./internal/kube/resources/... ./internal/cmd/app/...` passes
- [ ] Run `kyma app push --name my-app --image <image>` twice against a cluster — second run should succeed without errors

## Related to 

Fixes https://github.com/kyma-project/cli/issues/2905
#2899 
#2410

### DIFF
--- a/internal/kube/resources/deployment.go
+++ b/internal/kube/resources/deployment.go
@@ -33,12 +33,6 @@ type CreateDeploymentOpts struct {
 	Insecure                   bool
 }
 
-func CreateDeployment(ctx context.Context, client kube.Client, opts CreateDeploymentOpts) error {
-	deployment := buildDeployment(&opts)
-	_, err := client.Static().AppsV1().Deployments(opts.Namespace).Create(ctx, deployment, metav1.CreateOptions{})
-	return err
-}
-
 func ApplyDeployment(ctx context.Context, client kube.Client, opts CreateDeploymentOpts) error {
 	deployment := buildDeployment(&opts)
 	deployment.TypeMeta = metav1.TypeMeta{
--- a/internal/kube/resources/deployment.go
+++ b/internal/kube/resources/deployment.go
@@ -35,10 +35,6 @@ type CreateDeploymentOpts struct {
 
 func ApplyDeployment(ctx context.Context, client kube.Client, opts CreateDeploymentOpts) error {
 	deployment := buildDeployment(&opts)
-	deployment.TypeMeta = metav1.TypeMeta{
-		APIVersion: "apps/v1",
-		Kind:       "Deployment",
-	}
 	unstrObj, err := runtime.DefaultUnstructuredConverter.ToUnstructured(deployment)
 	if err != nil {
 		return err
@@ -85,6 +81,10 @@ func buildDeployment(opts *CreateDeploymentOpts) *appsv1.Deployment {
 	}
 
 	deployment := &appsv1.Deployment{
+		TypeMeta: metav1.TypeMeta{
+			APIVersion: "apps/v1",
+			Kind:       "Deployment",
+		},
 		ObjectMeta: metav1.ObjectMeta{
 			Name:      opts.Name,
 			Namespace: opts.Namespace,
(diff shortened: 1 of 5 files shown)


internal/kube/resources/deployment.go
```diff
@@ -31,10 +33,17 @@ type CreateDeploymentOpts struct {
 	Insecure                   bool
 }
 
-func CreateDeployment(ctx context.Context, client kube.Client, opts CreateDeploymentOpts) error {
+func ApplyDeployment(ctx context.Context, client kube.Client, opts CreateDeploymentOpts) error {
 	deployment := buildDeployment(&opts)
-	_, err := client.Static().AppsV1().Deployments(opts.Namespace).Create(ctx, deployment, metav1.CreateOptions{})
-	return err
+	deployment.TypeMeta = metav1.TypeMeta{
```
musztardem: We have the `deployment` definition built in the `buildDeployment` function starting from line 87, maybe let's move this `TypeMeta` there

internal/kube/resources/service.go
```diff
@@ -16,10 +16,17 @@ import (
 	"k8s.io/utils/ptr"
 )
 
-func CreateService(ctx context.Context, client kube.Client, name, namespace string, port int32) error {
+func ApplyService(ctx context.Context, client kube.Client, name, namespace string, port int32) error {
 	service := buildService(name, namespace, port)
-	_, err := client.Static().CoreV1().Services(namespace).Create(ctx, service, metav1.CreateOptions{})
-	return err
+	service.TypeMeta = metav1.TypeMeta{
```
musztardem: same thing here

musztardem (changes requested):
I'd adjust the deployment/service building functions, otherwise LGTM

internal/kube/resources/deployment.go
kwiatekus: thx, done

internal/kube/resources/service.go
kwiatekus: thx, done

musztardem (approved):

PR #2909, opened 2026-05-07
feat(app push): add --build-tag flag for custom image tags

### LINKED ISSUE #2907: app push: allow custom image tag via --build-tag flag
**Description**

The `kyma app push` command auto-generates a timestamp-based image tag (`<name>:2006-01-02_15-04-05`) when building from source (`--code-path` or `--dockerfile`). There is no way to supply a custom tag.

Add an `--build-tag` flag that lets users specify the tag used for the built image. When the flag is omitted, the current timestamp fallback is preserved.

**Reasons**

In CI/CD pipelines (e.g. GitHub Actions) it is common practice to tag images with the Git commit SHA for traceability and reproducibility:

```bash
kyma app push --name my-app --code-path . --build-tag ${{ github.sha }}
```

Without this flag:
- Every build gets a non-deterministic timestamp tag, making it impossible to correlate a running image with a specific commit.
- There is no way to implement image promotion or rollback based on a known identifier.
- Pipelines cannot enforce a consistent tagging strategy.

The flag should:
- Apply only to source builds (`--code-path`, `--dockerfile`); `--image` already embeds the full tag.
- Validate the value against Docker tag rules (max 128 chars, `[a-zA-Z0-9_][a-zA-Z0-9_.-]*`, no spaces or special characters).
- Return a clear error message on invalid input.

**Attachments**

Relevant code: [`internal/cmd/app/push.go#L355-L374`](https://github.com/kyma-project/cli/blob/main/internal/cmd/app/push.go#L355-L374) — `buildImage()` where the timestamp tag is constructed.

kwiatekus:
## Summary

- Adds `--build-tag` flag to `kyma app push` allowing users to supply a custom Docker image tag when building from source (`--code-path` or `--dockerfile`)
- Falls back to the existing timestamp-based tag when the flag is omitted (no behaviour change for existing users)
- Flag is mutually exclusive with `--image` (which already embeds the tag)
- Tag is validated against Docker tag rules: starts with `[a-zA-Z0-9_]`, only `[a-zA-Z0-9_.-]` allowed, max 128 chars

## Motivation

Enables CI/CD pipelines to use deterministic, traceable image tags:
```bash
kyma app push --name my-app --code-path . --build-tag ${{ github.sha }}
```

Closes #2907

## Test Plan
- [ ] `go test ./internal/cmd/app/... -v` — 15 tests pass (13 validation cases + 2 resolveImageTag cases)
- [ ] `go build ./...` — no errors
- [ ] Manual: `kyma app push --name my-app --code-path . --build-tag abc1234` uses tag `abc1234`
- [ ] Manual: `kyma app push --name my-app --code-path .` still uses timestamp tag (no regression)
- [ ] Manual: `kyma app push --name my-app --image myimage:latest --build-tag abc` fails with exclusivity error
- [ ] Manual: `kyma app push --name my-app --code-path . --build-tag "bad tag"` fails with clear validation error

### DIFF
--- a/internal/cmd/app/push.go
+++ b/internal/cmd/app/push.go
@@ -27,6 +27,7 @@ type appPushConfig struct {
 	namespace                  string
 	image                      string
 	imagePullSecretName        string
+	imageTag                   string
 	dockerfilePath             string
 	dockerfileSrcContext       string
 	dockerfileArgs             types.Map
@@ -43,7 +44,6 @@ type appPushConfig struct {
 	mountServiceBindingSecrets types.ServiceBindingSecretArray
 	quiet                      bool
 	insecure                   bool
-	imageTag                   string
 }
 
 func NewAppPushCMD(kymaConfig *cmdcommon.KymaConfig) *cobra.Command {
--- a/internal/cmd/app/push.go
+++ b/internal/cmd/app/push.go
@@ -332,7 +332,7 @@ func createDeployment(cfg *appPushConfig, client kube.Client, image, imagePullSe
 
 func buildAndImportImage(client kube.Client, cfg *appPushConfig, registryConfig *registry.InternalRegistryConfig) (string, clierror.Error) {
 	out.Msgln("Building image\n")
-	imageName, err := buildImage(cfg)
+	imageName, err := buildImage(cfg, cfg.imageTag)
 	if err != nil {
 		return "", clierror.Wrap(err, clierror.New("failed to build image from Dockerfile"))
 	}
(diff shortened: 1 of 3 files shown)


internal/cmd/app/push.go
```diff
@@ -178,6 +188,17 @@ func (apc *appPushConfig) complete() clierror.Error {
 		}
 	}
 
+	if apc.buildTag != "" {
```
musztardem: This lives in the `complete` method which as  I understand is responsible for populating config data based on some logic. We don't populate the `buildTag` here but rather check if it's complaint with regex so I would move this piece to the `validate` method

grego952 (approved):

internal/cmd/app/push.go
kwiatekus: done, thx

musztardem (approved):

PR #2917, opened 2026-05-19
Fix display of a non existent contexts after busola start

FilipRudy:
**Description**

- Previously, when we started busola using `kyma alpha dashboard` without providing any kubeconfig, user was still prompted to pick a context for a non-existent cluster ending with an error. This PR fixes this issue.

**Related issue(s)**

- #2554

### DIFF
--- a/internal/busola/container.go
+++ b/internal/busola/container.go
@@ -57,7 +57,7 @@ func (c *ContainerRunner) Start(apiConfig *api.Config) error {
 		return fmt.Errorf("failed to create temp dir %q: %w", tmpDir, err)
 	}
 
-	if c.apiConfig != nil && c.apiConfig.CurrentContext != "" {
+	if c.apiConfig != nil {
 		config, err := clientcmd.Write(*c.apiConfig)
 		if err != nil {
 			return fmt.Errorf("failed to serialize kubeconfig: %w", err)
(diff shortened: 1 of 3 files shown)


internal/busola/container.go
```diff
@@ -67,6 +67,8 @@ func (c *ContainerRunner) Start(apiConfig *api.Config) error {
 		if err := os.WriteFile(kubeconfigPath, config, 0700); err != nil {
 			return fmt.Errorf("failed to write kubeconfig at %q: %w", kubeconfigPath, err)
 		}
+	} else {
+		c.apiConfig = nil
```
pPrecel: Why do you set apiConfig to nil if the else statement will execute only when apiConfig is already nil?

internal/cmd/alpha/dashboard/dashboard.go
```diff
@@ -70,3 +78,12 @@ func runDashboard(cfg *dashboardConfig) clierror.Error {
 
 	return nil
 }
+
+func kubeconfigProvided() bool {
```
pPrecel: Let me check if I understand the reason for adding this function. You add this to avoid a situation where the user did not pass the kubeconfig, and the CLI took the default one, and it is not in a healthy state, right? I think a better option would be checking if the client is operational (if it can call anything) and passing the kubeconfig only if it can.

Checking the `--kubeconfig` flag is not enough because a kubeconfig can be passed using env (`KUBECONFIG`) or saved to the default location, which is not detectable (`~/.kube/config`).

pPrecel (approved):

PR #2939, opened 2026-06-19
add community modules to alpha module list

anoipm:
**Description**

- `kyma alpha module list` now shows community modules in a dedicated section below core modules.

**Related issue(s)**
See also:
- #2819


internal/modulesv2/repository/installedmodules.go
```diff
@@ -113,5 +114,46 @@ func (r *installedModulesRepository) resolveInstallationState(ctx context.Contex
...
+	}
+
+	var result []entities.CommunityModuleInstallation
+	for _, mt := range allTemplates.Items {
+		if !isCommunityModule(&mt) {
+			continue
+		}
+
+		if mt.Spec.Manager == nil {
+			continue
```
musztardem, line 135: I think it's not a usual thing to have the manager missing in the spec so I would add at least some `out.Debug` log about invalid module template

internal/modulesv2/repository/modulecrstate.go
```diff
@@ -48,7 +71,7 @@ func (r *moduleCRStateRepository) GetModuleCRState(ctx context.Context, module e
 	return highestStateFromList(crList.Items), nil
```
musztardem: this is the same code as in the `GetModuleCRStateFromTemplate` method, let's do

```go
return r.GetModuleCRStateFromTemplate(ctx, moduleTemplate)
```

internal/modulesv2/repository/installedmodules.go
anoipm: fixed

internal/modulesv2/repository/modulecrstate.go
anoipm: fixed

musztardem (approved):

grego952 (approved):

